PR #9, opened 2015-01-30
Suffix common prefixes from title sort

b333z:
This adjusts the sorting order slightly to be able to ignore common words like "The ", "An ", "A ".

My kodi's seem to just ignore "The " so that's the only one I have enabled, but would be good to pull from host via JSOONAPI  if/when possible.

Thought this might be useful for the moment and might assist with whatever you decide to do with sorting in the future.

### DIFF
--- a/app/src/main/java/com/syncedsynapse/kore2/provider/MediaDatabase.java
+++ b/app/src/main/java/com/syncedsynapse/kore2/provider/MediaDatabase.java
@@ -455,4 +455,47 @@ public class MediaDatabase extends SQLiteOpenHelper {
         db.execSQL("DROP TABLE IF EXISTS " + Tables.MUSIC_VIDEOS);
 		onCreate(db);
 	}
+
+    /**
+     * Tokens to move from prefix to suffix when sorting titles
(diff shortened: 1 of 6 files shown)


SyncedSynapse:
Thank you very much. I'll merge it as is, and tomorrow i'll include an option to enable/disable this.

I'm not sure if this will have an impact on performance with large libraries. Ideally there should be a new field on each table with the title without these articles, but well, this is done, so i'll go with it, and if people complain i'll change it in the future.

I didn't found any option to get the config from jsonrpc, though.

b333z:
Good stuff, SQLite's seems to be pretty efficient, I didn't notice a change here by eye with that single token, but I did notice on my travels that adding a column to avoid processing does seem to be a common approach, I'm sure if you put quite a few tokens in there a pre-calculated column would be the go, its hard to keep track of phones and the CPU v's storage tradeoffs!

Yeah I was unable to find an interface to pull that config from jsonrpc either, I thought you may have found it stashed somewhere, either way I got the impression it might land in the future.

PR #10, opened 2015-01-31
Added German Translation

jonas2515:
Kodi for german users.
Could somebody send a screenshot of the list which includes this string:
<string name="cast_list_text">%1$s und %2$s</string>

It's hard to translate "as", if you don't know the context.

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -1,0 +1,292 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+   Copyright 2015 Synced Synapse. All rights reserved.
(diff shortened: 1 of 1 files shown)


SyncedSynapse:
Thank you very much.

%1$s und %2$s is used in the cast list (eg: Sean Connery as James Bond).

jonas2515:
Ok thank you, now it can be merged.

SyncedSynapse:
How would you like to be credited in the credits (name/alias/whatever)?

jonas2515:
Use my alias, thank you.
I just fixed some typos, I'll make a new pull request. Now it should be perfect.

PR #12, opened 2015-02-01
In song sync batch to db avoiding OOM

b333z:
I had trouble getting my songs to sync, getting out of memory error, I have a quite a few songs and a device on the old side.

This changes the song sync, to write the songs to the db each time before it recurses to get more, rather than collecting all songs then writing them in a single bulk insert.  This seems to be the same approach taken for other parts of the sync.

Submitting for review, tests ok for me, not sure ready for merge but posting in case useful.

### DIFF
--- a/app/src/main/java/com/syncedsynapse/kore2/service/LibrarySyncService.java
+++ b/app/src/main/java/com/syncedsynapse/kore2/service/LibrarySyncService.java
@@ -1168,7 +1168,16 @@ public class LibrarySyncService extends Service {
             action.execute(hostConnection, new ApiCallback<List<AudioType.DetailsSong>>() {
                 @Override
                 public void onSucess(List<AudioType.DetailsSong> result) {
-                    allResults.addAll(result);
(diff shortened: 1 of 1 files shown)


SyncedSynapse:
Thanks. I can't review/test this right now, but i'll take a look in a few days. I agree that syncing should be improved and this seems a legit way to do it.

SyncedSynapse:
Looks good and makes sense. Don't really know why i did it the way it was (actually i know, i was in a hurry to do it...).

I'm going to merge it and do the same thing to the other syncs, where possible.

Tomorrow or Friday could you please pull it and test it to check if everything works for me.

Thanks

b333z:
Tested your changes, looks good, tests out ok from here, did pickup an somewhat unrelated issue in my travels, if you swipe a refresh on the AlbumListFragment that is targeted at a single artist it will resync the entire music collection again.  So we need a artistId == -1 ? syncAllMusic() : syncSingleArtist(artistId) in the onRefresh, was hoping to take a crack at it to help out, but afk for a few days, so will create an issue for it.

Issue #19, opened 2015-02-11
Remote - missing "Title" button


tasn:
Hey,

On the XBMC remote (the obsolete open source one), there's a button called "Title" that let's you get into extra menus and is useful. A good example is to go from the main kodi screen to:
Videos -> Files
You now need the "Title" button in order to expose the extra menu that will let you rescan content and etc. Kore has "Codec" instead of the "Title" button, which does nothing in this case.

Thanks.

SyncedSynapse:
Done 9466d9e0a897433b9ca2468bbfbdb1f0b4986da1

tasn:
Thanks for the quick resolution.

PR #20, opened 2015-02-13
Make codec button switch between codec/context if playing/notplaying

b333z:
This is what I have been using to allow the remote to trigger the context menu.  Without which it is difficult to do a few things such as queue a playlist.

This commit makes the codec button context sensitive so if media is loaded ( ```switchToPanel(R.id.media_panel)```) the codec button triggers, otherwise the context menu triggers (```switchToPanel(R.id.info_panel)```).

This is not ideal, as even though media is loaded/playing/paused and the media_panel might be loaded, the media could be in the background and you could be wandering around the menus wanting to trigger the context menu, forcing you to stop the media so that switchToPanel goes back to the info_panel to trigger the context menu.  Am unsure if you can ask kodi if the media is in the background? Also the bitmap would need to be also switched on the codec button which I have not done.

I probably never need a codec button and would have no issue if it were just a context button, but this seemed like a good compromise at the time.  Might be better to have it always a context menu button and hide the codec function away in a menu or something.

Anyhow noticed someone else reported an issue #19 about this so posting this in case useful, feel free to close or I can work on it further if required.

### DIFF
--- a/app/src/main/java/com/syncedsynapse/kore2/jsonrpc/method/Input.java
+++ b/app/src/main/java/com/syncedsynapse/kore2/jsonrpc/method/Input.java
@@ -349,6 +349,27 @@ public class Input {
     }
 
     /**
+     * Open context menu
+     */
+    public static final class ContextMenu extends ApiMethod<String> {
+        public final static String METHOD_NAME = "Input.ContextMenu";
+        /**
+         * Open context menu
+         */
+        public ContextMenu() {
+            super();
+        }
+
+        @Override
+        public String getMethodName() { return METHOD_NAME; }
+
+        @Override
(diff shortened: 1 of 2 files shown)


SyncedSynapse:
Hi, thanks for this, but you (and others) have convinced me that the codec button isn't that helpful, so what i was planning to do was to substitute it by the context menu (labeled either "context" or "title", not sure yet) in all situations, and maybe present the codec menu on a long press on the "info" button.
I was planning to do this as soon as i end up what i'm working right now (notifications), so i'm not sure i should merge this right now.

b333z:
That's cool, that sounds like a good plan, I'll close this one off

Issue #14, opened 2015-02-01
[Feature Request] Use hardware volume keys on device


2Zero:
As a user I would find it useful to control the volume of Kodi using the hardware volume keys on my android device whilst the remote screen of Kore is active.

SyncedSynapse:
Done

PR #21, opened 2015-02-14
Add support for hardware volume keys

b333z:
This commit adds support for the hardware  volume keys as requested in #14


SyncedSynapse:
Thanks. Not sure why, but i was convinced that more permissions were needed to control volume keys... don't know where i got that ideia...

PR #27, opened 2015-02-20
update german translation

pollonamid:
maybe it would be good to split the contentDescription strings from the button caption strings. In German the next translation in the setup view isn't the same like next button on the remote.


jonas2515:
Thank you, I will translate the new strings soon.

PR #33, opened 2015-02-26
Remove the need of creating extra file to be able to build

trishika:
Setting the `IAP_KEY` varaible and creating the `app/keystore.properties` file are not needed anymore with this set of patch.
This make it easier to build a debug version of the app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,12 +11,6 @@ Building
 
 Git pull should get you almost all you need, as long as you have a working [Android build system][4]
 
-You'll need to create the following files:
-
-1. `gradle.properties`, with the value `IAP_KEY` set to something (this is used for the in-app purchase).
-2. `app/keystore.properties` with the values `store`, `alias`, `pass`, `storePass` set. This is for signing the release build. Alternatively, comment the `signingConfigs` in `app/build.gradle`.
-
-
 Credits
 -------
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -14,6 +14,10 @@ android {
         buildConfigField("String", "IAP_KEY", "\"${rootProject.property("IAP_KEY")}\"")
     }
 
+    lintOptions {
+        abortOnError false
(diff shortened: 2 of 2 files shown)


SyncedSynapse:
Hi,
I haven't merged this yet because i'm torn about it. 
I agree with the removing of the IAP_KEY, but i'm not so sure about the signing.release part (and the lintOptions). If that stays as is, you can still build a debug release, without creating a signing certificate, can't you

trishika:
From my part without creating a keystore.properties file, the build fails : 

    $ ./gradlew build
    [snip]
    A problem occurred evaluating project ':app'.
    > [snip]/keystore.properties (No such file or directory)

Same things without the lintOptions : 

    $ ./gradlew build
    [snip]
    FAILURE: Build failed with an exception.

    * What went wrong:
    Execution failed for task ':app:lint'.
    > Lint found errors in the project; aborting build.
  
      Fix the issues identified by lint, or add the following to your build script to proceed with errors:
      ...
      android {
           lintOptions {
              abortOnError false
          }
      }
      ...

SyncedSynapse:
The `build` target depends on `assemble` and `check`, and `assemble` depends on `assembleRelease`, which can't be built if you don't have a key.

To build a debug variant you should be running `./gradlew assembleDebug` directly, instead of trying to generate the release build each time (tip, `./gradlew iD` builds a debug variant, installs it on your device and is faster to type).

trishika:
it still give me the same error about the `keystore.properties` file missing even with the `installDebug` or `assembleDebug` target.

SyncedSynapse:
Right. I've used your code here 4d7a945c5303da1720ccfc36be288a09007eda05 . I didn't disable the abortOnError option, but corrected the errors that lint was complaining about.

Could you test it on your end.

trishika:
Works for me now. thx

PR #36, opened 2015-03-09
FIX: draw selector on host selection (non-touch usage)

koying:
When using an actual remote (non-touch), no highlight was drawn around the host cards

FYI, a distant project of mine would be to make Kore a thin Kodi client on AndroidTV which would use, e.g., an headless Kodi as server.

### DIFF
--- a/app/src/main/res/drawable/selector.xml
+++ b/app/src/main/res/drawable/selector.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:drawable="@android:drawable/list_selector_background" android:state_focused="true"/>
+    <item android:drawable="@android:color/transparent"/>
+</selector>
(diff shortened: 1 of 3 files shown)


SyncedSynapse:
To be honest, i didn't pay much attention to non-touch usage in the UI (with physical keys for instance), so this probably happens in a lot of places...
If you need help, let me know.

SyncedSynapse:
koying,

Is the <FrameLayout> on grid_item_host.xml strictly necessary? It is slightly breaking touch feedback on other areas (Host list - fragment_host_list.xml), because the ripple extends past the card (on lollipop).

I'm not able to replicate the issue you had, but could you check if removing the <FrameLayout> and using
        android:listSelector="?attr/selectableItemBackground"
in fragment_add_host_zeroconf.xml works.

koying:
No, doesn't work unfortunately.
Strangely, the movie grid, which seems to be using the same layout, does show the highlight on the card...

SyncedSynapse:
(my previous message got garbled, don't know why)

Can you check if this patch solves the issue:
http://pastebin.com/wU5pmBRd

If it does i'll commit it because it maintains correct touch feedback on the list.

koying:
Yep, working fine, thanks :)

Out of curiosity, what did fix it? (The diff is messed up by the revert)

SyncedSynapse:
The listSelector should indeed be selectableItemBackground, but drawSelectorOnTop="true" also needs to be set. I also cleaned up a little bit the grid view, as it wasn't consistent with the others.

It's commited, so you might want to pull it all now.

PR #41, opened 2015-03-16
Add json rpc method/types to support file browsing/playing

DanhKE6D:
I would like to add the new json rpc commands / data types to support music/video file browsing/queuing/playing feature on Kore.


SyncedSynapse:
jenkins build kore

SyncedSynapse:
Hi Dahn, i was looking through this and noticed that it is not consistent with the json rpc API. The implemented API closely follows the json rpc specification available here: 
http://kodi.wiki/view/JSON-RPC_API/v6
(or in doc/introspect.json) and the structure should be the same.
Looking through this i noticed the following:
1. Files.GetDirectory should return List.Item.File. Files.GetSources should return List.Items.Sources.  In the code, both are returning a Files.FileLocation, which doesn't exist in the json rpc document. I understand that you probably looked at the responses, and as they were similar, created that type to represent it, but to keep it consistent with the specification, they need to return different things. There's no need to implement all the specification, but the basics should get implemented.
2. Player.Open with a media_uri is a little different in the specification. It should take a PlaylistType.Item, which is already implemented, and in there you specify the "file" that should be opened. No need to create the type PlayerType.Item (this type doesn't exist in the json rpc specification)
3. In Playlist.Add there's no need to create a new method that takes a file. The existing method takes a Playlist.Item, which has a field for specifying the file to be added.

I know that the json rpc api can be confusing, but it's important to keep it consistent with the specification and as complete as possible.

DanhKE6D:
Thanks. Let me redo it. This is my first experience with json rpc. Yes, I did look at the responses and create objects as I go along. I did not know that I didn't follow the protocol.

DanhKE6D:
Hi SyncedSynapse,

I redo the implementation of File.GetSources and File.GetDirectory and a few jsonrpc data types. Hopefully, this is now correct.

SyncedSynapse:
jenkins build kore please

SyncedSynapse:
Looking better. Just a question:

Does the implementation of ListType.ItemFile have everything you'll need? I mean, looking at jsonrpc, List.Item.File has file/filetype/lastmodified/mimetype/size and in the implementation you're only getting filetype. This is returned in GetDirectory, and i would thought that the other attributes would be relevant, or am i missing something. Do you only want the filetype (which only indicates whether it is a regular file or a directory)?

Minor stuff:
Files.GetSources takes a parameter media, which in jsonrpc is an enum defined in Files.Media used in other places. There should be an interface FilesType.Media with those strings defined (take a look at for instance AudioType.Fields.Artists for an example)

In the same spirit Files.GetDirectory should take the media parameter, or else, if you don't need it, you might omit it altogether, but not pass it into jsonrpc (the default is "files" already).

DanhKE6D:
I have been learning quite bit about json rpc API for the last couple days. Once I was familiar with the API document, I understand your json prc library a whole lot better :)

To answer your question, filetype is all what I need to decide if the item can be played/queued or to traverse the directory up or down.

I will implement an interface for Files.Media. It it currently defined in the fragment that uses File.GetSources.

File.GetDirectory: I can remove the media parameter if file is already default.

Question: Can I have Files.Media interface pushed in on the next PR which is the Media File Browser PR. I completed already. Did a bit testing. It looked good on my setup here.

SyncedSynapse:
Thanks. I've merged this. I still think that ItemFile should have the other attributes, for completeness, but i can add that.

Regarding the last PR, feel free to push it, though i'd like to take a look at it, and i will probably only be able to do so in the weekend.

PR #45, opened 2015-03-19
Implement media file browsing/queuing/playing feature

DanhKE6D:
Here is my media file browser implementation.

Danh

korenavdrawer
musicbrowser


SyncedSynapse:
Looks good (well, except for the files icon :) but i'll change that). I'll look over it and test it on the weekend.
Thanks.

DanhKE6D:
I knew it :) I am no good with icon drawing so I download a couple free ones trying to get the featuring going. I knew before hand that you would fix the icons :) because I can tell that you are a perfectionist when I read your code.

SyncedSynapse:
Looks good. Merging and i'll do some cosmetic changes afterwards.

If you'd like to improve this, here are some suggestions:
1. Hitting "Back" should go back one level in the directory structure.
2. Clicking a file should start playing it automatically. I'm not sure about this, as i don't use kodi in file mode, but it seems logical
3. Pictures should also be present. I'll probably add this
4. Info about the file size should be present in the list. Also it is possible to present some kind of picture (kodi ui does it), but i haven't figure out where it should be obtained from.

Thanks

DanhKE6D:
@SyncedSynapse 
Thanks for the suggestions. I am planning to do:
1. Will implement on the next PR.
2. I can insert the selected media into the playlist queue and start playing immediately if there is nothing playing. I still prefer to keep the context popup menu because I can play the whole folder pretty quickly when enter the folder then select 'play from here' to insert the content of the whole folder into the playlist.
3. I can also do the pictures. I did not implement it because I don't use Kodi for picture slideshow that much.

Thanks

Issue #47, opened 2015-03-23
Browsing Files show all files instead of only the correct type


SyncedSynapse:

DanhKE6D:
I am working on fixing this too. Currently when browsing files, json rpc reports the file types as "unknown" because I don't believe it reads the media metadata during browsing. Should I assume the media type based on file extension(s) then?

SyncedSynapse:
No, don't use the file extensions. Files.GetDirectory also receives a parameter "Media" which should filter the type of files it returns. Have you tried it?

SyncedSynapse:
@DanhKE6D 
I've played a little bit with `Files.GetDirectory`, and it really should be invoked with the parameters `media` and `properties`.
`media` filters the output: if you set it to `video` it only returns videos, etc.
`properties` makes it return much more info, including metadata. Take a look at the jsonrpc to check what can be in there.

Check a request like this:
`{"jsonrpc":"2.0","id":1,
"method":"Files.GetDirectory",
"params":{"directory":"/home/amsilva/Public/Music Videos/","media":"video","properties":["file","mimetype","title","size","thumbnail","fanart","album","artist","track","showtitle","season","episode","year","duration"]}}`

If the files are on your library it returns all of the metadata, and even if it isn't, it at least returns the `thumbnail`, which can be used to set the picture on each row. 
Nice.

I'll have some time tomorrow, and i can start this. Have you changed anything yet?

DanhKE6D:
@SyncedSynapse 
I was quite busy the last couple days. I started to fix the back key so that it would traverse one directory up. I still need to store and restore the current directory after rotation is completed. If you have time, please start it. I will do a PR for #48 after you are done. Thank you.

DanhKE6D:
@SyncedSynapse 
I played with the properties that you suggested using the new Files.GetDirectory method that you just extended. Only file, mimetype, and size return relevant values. For example, in video file browsing mode, with the mimetype property selected, only video files are returned even though the directory has a mix of music, video and picture files.
I was hoping to get a thumbnail, but that was not the case.
I will create another PR for this fix.

SyncedSynapse:
Fixed on a8f9c3e660e448449357418523b53da3c1c35f3b

PR #51, opened 2015-03-29
File.GetDirectory now uses mimetype property to filter out unrelated med...

DanhKE6D:
...ia/files

Fix for bug report #47


SyncedSynapse:
I've updated this on one of the last commits. Get the latest version to see the changes.

PR #100, opened 2015-07-19
Added Vibration on Remote Arrow Press

### LINKED ISSUE #95: [FEATURE REQUEST] Please add option to enable vibration on button press
@SyncedSynapse :

Could you please add an option to Kore which allows enabling vibration on button press, so that the device vibrates when you press a button on the touchscreen?

It would be much appreciated.

Regards

akshay2000:
Arrow buttons on remote give out 100ms vibrations on press. Setting to enable and disable the vibrations has been added too.
Fixes #95.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -6,7 +6,8 @@
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-permission android:name="android.permission.CHANGE_WIFI_MULTICAST_STATE"/>
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE"/>
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>/>
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
+    <uses-permission android:name="android.permission.VIBRATE"/>
 
     <application
         android:allowBackup="true"
--- a/app/src/main/java/org/xbmc/kore/Settings.java
+++ b/app/src/main/java/org/xbmc/kore/Settings.java
@@ -88,6 +88,10 @@ public interface Settings {
     public static final String KEY_PREF_USE_HARDWARE_VOLUME_KEYS = "pref_use_hardware_volume_keys";
     public static final boolean DEFAULT_PREF_USE_HARDWARE_VOLUME_KEYS = true;
 
+    // Vibrate on remote button press
+    public static final String KEY_PREF_VIBRATE_REMOTE_BUTTONS = "pref_vibrate_remote_buttons";
+    public static final boolean DEFAULT_PREF_VIBRATE_REMOTE_BUTTONS = false;
+
     // Current host id
     public static final String KEY_PREF_CURRENT_HOST_ID = "current_host_id";
     public static final int DEFAULT_PREF_CURRENT_HOST_ID = -1;
--- a/app/src/main/java/org/xbmc/kore/ui/RemoteFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/RemoteFragment.java
@@ -136,8 +136,8 @@ public class RemoteFragment extends Fragment
 
     private Animation buttonInAnim;
     private Animation buttonOutAnim;
-    // Touch listener that provides touch feedbacl
-    private View.OnTouchListener feedbackTouckListener;
+    // Touch listener that provides touch feedback
+    private View.OnTouchListener feedbackTouchListener;
 
     // EventServer connection
     private EventServerConnection eventServerConnection = null;
@@ -161,7 +161,7 @@ public class RemoteFragment extends Fragment
 
         createEventServerConnection();
 
-        feedbackTouckListener = new View.OnTouchListener() {
+        feedbackTouchListener = new View.OnTouchListener() {
             @Override
             public boolean onTouch(View v, MotionEvent event) {
                 switch (event.getAction()) {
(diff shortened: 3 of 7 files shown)


wp9015362:
Thank you for having submitted a pull request for my [feature request](https://github.com/xbmc/Kore/issues/95).

But:

> @akshay2000 wrote:
> 
> Arrow buttons on remote give out 100ms vibrations on press.

Does your pull request only add vibration for the arrow buttons?

My feature request was about adding vibration to all buttons on the remote.

So, any chance you could add vibration to all buttons on the remote?

My feature request came from the intention to give Kore a tactile feedback.

So, it wouldn't make much sense to only have that tactile feedback on the arrow buttons.

All buttons on the remote should have that tactile feedback IMHO.

Any chance you could do it?

It would be much appreciated.

Regards

akshay2000:
@wp9015362  I thought of adding tactile feedback everywhere, but that doesn't really provide any benefit. I have done some A/B testing earlier (for [Windows Phone Kodi Remote](https://github.com/akshay2000/XBMCRemoteRT)) and that analytics data suggests that users prefer to have remote keys to vibrate.  
Reasoning behind this could be that when you have all the buttons vibrating, it's same as all buttons *not* vibrating. In essence, there is no way for the user to differentiate between frequently used arrow keys and infrequently used other keys without looking at the phone.  
Maybe you could try this version from [this](https://www.dropbox.com/s/7x6zv9wu0qrid0f/app-debug.apk?dl=0) debug APK. Try and see if you like it. If not, I'll wait for one of the Team Kodi members to comment on it. I'm not sure if they even want to accept this.

wp9015362:
@akshay2000 :

I disagree. I think having vibration on all buttons is great.

I know that it is great because I have an LG G2 smartphone.

The LG G2 / G3 / G4 smartphones come with an app called "LG Quick Remote".

Here you can see a video of it:

https://www.youtube.com/watch?v=5o6Z_YyfZXQ

The LG Quick Remote app has a setting called "Vibrate on tap".

When this setting is enabled, all buttons on the remote do vibrate when they are being pressed.

And this feels amazing.

What's also nice about the LG Quick Remote vibration is, that even when you press and hold a button on the remote, it continues to vibrate. But not with a constant vibration. When you hold down the channel plus or volume plus button for example it makes one short vibration after another with a little break inbetween, everytime a IR blast is being send out.

And it even makes the notification light blink once per IR blast.

It's really nice!

If Kore would do it like that, that would be really awesome.

Please do not just make the arrow buttons vibrate.

Please at least add an option to also make all buttons vibrate.

And, sorry, but I don't want to install APKs manually.

And I really hope that Team Kodi will accept this. Because when you have used LG's Quick Remote, you really miss the vibration feature in Kore.

Regards

wp9015362:
PS:

It would also be nice if you could add an option to configure the strength of the vibration.

Regards

akshay2000:
@wp9015362 Interesting. I had implemented the ability to send continuous vibration pulses when button is held down. But then I removed it.  
At this point, I really don't know what the feature roadmap Team has planned for the app. And since you are not planning to install APK (I didn't inject malware, I promise!) there is no point in implementing further features. I don't even have a device to test vibrations. So, if the Team confirms that the feature is planned, then I will go ahead and implement it. If not, you're either stuck with no tactile feedback or manually managing APK.

SyncedSynapse:
I'm not opposed to adding vibration (as an option), and i'll take a look at this later. As for if all buttons should vibrate i don't have a strong opinion about this. I'll try it and see if i form an opinion.

@akshay2000 Good work on the remote for windows (haven't tried it, but looks good). Regarding this PR, i've just glanced at it, and it looks good. Just check if you need the vibrator on the remoteFragment, i believe it is superfluous.

wp9015362:
> @SyncedSynapse wrote:
> 
> I'm not opposed to adding vibration (as an option), and i'll take a look at this later. As for if all buttons should vibrate i don't have a strong opinion about this. I'll try it and see if i form an opinion.

Well, it would be nice if you would let the users decide for themselves via options.

You could add options for:

- Configuring if the vibration is enabled on all buttons or only on the arrow buttons
- Configuring the strength of the vibration
- Configuring the length of the vibration
- Configuring if the notification light blinks when a button is pressed

That way the users could configure how they like the vibration themselves.

That would be better than forcing one specific vibration config on the user.

Regards

MartijnKaijser:
i'm ok with the vibration as it's default off. We should not add too many option as it's overkill.

wp9015362:
Well, yeah, okay, the options for the vibration strength/length and notification light might not be that much needed (but still nice to have anyway).

But an option to enable vibration for all buttons instead of only for the arrow buttons would be essential IMHO.

Because having vibration only for the arrow buttons would make the vibration feature pretty much useless. At least for me.

Regards

akshay2000:
> We should not add too many option as it's overkill.

Right. Too many choices paradox.  
So, here is the plan for me:

* Remove unnecessary references.
* Enable vibration for all the remote buttons.
* Just keep one setting to turn it on or off.
* Enable continuous vibration pulse for buttons that send repeated signals.

By the way, @SyncedSynapse Is there a way we could get Windows Phone/Windows project under the umbrella along with iOS and Android apps?

akshay2000:
I have added vibrations to all the buttons on remote. This request can now be merged.

SyncedSynapse:
@akshay2000 I still haven't been able to test this (i'm on the middle of something else), but looking at the code it looks like you've added vibration to *all* buttons, including play/ff/re and home/movies/etc. 

I think these buttons shouldn't have vibration, as they are "normal" Android buttons and these don't have vibration associated with them. I think that vibration should only be applied to the d-pad and maybe the surrounding buttons, but not the "normal" ones. Also, vibration helps when the user is navigating Kodi's Ui and is not looking at the phone, and this only applies to the d-pad.

akshay2000:
@SyncedSynapse That makes sense. In fact, that was my initial idea. I'll make the changes and push a commit.

SyncedSynapse:
Looks good, though i feel that 100ms vibration is a little too much. I might change this...

PR #107, opened 2015-07-31
Added playing or queueing artist/album/genre from music screen

poisdeux:
Implemented context menus in music items shown on the music screen. 
This allows the user to add everything from an artist, album, or genre by popping up the context menu and selecting either play or Add to playlist.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/AlbumListFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AlbumListFragment.java
@@ -41,6 +41,7 @@ import android.widget.AdapterView;
 import android.widget.CursorAdapter;
 import android.widget.GridView;
 import android.widget.ImageView;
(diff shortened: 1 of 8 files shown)


SyncedSynapse:
I will only be able to look and test this in a few weeks, but from a quick glance it looks good. Not sure about the name "MediaManager" for a class that just keeps utilities methods - MediaManager makes you think it does something "important", which it doesn't in this case, but i'll look into it in more detail later.

SyncedSynapse:
Looks good. Like i said, i'm not a fan of the name "MediaManager" and i will probably change it to reflect its Utils nature.
Thanks.

poisdeux:
How about MediaPlayerUtils?

SyncedSynapse:
I've already changed it to MediaUtils, but that makes more sense, so i will change it again to that.

Issue #99, opened 2015-07-13
[Bug]Enter/Return on external/physical keyboard for text input


gvkt:
This is an edge case that is likely not a high priority but if super-easy to fix, it would be nice for consistency.

When the text input dialog is up to send text to a dialog in Kodi, typically you use the on-screen keyboard to enter the text on the local dialog on the remote (which is a very useful feature) and then hit the enter/return button on the on-screen keyboard which will be labeled Send in that instance of the keyboard. This sends the text to Kodi and works as expected.

However, if one is using an external keyboard or a device with a built in physical keyboard, hitting the enter/return key on that keyboard after input of text just pops down the local text input dialog and sends nothing to Kodi. The text dialog on Kodi remains empty and up while there is no text dialog locally on the remote any more. So if you happened to do this on the external/physical keyboard, you have to select Send Text Input form the overflow menu, enter the text again and remember to tap Send on the onscreen dialog.

PR #113, opened 2015-08-19
#99, fixed external keyboard return key to send text instead of just …

nykysoft:
…dismiss dialog

There was no handler for external keyboard enter key. I just added handler (if statement) for that case. When user is pressing enter key there is actionId 0 (unspecified) and event  keycode ENTER_KEY.


SyncedSynapse:
I can't test this, but i trust you, so in it goes.

PR #116, opened 2015-09-02
Improved keeping user informed on sync progress

poisdeux:
Implemented binding to LibrarySyncService to check if there are items currently
syncing or queued to sync. This makes it possible to inform user of background
sync processes. 
Especially with slow devices running Kodi and/or large libraries, syncing takes a long time. When switching between different screens the refresh animation would not show. By checking if a sync process for the currently displayed items is still in progress we now manually start the refresh animation or not.

Redesigned the code quite a bit to prevent me from having to copy/paste duplicate code into every list fragment that should notify user of a running sync process. For this I created two abstract classes.
Hope you don't object :).

This pull request also fixes an issue with SwipeRefreshLayout from appcompat library. It does not
always show the refresh animation when refresh is set to true. See https://code.google.com/p/android/issues/detail?id=77712 for details

### DIFF
--- a/app/src/main/java/org/xbmc/kore/service/LibrarySyncService.java
+++ b/app/src/main/java/org/xbmc/kore/service/LibrarySyncService.java
@@ -16,16 +16,23 @@
 package org.xbmc.kore.service;
 
 import android.annotation.SuppressLint;
+import android.app.Activity;
 import android.app.Service;
+import android.content.ComponentName;
 import android.content.ContentResolver;
 import android.content.ContentValues;
+import android.content.Context;
 import android.content.Intent;
+import android.content.ServiceConnection;
 import android.net.Uri;
+import android.os.Binder;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.HandlerThread;
 import android.os.IBinder;
 import android.os.Process;
+import android.support.v4.app.Fragment;
+import android.util.Log;
 
 import org.xbmc.kore.host.HostInfo;
 import org.xbmc.kore.host.HostManager;
@@ -44,6 +51,7 @@ import org.xbmc.kore.utils.Utils;
 import java.lang.System;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 
 import de.greenrobot.event.EventBus;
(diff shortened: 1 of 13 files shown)


poisdeux:
Too many changes in a single request?

SyncedSynapse:
Not enough time to check it... sorry. It's still marked as unread on my inbox, and i hope to get around it soon.

SyncedSynapse:
I would like to apologize for taking so long to check this. It looked like a big change so i didn't dive into it, but it's actually straightforward.

So, first of all, i had some trouble merging this, given the current state of the repo. Git complains about `ArtistListFragment.java`, so i had to do a `git apply --reject` and manually incorporate the rejected changes (this might be because i changed the `MediaManager` name in a previous PR). Github isn't complaining, so it is possible that it might be able to merge it seamlessly, but i'm not sure i want to try it blindly.

I'd like to ask you to confirm this. Could you get a current version of the repo, apply this PR and check the rejections. The corrected `ArtistListFragment.java` is here: http://pastebin.com/2XjPkZCC.

I haven't checked every line of code, but afaict you have 2 big changes here:
1. The first is a refactoring of the lists code, which makes sense and should have been done from the start (guess i was lazy or in a hurry to just have something working...).
2. The second is making the `LibrarySyncService` boundable, so that when the list is entered you can check if an update is running and show the refresh animation.

Is this it, or did i miss something? Should i be aware of/check anything else that might be going on?

Lastly, i'm only able to test this for the music library. My movies and tv shows library sync too fast to be able to get out of that screen and return to it before the sync finishes. Have you been able to test those?

Thanks

poisdeux:
No need to apologize. I was just wondering if there was something I could do to make this request more manageable.

I had a merge conflict in ArtistListFragment when I merged the changes on master on 4th september:
https://github.com/poisdeux/Kore/commit/bee35c216d5e668fa24e47a35a220d95966372f6
But these conflicts were resolved. So unless you were trying to merge an older version the merge conflict should not occur. 
I checked your version on pastebin with mine and could only notice two whitespace differences. I do not think Git will have a problem with that. 
I fetched your latest changes on master and merged the changes into this branch. It should now be a simple fast-forward merge.

You are right about the two big changes this request introduces. 
I also added showRefreshAnimation() in AbstractListFragment to fix starting the refresh animation manually. It is a strange bug, as I only encountered it using an Android Virtual Device. But as other people seem to encounter the same problem I implemented the proposed fix.

I tested the service binding for movies, music, and tv shows with success. However, I just noticed that I forgot to implement it in addons. Should I include it in this request or create a new PR?

SyncedSynapse:
Ok, i'll just test it for a while more and merge it, probably tomorrow.

The addons list doesn't use the sync service, it's a direct jsonrpc call to Kodi, so this doesn't apply.

PR #117, opened 2015-09-03
Implemented checking if network is enabled when searching for hosts

poisdeux:


SyncedSynapse:
One question: do you have any reason to keep the "Next" button available when there's no network connection? If there's no network the next step won't work either.

poisdeux:
Only reason is that with manual configuration the network connection is not required. We might want to add the check at the manual configuration screen as well for the test button. Although, personally I think at that point the user should be aware network is not configured from the previous screen...

poisdeux:
BTW: wouldn't "manual" be a better label for the "next" button?

SyncedSynapse:
Well, if it can't check the info the user inputs in manual configuration, it won't save it, so there's really no point in going forward. I've removed the next button when there's no network connection.

Regarding the label, maybe you're right, but somehow i feel that, given that you're in a wizard, "next" is more appropriate.

PR #120, opened 2015-09-18
Update fr translations

sohelzerdoumi:
I just forgot one key **vibrate_on_remote** because i don't know how to translate it neatly.

### DIFF
--- a/app/src/main/res/values-fr/strings.xml
+++ b/app/src/main/res/values-fr/strings.xml
@@ -276,6 +276,45 @@
 
         Si vous avez besoin d\'aide, consultez le <b><a href="http://forum.kodi.tv/forumdisplay.php?fid=129">forum</a></b>
     ]]></string>
+    <string name="no_network_connection">Aucune connexion réseau</string>
+    <string name="reboot">Redémarrer</string>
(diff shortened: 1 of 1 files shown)


SyncedSynapse:
This breaks the build because some apostrophes aren't preceded by a \ (for instance: `Jouer à partir d'ici` should be `Jouer à partir d\'ici`.

Regarding vibrate_on_remote, wouldn't something like `Vibrer à appuyer sur les boutons` work?

sohelzerdoumi:
I fix apostrophes but no this translation don't work. I don't understand the feature behind the sentence.

SyncedSynapse:
When it is enabled, the buttons on the remote screen (left/right/up/down, etc) vibrate to give a tactical feedback. If you switch it on doesn't it work (assuming your device can vibrate)?

sohelzerdoumi:
I just tried and now i understand the feature. The english sentence is ambiguous, i thought the remote screen was my computer or my raspberry.  But, it seems to be a "vibrate on screen touched".
I will translate the last key now.

sohelzerdoumi:
Translation done.

PR #128, opened 2015-10-23
Implemented using poster image from gridview item as placeholder

poisdeux:
In most detailed views of a music album, tv show, or movie, the same image is
shown as is used in the overview of artists/albums, tv shows, or movies.
Instead of showing a CharacterDrawable as placeholder while the poster is loading
we now show the image from the overview as placeholder.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/AbstractDetailsFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AbstractDetailsFragment.java
@@ -1,0 +1,233 @@
+/*
+ * Copyright 2015 Martijn Brekhof. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.xbmc.kore.ui;
+
+import android.annotation.SuppressLint;
+import android.content.Intent;
+import android.graphics.Bitmap;
+import android.graphics.drawable.BitmapDrawable;
+import android.graphics.drawable.Drawable;
+import android.os.Bundle;
+import android.support.annotation.Nullable;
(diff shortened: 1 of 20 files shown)


SyncedSynapse:
There's a simpler way to achieve this: rely on Picasso cache. Check out the attached diff and see if it does what you want (change the file extension). Note that it only changes the movie and tv show list, not the album list, but i think there's not really an issue there.

The thing is, Picasso has a cache which takes into account the image url and its size. The size can and is setted (?) programatically, but the list and the details fragments were using different sizes, so in the details screen the cached version wasn't being used. If the sizes are equal the cache version is used and the CharacterDrawable is avoided. 

Initially i used different sizes to conserve memory in the list (where the images are smaller), but i'm not sure that is very useful. Using different sizes for the list and the detail duplicates the image in the cache, which can negate the benefit of a smaller cache footprint in the list.

Related to this is the use of the constant `UIUtils.IMAGE_RESIZE_FACTOR`. This was used in the list to further lighten the cache, but it has always been assigned 1.0, so it hasn't got any effect. To see this whole cache working, apply the patch, then play with the constant and you'll see that when it isn't 1.0 you'll get a CharacterDrawable.

If this was your goal, i would prefer this solution for now. I see merit in having the image being handled from the list to the detail, but to achieve something else: make a transition animation in which the image in the list transitions into its place on the detail (one of the standard animations introduced in Lollipop). I implemented a version of it a while ago, but it wasn't perfect so i ditched it, but i plan to review it.

PS. I'll be checking if this change has any impact on the memory footprint, but i appreciate if you could check too.

[art.diff.txt](https://github.com/xbmc/Kore/files/21969/art.diff.txt)

poisdeux:
That is exactly what I am trying to solve. Excellent solution; loading the same size image in the list fragment as is used in the detail fragment.
I do not quite understand what you mean by "there is not really an issue with the album list". It has the same problem as the movie and tv show lists do; when switching to the detailed fragment the same poster image is shown but the detailed fragment first shows the character drawable.
As the size difference in the album list and detailed view is rather big, I think we should either redesign the detailed fragment view using a smaller poster image, or use my suggested placeholder solution.

SyncedSynapse:
Regarding the album list, there are 2 situations: if there's a poster and fanart, both are displayed like in other details fragment, and the poster will hit the cache. If there's no fanart (which is usual), the poster is "promoted" and shown on the fanart imageview. Now, in this second situation, it won't hit the cache because the image is quite big, but it also won't show a character drawable, as that is never shown for images loaded in the fanart imageview.
I'm going to commit the cached solution, extended to the albums and addons, as it is simpler and probably easier to maintain in the future (even though it is relying in a not entirely obvious mechanism).

Issue #103, opened 2015-07-22
Too easy to refresh local list of artists, albums, etc. by mistake


axelsimon:
Hi there,

Context: I have a rather large collection of music, so the Kore's local list of artists takes quite some time to build/refresh.

Issue: While the material design concept of pulling from the top of the list to refresh a list is aesthetically pleasing, and makes a lot of sense for lists of entries that are often updated and are in chronological order (ie: Twitter feed), it doesn't make so much sense in our context: we don't really browse to the top of the list and then want to see if there's anything new.
Worse, it's happened to me twice recently as I was scrolling up and down my list of artists to involuntarily refresh the list. (This also puts Kore in that weird in-between state where it takes a long time to finish, but if I go into an artist and interrupt the update, I end up with missing artists and/or empty artists, so I'm just left waiting)

Proposed solution: I think it would be better to have the refresh option as a dedicated button.
Two options I can think of for that:
- put the refresh icon (set of revolving arrows or whatever) next to the magnifying glass and remote icon, top right of the music screen
- or put refresh icon directly next to the word Music, in the side pull menu, so that 90% of the button goes to the Music screen and 10 % refreshes the content of the Music screen.

Sorry for the lengthy explanation and thanks for considering this issue :)

cheers

axe

gvkt:
+1 Happens to me too very often.

I don't think you need a separate button. The way to pull down as currently implemented is awkward and not best practice in mobile apps. In standard controls that provide this with a pull to update, a flick for scrolling should not activate it, only a drag. In addition, it should show a visual indication of the end being reached and start to "stretch" and not invoke the reload until it is "stretched" some distance. This allows the user to let go and not have it refresh/reload. The current private implementation does neither and so going to the top and not being sure whether one is at the top can easily result in invoking the reload.

poisdeux:
I agree and I created an implementation I am currently still testing and cleaning up. If you are familiar with building an android app from source you can get it at https://github.com/poisdeux/Kore/tree/issue%23103 
Would be nice if you could test it as well. It adds a refresh icon to the toolbar and animates it while a sync process is in progress.

poisdeux:
After playing with the refresh icon in the toolbar I must say I do not really like it. Especially the animation draws too much attention away from the actual app. 
My proposal: adding a Refresh option to the overflow menu and disable the slide refresh option. When user selects refresh from the menu it will start the refresh animation in the list just like with the current slide refresh action.

SyncedSynapse:
[...]

PR #131, opened 2015-11-03
Implemented request #103

poisdeux:
This implements feature request #103 

* Disabled swipe-to-refresh on all fragments extending AbstractListFragment.
* Added refresh menu item in overflow menu to allow user to refresh the
  content in *ListFragments.
* Added refresh menu item to *DetailsFragments for consistency. But
  kept the swipe to refresh action as well.
* Moved showRefreshAnimation() from AbstractListFragment to UIUtils as it is
  now used for detailed fragments as well.
* Added SwipeRefreshLayout to DetailsFragments layouts to be able to show
  the refresh animation
* Introduces new class AbstractDetailsFragment to reduce code duplication.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -53,6 +53,7 @@
         <activity android:name="org.xbmc.kore.ui.AddonsActivity"/>
         <activity android:name="org.xbmc.kore.ui.SettingsActivity"/>
         <activity android:name="org.xbmc.kore.ui.FileActivity"/>
+        <activity android:name="org.xbmc.kore.ui.PVRActivity"/>
         <activity android:name="org.xbmc.kore.ui.AllCastActivity"/>
 
         <!-- Providers -->
--- a/app/src/main/java/org/xbmc/kore/Settings.java
+++ b/app/src/main/java/org/xbmc/kore/Settings.java
@@ -17,6 +17,8 @@ package org.xbmc.kore;
 
 import android.text.format.DateUtils;
 
+import java.util.Set;
+
 /**
  * Interface that contains various constants and the keys for settings stored in shared preferences
  */
@@ -96,4 +98,5 @@ public interface Settings {
     public static final String KEY_PREF_CHECKED_EVENT_SERVER_CONNECTION = "checked_event_server_connection";
     public static final boolean DEFAULT_PREF_CHECKED_EVENT_SERVER_CONNECTION = false;
 
+    public static final String KEY_PREF_NAV_DRAWER_ITEMS = "pref_nav_drawer_items";
 }
--- a/app/src/main/java/org/xbmc/kore/jsonrpc/method/Player.java
+++ b/app/src/main/java/org/xbmc/kore/jsonrpc/method/Player.java
@@ -433,13 +433,17 @@ public class Player {
     public static final class Open extends ApiMethod<String> {
         public final static String METHOD_NAME = "Player.Open";
 
+        public final static String TYPE_PLAYLIST = "playlist",
+                TYPE_CHANNEL = "channel";
+
         /**
          * Start playback of either the playlist with the given ID, a slideshow with the pictures
          * from the given directory or a single file or an item from the database.
-         * @param playlistId
-         * @param position
+         * @param itemType This should always be TYPE_PLAYLIST
+         * @param playlistId Id
+         * @param position Position to start
          */
-        public Open(int playlistId, int position) {
+        public Open(String itemType, int playlistId, int position) {
             super();
             final ObjectNode item = objectMapper.createObjectNode();
             item.put("playlistid", playlistId);
@@ -458,14 +462,22 @@ public class Player {
         }
 
         /**
-         * Select the active player
-         * @param playlistId playlist ID to select
+         * Starts playing a playlist or channel
+         * @param itemType TYPE_PLAYLIST or TYPE_CHANNEL
+         * @param itemId Corresponding ID to open
          */
-        public Open(int playlistId) {
+        public Open(String itemType, int itemId) {
             super();
             final ObjectNode item = objectMapper.createObjectNode();
-            item.put("playlistid", playlistId);
-            addParameterToRequest("item", item);
+            switch (itemType) {
+                case TYPE_PLAYLIST:
+                    item.put("playlistid", itemId);
+                    addParameterToRequest("item", item);
+                    break;
+                case TYPE_CHANNEL:
+                    item.put("channelid", itemId);
+                    addParameterToRequest("item", item);
+            }
         }
 
         @Override
--- a/app/src/main/java/org/xbmc/kore/ui/AbstractDetailsFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AbstractDetailsFragment.java
@@ -1,0 +1,265 @@
+/*
+ * Copyright 2015 Martijn Brekhof. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.xbmc.kore.ui;
+
+import android.content.Intent;
+import android.content.ServiceConnection;
+import android.os.Bundle;
+import android.support.annotation.Nullable;
+import android.support.v4.app.Fragment;
+import android.support.v4.widget.SwipeRefreshLayout;
+import android.view.LayoutInflater;
+import android.view.Menu;
+import android.view.MenuInflater;
+import android.view.MenuItem;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.Toast;
+
+import org.xbmc.kore.R;
+import org.xbmc.kore.host.HostInfo;
+import org.xbmc.kore.host.HostManager;
+import org.xbmc.kore.jsonrpc.ApiException;
+import org.xbmc.kore.jsonrpc.event.MediaSyncEvent;
+import org.xbmc.kore.service.LibrarySyncService;
+import org.xbmc.kore.service.SyncUtils;
+import org.xbmc.kore.utils.LogUtils;
+import org.xbmc.kore.utils.UIUtils;
+
+import de.greenrobot.event.EventBus;
+
+abstract public class AbstractDetailsFragment extends Fragment
+        implements SwipeRefreshLayout.OnRefreshListener,
+        SyncUtils.OnServiceListener {
+    private static final String TAG = LogUtils.makeLogTag(AbstractDetailsFragment.class);
+
+    private HostManager hostManager;
+    private HostInfo hostInfo;
+    private EventBus bus;
+    private String syncType;
+
+    private SwipeRefreshLayout swipeRefreshLayout;
+
+    private ServiceConnection serviceConnection;
+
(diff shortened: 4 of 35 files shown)


SyncedSynapse:
Didn't thoroughly check the code, but from a first look i have the following comments:

1. I don't agree with removing the swipe to refresh from the lists. I admit tinkering with the swipe distance, but not remove it nor make it overly difficult to trigger. It's a standard pattern on Android, it is in production since the beginning, users are used to it and removing it will cause unnecessary confusion.
2. I don't object to adding a "refresh" option on the overflow menu (as long as it is hidden). I'm not in favor of it because it's a duplication of functionality, but i'm not against it (as long as the overflow menu doesn't have many options. If in the future it gets more "crowded", the refresh option is a candidate to go).
3. ItemId shouldn't be on the refactored abstract details fragment. That info is specific to each concrete fragment, and isn't abstract. The fact that in most cases there is only one id doesn't matter, as conceptually that info should be maintained by the fragment that knows what it means and what to do with it. In fact, in the episode details fragment there are 2 item ids (tvshowid and episodeid), which shows that you can't trust that there is only one. In the abstract fragment you don't really need the itemid, except in the sync function, which can call a virtual method overriden in each fragment to fill the necessary info (i might have missed it somewhere else).
4. The album details fragment didn't include a refresh on purpose. The thing is, there isn't a way to sync a single album (or a single artist), and music collections can be big, so triggering a full music sync (potentially slow) when the user is expecting just an album sync (supposedly quick)  doesn't feel right. Furthermore, the need for a single album sync isn't really that great because its info isn't volatile: in the tvshow and movies there is the "seen" bit which is volatile and it's useful to keep synced. Now, you could say "what about music videos, they are the same and also have a sync"... alas, i never claimed i was consistent :)... The thing is that i believe that a music video sync doesn't take as long as the music library, so i kept it.
By the way, there isn't a single album sync implemented because i was lazy - it's not as straightforward as the others (foreign keys on artists/genres) - and because i really think it's not as necessary as single tvshow/movie sync.
Summing up, i'm not convinced we should have a full music library sync triggered on an album details view.

Now, nitpicking:
1. I understand that this refactoring is easier to do on one big lump, but it's really difficult to review this many changes, and it is much easier for bugs to slip by. In the future i would appreciate smaller commits.
2. Try changing your settings to use spaces (4) instead of tabs. I don't want to enter in this holy war, but the fact of the matter is that all the existing code is using spaces and it should be consistent. Likewise, there are some settings regarding parameter alignment that you could review, as i saw some diffs that were only due to whitespace.
3. Just make the hostConnection/HostInfo in the abstractDetailsFragment and friends protected and drops the getters. Or at least make the getters protected, no need for them to be public.

poisdeux:
On Wed, Nov 4, 2015 at 12:29 AM, Synced Synapse <notifications@github.com>
wrote:

> Didn't thoroughly check the code, but from a first look i have the
> following comments:
>
>    1. I don't agree with removing the swipe to refresh from the lists. I
>    admit tinkering with the swipe distance, but not remove it nor make it
>    overly difficult to trigger. It's a standard pattern on Android, it is in
>    production since the beginning, users are used to it and removing it will
>    cause unnecessary confusion.
>
> Ok, I will add it again.

I agree that the swipe to refresh is a standard pattern on Android and
removing it will probably confuse users. But I do not think that the way
Kore applies it, is standard. If you take a look at google's design
guidelines:

https://www.google.com/design/spec/patterns/swipe-to-refresh.html

you will see they specifically target lists where the most recent content
appears at the beginning of the list. This was also noted by axelsimon in
#103.

Another reason why I personnally wanted the refresh menu item, is that with
big lists it is very inconvenient to scroll back to the top of the list to
start the refresh when I find out a new artist is not yet available on Kore.


>    1. I don't object to adding a "refresh" option on the overflow menu
>    (as long as it is hidden). I'm not in favor of it because it's a
>    duplication of functionality, but i'm not against it (as long as the
>    overflow menu doesn't have many options. If in the future it gets more
>    "crowded", the refresh option is a candidate to go).
>
>
>    1. ItemId shouldn't be on the refactored abstract details fragment.
>    That info is specific to each concrete fragment, and isn't abstract. The
>    fact that in most cases there is only one id doesn't matter, as
>    conceptually that info should be maintained by the fragment that knows what
>    it means and what to do with it. In fact, in the episode details fragment
>    there are 2 item ids (tvshowid and episodeid), which shows that you can't
>    trust that there is only one. In the abstract fragment you don't really
>    need the itemid, except in the sync function, which can call a virtual
>    method overriden in each fragment to fill the necessary info (i might have
>    missed it somewhere else).
>
> Ok, I will change it.


>
>    1. The album details fragment didn't include a refresh on purpose. The
>    thing is, there isn't a way to sync a single album (or a single artist),
>    and music collections can be big, so triggering a full music sync
>    (potentially slow) when the user is expecting just an album sync
>    (supposedly quick) doesn't feel right. Furthermore, the need for a single
>    album sync isn't really that great because its info isn't volatile: in the
>    tvshow and movies there is the "seen" bit which is volatile and it's useful
>    to keep synced. Now, you could say "what about music videos, they are the
>    same and also have a sync"... alas, i never claimed i was consistent :)...
>    The thing is that i believe that a music video sync doesn't take as long as
>    the music library, so i kept it. By the way, there isn't a single album
>    sync implemented because i was lazy - it's not as straightforward as the
>    others (foreign keys on artists/genres) - and because i really think it's
>    not as necessary as single tvshow/movie sync. Summing up, i'm not convinced
>    we should have a full music library sync triggered on an album details view.
>
> I wasn't too sure about this either. I mainly added it for consistency. I
will change it back.


>
> Now, nitpicking:
> 1. I understand that this refactoring is easier to do on one big lump, but
> it's really difficult to review this many changes, and it is much easier
> for bugs to slip by. In the future i would appreciate smaller commits.
>
Do you mean you would prefer the refactoring of the code in a separate PR
and use another PR to add the refresh option, in this case?

2. Try changing your settings to use spaces (4) instead of tabs. I don't
> want to enter in this holy war, but the fact of the matter is that all the
> existing code is using spaces and it should be consistent. Likewise, there
> are some settings regarding parameter alignment that you could review, as i
> saw some diffs that were only due to whitespace.
>
I prefer using spaces as well. However, I thought I was save using Android
Studio with its default settings. Do you have any pointers in what needs to
be changed regarding parameter alignment?

> 3. Just make the hostConnection/HostInfo in the abstractDetailsFragment
> and friends protected and drops the getters. Or at least make the getters
> protected, no need for them to be public.
>
Didn't notice that. Sorry about that.

axelsimon:
On 04/11/15 14:14, Martijn Brekhof wrote:
> I agree that the swipe to refresh is a standard pattern on Android and
> removing it will probably confuse users. But I do not think that the way
> Kore applies it, is standard. If you take a look at google's design
> guidelines:
>
> https://www.google.com/design/spec/patterns/swipe-to-refresh.html
>
> you will see they specifically target lists where the most recent content
> appears at the beginning of the list. This was also noted by axelsimon in
> #103.
Quick suggestion here, which may or may not be smart, but if the swipe
to refresh action at the top of the library should be phased out, a
temporary solution to help users understand where the “refresh library”
action went would be to display a text on swipe.

Ie: you pull the from the top, and rather than seeing the refresh icon
appear, you see a message telling you the action has been moved to the
overflow menu.

Then in time, you just remove that, once people have adjusted to where
the action now is.

I realise however it might not be (easily) possible to show something
else on pull to refresh.

Cheers

axel

SyncedSynapse:
@poisdeux I believe you're reading that guideline too strictly. Are you defending that swipe to refresh can only be used in lists with items that have some kind of timestamp and are sorted in chronological order? That might be the ideal scenario, but i don't think is the only one. Take a look at the Drive app, or the Docs/Spreadsheets and you'll see the same pattern used here, and the items aren't arranged in chronological order.

Anyway, @axelsimon swipe to refresh will stay, though we can tweak the swipe distance to make it harder to trigger.

@poisdeux. Smaller commits would be nice. As for the current one, if it's not too much trouble, spliting it would be better, but it can go in as is.

In Android Studio take a look at Settings->Editor->Code Style. Don't use tabs, set the tab size and indent to 4 spaces, and the continuation indent to 8 (humm, i had something set to 6, but that doesn't make sense, i'll change it). Also keep an eye on spaces on expressions. It should be `if ((expression1) || (expression2))`, not `if ( ( expression 1 ) ) || ( ( expression2 ) )` (spotted this somewhere, but it's not consistent).

SyncedSynapse:
@poisdeux Sorry i made a mistake. Don't get rid of the getters (hostconnection, etc). Just make them protected. I wasn't thinking straight.

One other thing: i'll probably change the name of the Abstract*Fragment sometime in the future. Abstract is a really general name and not very informative. These fragments aren't really that general, they can't be used as base classes for the Addons list/details for instance, because the sync functions don't apply. They should probably be named something like SyncListFragment/SyncDetailsFragment. If you have any suggestions they are welcome.

poisdeux:
On Wed, Nov 4, 2015 at 8:44 PM, Synced Synapse <notifications@github.com>
wrote:

> @poisdeux <https://github.com/poisdeux> I believe you're reading that
> guideline too strictly. Are you defending that swipe to refresh can only be
> used in lists with items that have some kind of timestamp and are sorted in
> chronological order? That might be the ideal scenario, but i don't think is
> the only one. Take a look at the Drive app, or the Docs/Spreadsheets and
> you'll see the same pattern used here, and the items aren't arranged in
> chronological order.
>
I agree that swipe to refresh can be perfectly used in other scenarios as
well. I was merely trying to point out that I do not consider those
scenarios standard.


> Anyway, @axelsimon <https://github.com/axelsimon> swipe to refresh will
> stay, though we can tweak the swipe distance to make it harder to trigger.
>
> @poisdeux <https://github.com/poisdeux>. Smaller commits would be nice.
> As for the current one, if it's not too much trouble, spliting it would be
> better, but it can go in as is.
>
Ok, I will split it in three PRs:

1. Refactoring: introducing AbstractDetailsFragment
2. Add binding to LibrarySyncService for detail fragments, just as we did
for list fragments.
3. Add the refresh menu item

I will change this PR to represent 3 when 1 and 2 have been merged.

In Android Studio take a look at Settings->Editor->Code Style. Don't use
> tabs, set the tab size and indent to 4 spaces, and the continuation indent
> to 8 (humm, i had something set to 6, but that doesn't make sense, i'll
> change it). Also keep an eye
>
That's actually the default.

What I find annoying is that auto-indent lines seems to add extra spaces.
See for instance:

https://github.com/xbmc/Kore/compare/master...poisdeux:refactor/abstractdetailsfragment?expand=1#diff-511f10a905a2bacb46e3e47d187e1722L562

This makes the PR bigger than it should be. Do you have any idea how I can
prevent this?

SyncedSynapse:
Regarding the spliting, if it's not too much trouble i agree.

Regarding the auto-indent, i justo took a look, and my Android Studio is doing the same. I believe that the default changed in one of the latest versions and i didn't realize. Anyway, i have the following options set in Settings->Code Style->Java->Wrapping and Braces:
1. Method Call Arguments -> Align when multiline
2. Chained Method Calls -> Align when multiline

But given that these aren't the default, ignore my comment regarding the white spaces. If anything it's me that should revert to the default.

Nevertheless, you should check your tab settings, as new files are getting generated with tabs or 8 spaces indent.

What do you mean by "makes the PR bigger"?

poisdeux:
On Fri, Nov 6, 2015 at 4:16 PM, Synced Synapse <notifications@github.com>
wrote:

>
> What do you mean by "makes the PR bigger"?
>
The diffs of my Pull Requests contain a lot of whitespace diffs which are
not functional changes. So they should not actually be included. In other
words, it makes the DIFF of the PR bigger than it should be.
The two settings you mentioned seemed to fix the problem. As these are not
default, I think we should mention them in the README.md

SyncedSynapse:
Hey @poisdeux, i'm going to push some changes that add a simple PVR support. AFAICT it shouldn't conflict with your changes as they don't seem to touch the same files, but i'm going to upgrade gradle to the latest version, so you might want to check your builds after.

poisdeux:
I am going to fix this PR now to represent adding the refresh menu item. I will rebase it on master so if you have a local copy of this branch please remove it.

PR #136, opened 2015-11-12
This implements feature request #103

poisdeux:
Added refresh menu item in overflow menu to allow user to refresh the content in *ListFragments and *DetailsFragments.


SyncedSynapse:
This conflicts with #135 . Can you check it please.

I also tried applying them in reverse order, but the conflict is also there.

poisdeux:
Oops, sorry, I only checked it against master. I should have checked it against #135 as well. Should merge cleanly now.

SyncedSynapse:
Ok. Any reason for removing the `@Override` from the `onStop` method on `AbstractDetailsFragment`?

poisdeux:
Nope, *sigh* that's what I get when I do things inbetween...

poisdeux:
fixed it.

PR #138, opened 2015-11-20
Implemented shared element transition for movie posters

poisdeux:
* Refactored OnMovieSelectedListener interface to keep argument list small.
* Changed fragment transition animation to make shared element transition
  smooth and nicely integrated with the whole scene animation.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
@@ -193,6 +193,8 @@ public class MovieDetailsFragment extends AbstractDetailsFragment
 //        UIUtils.setPaddingForSystemBars(getActivity(), mediaPanel, false, false, true);
 //        mediaPanel.setClipToPadding(false);
 
+        //Disable download button until loader has finished and download info is available
+        downloadButton.setEnabled(false);
         return root;
     }
 
@@ -539,7 +541,7 @@ public class MovieDetailsFragment extends AbstractDetailsFragment
                 movieTitle, cursor.getString(MovieDetailsQuery.FILE));
 
         // Check if downloaded file exists
-        downloadButton.setVisibility(View.VISIBLE);
+        downloadButton.setEnabled(true);
         if (movieDownloadInfo.downloadFileExists()) {
             Resources.Theme theme = getActivity().getTheme();
             TypedArray styledAttributes = theme.obtainStyledAttributes(new int[]{
--- a/app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
@@ -81,7 +81,10 @@ public class MovieDetailsFragment extends AbstractDetailsFragment
     public static final String BUNDLE_KEY_MOVIEPLOT = "movie_plot";
     public static final String BUNDLE_KEY_MOVIEID = "movie_id";
     public static final String POSTER_TRANS_NAME = "POSTER_TRANS_NAME";
-
+    public static final String BUNDLE_KEY_MOVIEGENRES = "movie_genres";
+    public static final String BUNDLE_KEY_MOVIEYEAR = "movie_year";
+    public static final String BUNDLE_KEY_MOVIERUNTIME = "movie_runtime";
+    public static final String BUNDLE_KEY_MOVIERATING = "movie_rating";
     // Loader IDs
     private static final int LOADER_MOVIE = 0,
             LOADER_CAST = 1;
@@ -143,7 +146,10 @@ public class MovieDetailsFragment extends AbstractDetailsFragment
         args.putInt(BUNDLE_KEY_MOVIEID, vh.movieId);
         args.putString(BUNDLE_KEY_MOVIETITLE, vh.movieTitle);
         args.putString(BUNDLE_KEY_MOVIEPLOT, vh.movieTagline);
-
+        args.putString(BUNDLE_KEY_MOVIEGENRES, vh.movieGenres);
+        args.putInt(BUNDLE_KEY_MOVIEYEAR, vh.movieYear);
+        args.putInt(BUNDLE_KEY_MOVIERUNTIME, vh.movieRuntime);
+        args.putDouble(BUNDLE_KEY_MOVIERATING, vh.movieRating);
         if( Utils.isLollipopOrLater()) {
             args.putString(POSTER_TRANS_NAME, vh.artView.getTransitionName());
         }
@@ -155,7 +161,8 @@ public class MovieDetailsFragment extends AbstractDetailsFragment
     @TargetApi(21)
     @Override
     protected View createView(LayoutInflater inflater, ViewGroup container) {
-        movieId = getArguments().getInt(BUNDLE_KEY_MOVIEID, -1);
+        Bundle bundle = getArguments();
+        movieId = bundle.getInt(BUNDLE_KEY_MOVIEID, -1);
 
         if (movieId == -1) {
             // There's nothing to show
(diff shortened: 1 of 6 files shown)


SyncedSynapse:
This is something that i wanted to do, and i did a prototype, but it wasn't perfect so i didn't include it.

This is better, but it's still not seamless. The main issue is that the "Title" and "Undertitle" views can take more than one line if the text is long, and in those cases, there's a jump in the animation of the art view. It's pretty jarring, and i don't know what the correct solution is. Maybe the only solution is to make the Title and Undertitle views have a fixed number of lines, but then, longer titles won't show up in full...

Additionally, the use of shared elements transitions seems to alter some stuff inside Android, causing it to show up some views in their uninitialized state. The "Download" button for instance isn't present in the animation, and is abruptly shown up when the animation ends. This is only an issue when we add shared elements to the animation (meaning that in the previous version, it somehow was able to run the code that determined whether the button should show up before the animation started). This is obviously an issue with the animation framework, but i think that we should work around these issues.

To see what i'm talking about, go into developer settings, and slow down all the animations to 5x. The movie score/date/detail also suffer in this version.

And btw, add `@TargetAPI(21)` annotation to the relevant methods, otherwise `gradlew build` fails.

poisdeux:
I also noticed the ugly update effect after the animation ended. But I wanted to know what you thought of shared element transitions before I put in more effort to fix it. It seems to run smooth now :). But I am not sure if I tested all cases.

I believe the problem was that, when android measures the layout for the animation, the textviews are empty. When the loader finishes, the textviews are updated and the layout is revalidated. 
I could have postponed the transition to the point when the loader finishes and we have all needed info. But this may cause unacceptable lag. Instead, I fixed this by passing all info we already have in the MovieListFragment to the MovieDetailsFragment. Thus, allowing us to fill the textviews in the onCreateView after which android measures the layout.

app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
```diff
@@ -44,6 +44,8 @@
 
 import com.melnykov.fab.FloatingActionButton;
 import com.melnykov.fab.ObservableScrollView;
+import com.squareup.okhttp.internal.Util;
```
SyncedSynapse: Can be removed

app/src/main/java/org/xbmc/kore/ui/MoviesActivity.java
```diff
@@ -17,16 +17,29 @@
...
+import android.util.Log;
+import android.view.Gravity;
 import android.view.Menu;
 import android.view.MenuItem;
+import android.view.View;
 import android.view.Window;
+import android.widget.ImageView;
+import android.widget.TextView;
 
+import org.w3c.dom.Text;
```
SyncedSynapse: Lots of unused imports...

app/src/main/java/org/xbmc/kore/ui/MoviesActivity.java
```diff
@@ -67,10 +80,15 @@ protected void onCreate(Bundle savedInstanceState) {
 
             // Setup animations
             if (Utils.isLollipopOrLater()) {
-                movieListFragment.setExitTransition(null);
+                //Fade added to prevent shared element from disappearing very shortly at the start of the transition.
+                Fade fade = new Fade();
+                fade.setDuration(50);
+                movieListFragment.setExitTransition(fade);
```
SyncedSynapse: Can't the default android.R.transition.fade be used? It's used in the reenter transition (line bellow), so it could be inflated once, and be used in these 2 places.

app/src/main/java/org/xbmc/kore/ui/MoviesActivity.java
```diff
@@ -168,31 +186,42 @@ private void setupActionBar(String movieTitle) {
...
+
             movieDetailsFragment.setEnterTransition(TransitionInflater
                     .from(this)
                     .inflateTransition(R.transition.media_details));
-            movieDetailsFragment.setReturnTransition(null);
+
+            movieDetailsFragment.setSharedElementReturnTransition(TransitionInflater.from(
+                    this).inflateTransition(R.transition.change_image));
+            movieDetailsFragment.setSharedElementEnterTransition(TransitionInflater.from(
+                    this).inflateTransition(R.transition.change_image));
```
SyncedSynapse: Simple refactor on the above 2 instructions:
`
            Transition changeImageTransition = TransitionInflater.from(this).inflateTransition(R.transition.change_image);
 movieDetailsFragment.setSharedElementReturnTransition(changeImageTransition);
 movieDetailsFragment.setSharedElementEnterTransition(changeImageTransition);
`

app/src/main/java/org/xbmc/kore/ui/MoviesActivity.java
```diff
@@ -168,31 +186,42 @@ private void setupActionBar(String movieTitle) {
...
 
         // Set up transitions
         if (Utils.isLollipopOrLater()) {
+            MovieDetailsFragment movieDetailsFragment = MovieDetailsFragment.newInstance(vh);
+            FragmentTransaction fragTrans = getSupportFragmentManager().beginTransaction();
+
             movieDetailsFragment.setEnterTransition(TransitionInflater
                     .from(this)
                     .inflateTransition(R.transition.media_details));
-            movieDetailsFragment.setReturnTransition(null);
```
SyncedSynapse: Not to be removed. 
Removing this causes the details fragment to slide down on the return transition, which is very distracting and looks odd, with the list fragment fading into view.
I think that having null for the return transition gives the best effect, but a fade out also works (though it somewhat interferes visually with the fade in of the list fragment).

app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
```diff
@@ -209,10 +243,19 @@ public void onActivityCreated (Bundle savedInstanceState) {
...
         // Force the exit view to invisible
         exitTransitionView.setVisibility(View.INVISIBLE);
+        //As we make mediaPoster invisible in onStop() we need to make it visible here.
+        mediaPoster.setVisibility(View.VISIBLE);
         super.onResume();
     }
 
     @Override
+    public void onStop() {
+        //TODO: for some reason poster is included in the bottom slide animation, by making it invisible it is not noticeable for the user
```
SyncedSynapse: Remove TODO, just leave the comment.

app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
```diff
@@ -508,7 +546,7 @@ private void displayMovieDetails(Cursor cursor) {
                 movieTitle, cursor.getString(MovieDetailsQuery.FILE));
 
         // Check if downloaded file exists
-        downloadButton.setVisibility(View.VISIBLE);
+        downloadButton.setEnabled(true);
```
SyncedSynapse: Can't remember why the downloadButton was `gone` and only turned visible here... Can't think of a good justification for that, maybe it was just an oversight.
If that was the case, there's no need to do the disable/enable of it. Just leaving it be always on should work fine.

SyncedSynapse:
Looking good.
I've added some comments and i would appreciate if you could review them.

Apart from that, i just don't like using the `ViewHolder` to pass info to `onMovieSelected` and from there to the details fragment `newInstance`... I understand that it's handy, but it's an ugly hack, that is making the 2 fragments and the activity extremely coupled. Moreover, the viewholder is an implementation detail, it shouldn't be used on interfaces.
On the other hand i'm not sure if the alternatives are better: creating a class specifically for passing the info feels wasteful, and passing all the values explicitly is kind of too much verbose... 
Let it be for now, though i might change it in the future.

There's one more thing that would look good: the fab appearing on a circular reveal after the other animations ended. The fab has a method `show()` which might do this, but i'm not sure when to call it (when do we know that the shared element transition ended). Anyway, it's something for the future.

app/src/main/java/org/xbmc/kore/ui/MoviesActivity.java
poisdeux: You will see that the shared element will still be visible in the fading view. By making the fade shorter it is not noticeable. The trick I used in MovieDetailsFragment by making the poster invisible does not seem to work here.

app/src/main/java/org/xbmc/kore/ui/MoviesActivity.java
poisdeux: I thought the slide down was more consistent. I'll change it back.

app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
poisdeux: Ok, I'll just leave it on then.

app/src/main/java/org/xbmc/kore/ui/MoviesActivity.java
poisdeux: done

app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
poisdeux: fixed

app/src/main/java/org/xbmc/kore/ui/MoviesActivity.java
poisdeux: fixed

PR #143, opened 2015-12-04
Implemented shared element transitions for albums

poisdeux:
I removed using the poster as fanart if no fanart is available.
The reason is that the shared element transition is not very smooth
when the aspect ratio of the two image states is not the same.
Going from AlbumListFragment to AlbumDetailsFragment the poster
is resized correctly to the fanart size. However, returning to the
AlbumListFragment from the AlbumDetailsFragment the poster is
abruptly resized to the square aspect ratio needed for the list
fragment.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/AlbumDetailsFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AlbumDetailsFragment.java
@@ -59,6 +59,7 @@ import org.xbmc.kore.provider.MediaContract;
 import org.xbmc.kore.utils.FileDownloadHelper;
 import org.xbmc.kore.utils.LogUtils;
 import org.xbmc.kore.utils.UIUtils;
+import org.xbmc.kore.utils.Utils;
 
 import java.io.File;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


SyncedSynapse:
If you do a `gredlew build` it will fail with a lint error. You're missing a TargetApi Annotation.

Regarding the removal of using the poster as fanart: i understand your rationale, but for instance in my library i don't have fanart on most albums (i don't keep it neat), so it will look odd.
What about this: keep it as is in this PR (poster always showing up to enable the shared animation), and if there's no fanart, present the poster there. This will "duplicate" the poster image, but given that they have different dimensions, it might not look bad (don't know, haven't tested it).

poisdeux:
I think your proposed solution looks good, but I do not like GUI smeg that does not really add anything. 
I was thinking about another solution which involves removing the mediaArt completely when no fanart is available and show the album, title, and undertitle at the top of the screen. However, the UX might be less consistent in this case. What do you think?

SyncedSynapse:
Looks good now. I'm just going to do a slight change, which is not load a character avatar on the media art, as if both imageviews are loaded with character avatars they look odd.

Regarding not showing the mediaArt imageView, i'm not sure... On one hand, i agree that it might make sense, but i'm not sure that it would look (as) good (as now) .

PR #144, opened 2015-12-05
Added replace file pref

ghost:


SyncedSynapse:
What is this supposed to do? Is it supposed to be part of something bigger that you didn't PR yet?

PR #151, opened 2015-12-12
Only use IPv4 for EventServer connections

ThiefMaster:


app/src/main/java/org/xbmc/kore/utils/Utils.java
```diff
@@ -154,4 +157,19 @@ public static Bitmap drawableToBitmap (Drawable drawable, int width, int height)
 
         return bitmap;
     }
+
+    /**
+     * Gets an IPv4 address from a host name
+     * @param host The host to look up
+     * @return Inet4Address
+     */
+    public static Inet4Address getInet4AddressByName(String host) throws UnknownHostException {
```
SyncedSynapse: This would be better located in NetUtils.java.

ThiefMaster: Ah, didn't spot that file. I'll update the PR.

PR #152, opened 2015-12-12
Add setting to use seek steps instead of speed change

ThiefMaster:


SyncedSynapse:
Sorry for not replying earlier, but i'm thorn on this one...

I understand that you find it useful, but i'm not sure that it is something that the majority of users will want or even understand what it is.

I know it's optional, but IMO that (for itself) is not a valid justification for including it. The issue is that it is kind of confusing to explain, and like i said, i'm not sure users will use it, so it shouldn't be "polluting" the settings screen. And the fact that the settings already have some stuff that shouldn't be there isn't a reason for including more.

If kore had something like advancedsettings.xml i would have nothing against this neing there, but including it on the main settings page is something i'm reticent.

I can change my opinion though, so i'll leave this open.

ThiefMaster:
Hm, have you considered making that screen fully configurable? I.e. let the user choose which buttons he wants there? Android itself (or at least CM) does that for the global dropdown thingy which is kind of nice - everyone can put what he wants / uses most there.

SyncedSynapse:
A settings settings... not a fan.

Including an advanced settings submenu is something that might make more sense, but i'm still not convinced that this option is useful for a lot of users (personal opinion, might be wrong obviously).

poisdeux:
Can this be closed?

SyncedSynapse:
Like i said, i see the usefulness of this, and that's why i didn't close it.

@poisdeux what's your opinion?

poisdeux:
I actually never use the seek buttons but use the progress bar to rewind or fast forward.
So I'm actually in favour of removing the seek buttons completely.
Isn't a better option to add a seekbar to the notification?

SyncedSynapse:
No, a seekbar on the notification doesn't fit android guidelines. I strongly oppose that.

poisdeux:
Ah OK.
Well I don't mind making it optional to change the seek button's behaviour from speed to steps.
I do think that the change should affect all seek buttons and not just those in the notification.
Is there a different icon we could use to indicate forwarding/rewinding using a step size?

ThiefMaster:
The seekbar is not very useful to skip e.g. over the intro that's the same on every episode of a series. Usually you quickly remember how long it is (20 or 30s is common), so in my case I hit forward twice and skip exactly over the intro. With a seekbar this would be much more cumbersome; same with fast-forward.

In the main UI you can already use the left/right buttons (at least if EventServer is enabled) to seek a step forward/backward, which is why I didn't touch this part.

poisdeux:
I do think we should change the icon if we change its functionality.
How about https://materialdesignicons.com/icon/skip-forward to indicate a step instead of speed change?

SyncedSynapse:
This PR already changes the icons correctly (haven't tested it, but it's in the code).

@ThiefMaster Sorry for the delay (and the extra work...), i'll merge this if you rebase it.

SyncedSynapse:
This looks good, though i want to do some more tests later.

I just don't like the icon (the chevron) that is used for the steps. Poisdeux's suggestions seems better https://materialdesignicons.com/icon/skip-forward

@ThiefMaster Do you want to do it? I can merge it as is and change it later.

ThiefMaster:
> I can merge it as is and change it later.

sounds good to me!

PR #153, opened 2015-12-14
Implemented shared element transitions for addons

poisdeux:
Removed getting details in AddonDetailsFragment as the
AddonListFragment already provides all information required
in the AddonDetailsFragment. Therefore, we do not need to
retrieve the same information in the AddonDetailsFragment as well.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/AddonDetailsFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AddonDetailsFragment.java
@@ -63,6 +63,7 @@ public class AddonDetailsFragment extends Fragment {
     public static final String BUNDLE_KEY_DESCRIPTION = "description";
     public static final String BUNDLE_KEY_FANART = "fanart";
     public static final String BUNDLE_KEY_POSTER = "poster";
+    public static final String BUNDLE_KEY_ENABLED = "enabled";
 
     private HostManager hostManager;
     private HostInfo hostInfo;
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/xbmc/kore/ui/AddonDetailsFragment.java
```diff
@@ -225,70 +235,39 @@ public void onSuccess(String result) {
...
+        int posterHeight = resources.getDimensionPixelOffset(R.dimen.addondetail_poster_height);
+
+        UIUtils.loadImageIntoImageview(hostManager,
+                                       TextUtils.isEmpty(fanart)? poster : fanart,
+                                       mediaArt, artWidth, artHeight);
+        UIUtils.loadImageIntoImageview(hostManager,
+                                       poster,
+                                       mediaPoster, posterWidth, posterHeight);
 
-        setupEnableButton(addonDetails.enabled);
```
SyncedSynapse: You can't remove this call. It is used to show if the addon is enabled or not when the user enters the screen, so it is needed.

poisdeux:
Oops... Fixed now. The "enabled" boolean now also gets passed from the list fragment to the details fragment.

app/src/main/java/org/xbmc/kore/ui/AddonDetailsFragment.java
poisdeux: Moved it to the onCreateView method

SyncedSynapse:
I'll merge it as is, as i want to build a release, but this introduces a small bug: Select an addon on the list, enabe/disabe it, return to the list, select the same addon, the state (enabled/disabled) of the addon will be incorrect because it is getting that state from the list which was updated before enabling/disabling it.

Keeping the GetAddonDetails call on the details fragment would prevent this.

poisdeux:
So we only need to get the enabled state in the details fragment?

SyncedSynapse:
yes.

PR #155, opened 2015-12-16
Show Context Menu on "Select" long press.

RaafatAkkad:


SyncedSynapse:
There's a specific button for Context which already does this, it shouldn't be duplicated in another button.

Furthermore, the possibility of associating 2 actions with one button is something to be used sparingly and with care (as it hurts discoverability), and it certainly shouldn't be used in one of the "main" buttons (select + d-pad).

RaafatAkkad:
I agree with keeping the UI simple.

The patch replicates the long pressing functionality of Kodi (Long press Enter) and various remotes such as Yatse. Also, if you're constantly switching between Kore & a remote control (without a context button) (http://kodi.tv/kodi-16-alpha-3/) it will take some time to get used to it.

SyncedSynapse:
I see your point, but nevertheless i don't think that a long press should be added to the select button.

What might be worth considering is changing the select button to use the EventServer. And in the process review the way it is sending button presses to the EventServer. Then Select might send a long press, through EventServer, and that gets processed by Kodi as a normal remote. But i need to think better about this.

PR #175, opened 2016-01-05
Clear back stack when clicking FAB in movie/episode details

holahmeds:
Earlier, clicking the FAB would start a new instance of RemoteActivity so the back stack would look like this:
RemoteActivity
MovieDetailsFragment
MoviesActivity
RemoteActivity
...

Clicking on the back button would return you to MovieDetailsFragment which is unintuitive.

Now, ckicking on the FAB will clear the back stack up to the last instance of RemoteActity, so clicking on the back button would either close the app or return you to the activity that started RemoteActivity.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/utils/UIUtils.java
+++ b/app/src/main/java/org/xbmc/kore/utils/UIUtils.java
@@ -418,7 +418,8 @@ public class UIUtils {
     public static void switchToRemoteWithAnimation(final Context context,
                                                    int centerX, int centerY,
                                                    final View exitTransitionView) {
-        final Intent launchIntent = new Intent(context, RemoteActivity.class);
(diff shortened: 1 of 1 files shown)


SyncedSynapse:
Sorry, i don't agree with this. I think the user expects to be able to hit back and return to the details fragment after clicking on the FAB.

holahmeds:
It's not what I expected (hence the pull request).

The way I see it, the details fragment is like a sub menu/screen. Clicking on the FAB would close the screen, i.e. send you _back_ to the remote, and not open a new remote.

SyncedSynapse:
Guess we have different viewpoints. The way i see it, clicking on the fab can (optionally) take you to the remote, but you can get back (and it is useful, as the user might want to play another thing that was available on the list fragment).

holahmeds:
Very well, you may close the pull request then.

PR #177, opened 2016-01-08
Implemented restoring state in TVShowEpisodeListFragment

poisdeux:


app/src/main/java/org/xbmc/kore/ui/TVShowEpisodeListFragment.java
```diff
@@ -254,49 +297,36 @@ protected void onSyncProcessEnded(MediaSyncEvent event) {
...
     public void onLoadFinished(Loader<Cursor> cursorLoader, Cursor cursor) {
-        LogUtils.LOGD(TAG, "onLoadFinished, Loader id: " + cursorLoader.getId() + ". Rows: " +
-                cursor.getCount());
         switch (cursorLoader.getId()) {
             case LOADER_SEASONS:
                 adapter.setGroupCursor(cursor);
-                if (cursor.getCount() == 1) {
-                    // Force collapse and expand the group, to force a reload of the episodes
-                    // cursor, otherwise if it is already expanded it won't reload (and won't
-                    // apply the filters to the episodes list)
```
poisdeux: What filters are meant here?

SyncedSynapse: the hide watched filter, i believe

SyncedSynapse:
Feel kind of bad here, because, there's a lot of work in this, but i plan to redesign the tvshow details/episodes fragments soon, probably removing the groupview...

Anyway, i'll review and test this tomorrow.

poisdeux:
Are you talking about the ExpandableListView? If so, I think it is a horrible UI. I would prefer something similar to the albumlist -> albumdetails design. Also, using the season poster in the list as poster in the detailsfragment, we can add a shared element transition to make the UX more consistent.

SyncedSynapse:
Yes, i also don't like the ExpandableListView. I did it that way to get it done quickly, but i planned on revisiting it, but never have...

I plan to have only one fragment, with the tv show details, then the next unwatched episode (which can be directly started), then the seasons list (which when clicked takes you to the episode list), then the cast.

I believe this solves the most common use-case (starting the next episode) and will look better.

PR #178, opened 2016-01-08
Add option to play vimeo URLs on Kodi

tomerf:


SyncedSynapse:
Thanks.
I'm not very familiar with vimeo, but i noticed that sometimes URLs are of the form: http://vimeo.com/12332?from=outro_local

The regexp in this PR fails with those URLs because of the ?... part. Not sure if relevant.

PR #180, opened 2016-01-14
Implemented changing the position of items in the current playlist

poisdeux:
This enables users to reorder the current playlist by long pressing a
list item and drag it to a different list position.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/jsonrpc/method/Playlist.java
+++ b/app/src/main/java/org/xbmc/kore/jsonrpc/method/Playlist.java
@@ -170,4 +170,27 @@ public class Playlist {
             return jsonObject.get(RESULT_NODE).textValue();
         }
     }
+
+
+    public static final class Insert extends ApiMethod<String> {
+        public final static String METHOD_NAME = "Playlist.Insert";
+
+        /**
+         * Add item(s) to playlist
+         */
+        public Insert(int playlistId, int position, PlaylistType.Item item) {
+            super();
+            addParameterToRequest("playlistid", playlistId);
+            addParameterToRequest("position", position);
+            addParameterToRequest("item", item);
+        }
+
+        @Override
+        public String getMethodName() { return METHOD_NAME; }
+
+        @Override
+        public String resultFromJson(ObjectNode jsonObject) throws ApiException {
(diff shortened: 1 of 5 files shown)


poisdeux:
While changing the PlaylistFragment.java file I noticed the method switchToPanel which use I do not understand:
https://github.com/xbmc/Kore/blob/master/app/src/main/java/org/xbmc/kore/ui/PlaylistFragment.java#L455

Is this (still) needed?

SyncedSynapse:
Works fine in my tests. Though you are loosing the spacing between items in the list. This is because `DynamicListView` inherits from `ListView`, and the spacing in `fragment_playlist.xml` is done through the style `GridView`, which only applies to gridviews.

So, 2 solutions:
1. Inherit from GridView. This actually works almost fine. Dragging items is a little confusing when more than one column is present (landscape for instance), so i don't think it's a good idea.
2. Remove the `style` tag from the `DynamicListView` and add:

    android:fastScrollEnabled="true"
    android:divider="@android:color/transparent"
    android:dividerHeight="@dimen/default_grid_vertical_spacing"

To get the divider back.

Regarding your question, `switchToPanel` is needed, just take a look at all the functions that use it. It's basically a way to have an info panel.

poisdeux:
Ok, I changed the code to implement solution 2.

About the `switchToPanel`. I can see it is used at numerous places in the code. However, it doesn't really do anything. It seems it is a leftover from a previous design. Where switching between infopanel and playlist was performed by changing their visibility. Now it is handled by the adapter, as far as I can tell.

SyncedSynapse:
I'm not sure i understand your remark.

`switchToPanel` forces showing up the infopanel when the kodi stops playing, or there's a connection error, to show an appropriate message. Note that `playerOnStop`, `playerOnConnectionError`, etc are kind of event handlers, called when the kodi player changes state.

If it doesn't do that, the playlist will be always shown. I guess something could be done with `playlistView.setEmptyView`, but that would imply that we empty the list view manually on receiving the above event handlers to force showing up the empty view, which is less obvious IMO.

poisdeux:
Ah I see. Sorry, my bad. I think in one of my trials its usage became
unclear.

PR #181, opened 2016-01-23
Improved URLs parsing and added error message

tomerf:


app/src/main/java/org/xbmc/kore/ui/RemoteActivity.java
```diff
@@ -524,9 +529,15 @@ private Uri getYouTubeUri(String extraText) {
...
-                            Pattern.CASE_INSENSITIVE);
+            final Pattern pattern;
+            if (playuri.getHost().endsWith("vimeo.com")) {
+                pattern = Pattern.compile("^(?:https?:\\/\\/)?(?:www\\.|player\\.)?vimeo\\.com\\/(?:.*\\/)?(\\d+)(?:\\?.*)?$",
+                        Pattern.CASE_INSENSITIVE);
+            }
+            else {
+                pattern = Pattern.compile("^(?:https?:\\/\\/)?(?:www\\.|m\\.)?youtu(?:\\.be|be\\.com)\\/watch\\?(?:.*&)?v=([\\w-]+)(?:&.*)?$",
+                        Pattern.CASE_INSENSITIVE);
+            }
```
SyncedSynapse: This breaks the handling of Youtube URIs shared from the youtube app.

These URIs are of the form http://youtu.be/videoid

Didn't test sharing from the website.

tomerf: I only tested it with the website.
Fixed now.

PR #183, opened 2016-01-24
Correcting grammar

TwizzyDizzy:
I'm not a native english speaker, but "as", sounds far more correct than "of".


SyncedSynapse:
Agree.

I would appreciate if some native English speaker could review the messages, particularly the longer ones, as i'm sure there are other errors in them.

PR #184, opened 2016-01-25
Flip RTL text before sending it

royiarchy:
When sending RTL text, it is shown reversed in the input box and also searches the reversed string when submitting.
I solved this problem by reversing the text (only when it is RTL) before sending it to the media center.


SyncedSynapse:
Just to clarify: this occurs when your device is in a locale that uses RTL, right. What about Kodi, if it is in the same locale (with RTL) won't it expect a RTL text?

royiarchy:
It happens when the entered text is RTL, regardless of the device and Kodi's locale.

PR #185, opened 2016-01-26
Added settings options for allowed network types

poisdeux:
This allows the user to select other network connection types,
besides WiFi, for downloading media files.

The "Any" network type is added to allow downloading over USB network connections and probably bluetooth as well (didn't test that, as I do not have bluetooth on a device running Kodi).

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/AlbumDetailsFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AlbumDetailsFragment.java
@@ -27,10 +27,10 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.preference.PreferenceManager;
 import android.provider.BaseColumns;
-import android.support.v4.app.Fragment;
 import android.support.v4.app.LoaderManager;
 import android.support.v4.content.CursorLoader;
 import android.support.v4.content.Loader;
+import android.support.v4.widget.SwipeRefreshLayout;
 import android.text.TextUtils;
 import android.util.DisplayMetrics;
 import android.view.LayoutInflater;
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/xbmc/kore/ui/AlbumDetailsFragment.java
```diff
@@ -156,13 +157,7 @@ public static AlbumDetailsFragment newInstance(AlbumListFragment.ViewHolder vh)
...
 
     @Override
-    public void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
-    }
-
-    @Override
-    @TargetApi(21)
-    public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
+    protected View createView(LayoutInflater inflater, ViewGroup container) {
```
SyncedSynapse: lint error here. Keep the TargetApi line.

SyncedSynapse:
@poisdeux 
In the future please don't auto indent the whole file, only the lines you touch, so as to reduce the number of changes. Looking through all these changes is difficult and easy to miss something.

Also, make sure you always run lint (`./gradlew build` should do it) so as to catch the maximum number of issues.

app/src/main/java/org/xbmc/kore/ui/AlbumDetailsFragment.java
poisdeux: Fixed.

poisdeux:
Sorry about the diffs introduced by auto indenting. As my programming consists of a lot of copy/pasting I got used to auto indenting over the years. I do it without even thinking about it. I'll try to be more careful.

SyncedSynapse:
I'm probably going to create a "Advanced settings" section and move this setting to there, later.

PR #190, opened 2016-01-30
Added support for pausing when phone in call

tomerf:


app/src/main/java/org/xbmc/kore/Settings.java
```diff
@@ -53,6 +53,10 @@
     public static final String KEY_PREF_SHOW_NOTIFICATION = "pref_show_notification";
     public static final boolean DEFAULT_PREF_SHOW_NOTIFICATION = false;
 
+    // Pause during calls
+    public static final String KEY_PREF_PAUSE_DURING_CALLS = "pref_pause_during_calls";
+    public static final boolean DEFAULT_PREF_PAUSE_DURING_CALLS = true;
```
SyncedSynapse: Should be false by default. New versions shouldn't change the behavior.

app/src/main/java/org/xbmc/kore/service/PauseCallService.java
```diff
@@ -0,0 +1,155 @@
...
+ * something is playing, and keeps running while it is playing.
+ * This service stops itself as soon as the playing stops or there's no
+ * connection. Thus, this should only be started if something is already
+ * playing, otherwise it will shutdown automatically.
+ * It doesn't try to mirror Kodi's state at all times, because that would
+ * imply running at all times which can be resource consuming.
+ *
+ * A {@link HostConnectionObserver} singleton is used to keep track of Kodi's
+ * state. This singleton should be the same as used in the app's activities
+ */
```
SyncedSynapse: This description isn't really related to this class.

app/src/main/java/org/xbmc/kore/service/PauseCallService.java
```diff
@@ -0,0 +1,155 @@
...
+public class PauseCallService extends BroadcastReceiver
+        implements HostConnectionObserver.PlayerEventsObserver {
```
SyncedSynapse: Huuum, not sure about this being a BroadcastReceiver. Take a look at the documentation, particularly at the Lifecycle section to see what might go wrong: http://developer.android.com/reference/android/content/BroadcastReceiver.html

app/src/main/java/org/xbmc/kore/Settings.java
tomerf: This is a new behavior, not changing existing one.

I think this feature should be turned on by default (automatically pause the movie when answering calls).

SyncedSynapse:
I've added some comments to the PR in places that caught my attention at first sight.

Generically i have the following comments:
1. Kore isn't still prepared to use Android's 6 permission mechanism, but it will soon. Accessing READ_PHONE_STATE is a "Dangerous permission", so it should be guarded. This means that it probably can't be declared on the manifest, and it should ask the user for permission when he activates the corresponding setting, and then use `registerReceiver` to register the receiver (though, when to unregister is something to think about...).

2. Not sure about the BroadcastReceiver lifecycle. This implementation probably works, because you start the HostConnectionObserver, but it is not guaranteed to be kept alive.

app/src/main/java/org/xbmc/kore/Settings.java
MartijnKaijser: Well I think it shouldn't. Difference of opinion will always be present on options so I'd agree with @SyncedSynapse

tomerf:
1. I'm not familiar with Android 6's permission mechanism (Still waiting for LG to update my device). How will it work with older devices that don't have >6.0?
1. I see your point about the BroadcastReceiver lifecycle, I'll move handleState() to a Service class.

SyncedSynapse:
1. When an app is compiled against SDK >23 (which Kore isn't yet) and when in Android 6, the permission system changes. It is expected that, when the app needs a permission, it asks the user for it, and recovers gracefully if the user doesn't grant it (as opposed with the "old" system when permissions were granted at install time). This means that, if the app doesn't prompt the user and just assumes it has permission for it (for accessing the phone, in this case), it will break. 
For android < 6, the permissions will be asked on install time, nothing changes.

Accessing the phone state is a dangerous permission, and one that is difficult to justify why a remote app needs it (at install time). The new system is much more clear in these situations: when the user goes into the app's settings, sees that setting (disabled by default), taps it and a prompt shows up asking for the permission. In this context it is clear why it needs the permission.

I was looking at the developer documentation and come across this: http://developer.android.com/reference/android/telephony/PhoneStateListener.html

which seems to be a better way of handling this functionality. Haven't played with it, but looks straightforward and simpler than the alternative (service or broadcastreceiver).

I would say that this could be done like this:
1. Declare the permission on the manifest;
2. In the settings, when the user checks the setting for the first time, ask for the permission. There are standard APIs for this (check http://developer.android.com/training/permissions/requesting.html). If the user doesn't give permission, uncheck the option;
3. Create a `PhoneStateLister` that overrides `onCallStateChange`. Going this way avoids all the state book keeping that you have done in this PR. Just check the play status of Kodi and issue a Play Pause accordingly (no need to use `HostConnectionObserver`.

Now, there's the question on when to register the previous object with `TelephonyManager.listen` and when to unregister it... This might need to be a service similar to `NotificationService` that gets started when something starts playing and gets stopped when it stops (unregistering there).

tomerf:
I updated the commit with the PhoneStateListener but I still need to request for permissions.
I have the code for checking settings ready but I think it's best you will check everything else is ready for SDK v23 first.
Do you want to add this in the meantime or wait for everything?

SyncedSynapse:
I'll do the update to the permissions later, when i update the rest of Kore. Just wanted to make sure that it was in a state where the change would be easy to do.

I won't be able to review and test this right now, so give me a few more days.

SyncedSynapse:
Sorry for not replying earlier. I still think there are some issues with the design of this. Consider these issues:

1. The listener is being registered and unregistered in the `RemoteActivity`. Registering it there is ok imo, but unregistering isn't guaranteed to be called. Imagine that, after starting playback the user exits Kore, then if the video stops "naturally" or by some other means, the listener won't be unregistered.

2. Try this: disable the notification option and enable this option. Start playing something and exit Kore. Lock your phone. Now receive a call and answer it and you'll notice that the video isn't paused. That's because Kore isn't running, so the listener callbacks don't get called.

I'll give it some more thought on the weekend, but i think that this needs to be a service that implements `PhoneStateListener`, but the service part needs to be kept. It should be started in the `RemoteActivity`, and in the `OnStart` register itself with `TelephonyManager`. On the `playerOnStop` and `playerOnConnectionError` it should unregister itself and call `stopSelf` (this needs some thought in the case of music or playlist playing, as i guess that this way it would only run for one item. Maybe sending a delayed message to a handler, that then unregisters and stops it if nothing is playing).

Issue #187, opened 2016-01-28
[feature request] longpress to add episode to playlist


rigid:
it would be nice to have a context menu for episodes on longpress. "Add to playlist" and similar entries (like in the kodi gui context menu) would be useful items.

poisdeux:
Currently the context menu is activated through a context menu button in each list item. See the music lists for an example. Would that also work for you for episodes?

rigid:
Yes, i think that would be fine. Actually I event think that's better.
That way, longpress would stay free to enter some "multi-selection" mode in the future. (So multiple items could be Added to the playlist/Deleted/...)

poisdeux:
Implemented and merged into the mainline.

PR #193, opened 2016-02-01
Added context menu to TV episode list items

poisdeux:
The context menu adds play and queue functionality for tv show
episode items in the tv show episode list.

Replaced checkmark icon with the context menu icon. Now watched episodes are visualised by changing the colour of the episode number, title, and context menu icon.

This implements feature request #187

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/TVShowEpisodeListFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/TVShowEpisodeListFragment.java
@@ -38,6 +38,7 @@ import android.view.ViewGroup;
 import android.widget.CursorTreeAdapter;
 import android.widget.ExpandableListView;
 import android.widget.ImageView;
+import android.widget.PopupMenu;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/xbmc/kore/ui/TVShowEpisodeListFragment.java
```diff
@@ -549,11 +552,20 @@ public void bindChildView(View view, Context context, Cursor cursor, boolean isL
...
-                viewHolder.checkmarkView.setColorFilter(themeAccentColor);
+                viewHolder.episodenumberView.setTextColor(themeAccentColor);
+                viewHolder.titleView.setTextColor(themeAccentColor);
+                viewHolder.contextMenuView.setColorFilter(themeAccentColor);
             } else {
-                viewHolder.checkmarkView.setVisibility(View.GONE);
+                viewHolder.episodenumberView.setTextColor(defaultTextColor);
+                viewHolder.titleView.setTextColor(defaultTextColor);
+                viewHolder.contextMenuView.setColorFilter(themeAccentColor);
+                viewHolder.contextMenuView.clearColorFilter();
```
SyncedSynapse: Any reason for setting the color filter and then clearing it?

SyncedSynapse:
I'm not sure about coloring the episode number/title/menu with the accent color for watched episodes. That tends to draw attention to those episodes, which is exactly the opposite that should be done.

How about using the secondary color `?attr/appTextColorSecondary`.

app/src/main/java/org/xbmc/kore/ui/TVShowEpisodeListFragment.java
poisdeux: Nope, copy/paste leftover, which I forgot to remove.

poisdeux:
I see what you mean. However, instead of changing the text colour I think adding the checkmark before the episode number is a better solution. I don't have any time right now but I'll do it ASAP.

poisdeux:
Changed the layout and added the checkmark again. Hope you like it.

SyncedSynapse:
Looks good.

PR #194, opened 2016-02-01
Implemented showing artist details

poisdeux:
When user selects artist from ArtistListFragment it will now
show a screen displaying artist details. I've taken the TV show
details setup as an example.

This adds the following new functionality:

* Displaying artist fanart
* Displaying artist description
* Download all songs from an artist

In my opinion this makes the UX more consistent. Hope you like it as well.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/provider/MediaDatabase.java
+++ b/app/src/main/java/org/xbmc/kore/provider/MediaDatabase.java
@@ -88,7 +88,16 @@ public class MediaDatabase extends SQLiteOpenHelper {
                 ALBUM_GENRES + "." + MediaContract.AlbumGenres.HOST_ID + "=" + AUDIO_GENRES + "." + MediaContract.AudioGenres.HOST_ID +
                 " AND " +
                 ALBUM_GENRES + "." + MediaContract.AlbumGenres.GENREID + "=" + AUDIO_GENRES + "." + MediaContract.AudioGenres.GENREID;
-	}
+
+        /**
+         * Join to get Songs for an Artist
+         */
+        String SONGS_FOR_ARTIST_JOIN =
+                SONGS + " JOIN " + ALBUM_ARTISTS + " ON " +
+                SONGS + "." + MediaContract.Songs.HOST_ID + "=" + ALBUM_ARTISTS + "." + MediaContract.AlbumArtists.HOST_ID +
+                " AND " +
+                SONGS + "." + MediaContract.Songs.ALBUMID + "=" + ALBUM_ARTISTS + "." + MediaContract.AlbumArtists.ALBUMID;
+    }
 
     private interface References {
         String HOST_ID =
--- a/app/src/main/java/org/xbmc/kore/provider/MediaProvider.java
+++ b/app/src/main/java/org/xbmc/kore/provider/MediaProvider.java
@@ -77,7 +77,8 @@ public class MediaProvider extends ContentProvider {
     private static final int ALBUM_GENRES_LIST = 711;
 
     private static final int SONGS_ALL = 800;
-    private static final int SONGS_LIST = 802;
+    private static final int SONGS_ARTIST = 801;
+    private static final int SONGS_ALBUM = 802;
     private static final int SONGS_ID = 803;
 
     private static final int AUDIO_GENRES_ALL = 900;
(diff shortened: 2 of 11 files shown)


app/src/main/java/org/xbmc/kore/provider/MediaProvider.java
```diff
@@ -77,6 +77,7 @@
     private static final int ALBUM_GENRES_LIST = 711;
 
     private static final int SONGS_ALL = 800;
+    private static final int SONGS_ARTIST = 801;
     private static final int SONGS_LIST = 802;
```
SyncedSynapse: For coherence, this should be changed to SONGS_ALBUM

app/src/main/java/org/xbmc/kore/provider/MediaProvider.java
```diff
@@ -663,6 +668,16 @@ private SelectionBuilder buildQuerySelection(Uri uri, int match) {
...
+            case SONGS_ARTIST: {
+                final String hostId = MediaContract.Hosts.getHostId(uri);
+                final String artistId = MediaContract.Artists.getArtistId(uri);
+                return builder.table(MediaDatabase.Tables.SONGS)
+                              .join(MediaDatabase.Tables.ALBUM_ARTISTS)
+                              .where(MediaDatabase.Tables.SONGS + "." + MediaContract.Songs.HOST_ID + "=?", hostId)
+                              .where(MediaDatabase.Tables.SONGS + "." + MediaContract.Songs.ALBUMID + "=" +
+                                     MediaDatabase.Tables.ALBUM_ARTISTS + "." + MediaContract.AlbumArtists.ALBUMID)
+                              .where(MediaContract.AlbumArtists.ARTISTID + "=?", artistId);
+            }
```
SyncedSynapse: This section doesn't follow the same pattern as the others. 
Take a look at the `ARTIST_ALBUMS_LIST` section for an example. You'll need to create a `MediaDatabase.Tables.SONGS_FOR_ARTIST_JOIN` (similar to `MediaDatabase.Tables.ALBUMS_FOR_ARTIST_JOIN`) that does the join and then use it in this section.
No need to introduce the `builder.join` method.

app/src/main/java/org/xbmc/kore/ui/ArtistDetailsFragment.java
```diff
@@ -0,0 +1,120 @@
...
+import org.xbmc.kore.utils.TabsAdapter;
+import org.xbmc.kore.utils.UIUtils;
+import org.xbmc.kore.utils.Utils;
+
+import butterknife.ButterKnife;
+import butterknife.InjectView;
+
+/**
+ * Container for the TV Show overview and Episodes list
+ */
```
SyncedSynapse: Update comment

app/src/main/java/org/xbmc/kore/ui/ArtistOverviewFragment.java
```diff
@@ -0,0 +1,484 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+
+import butterknife.ButterKnife;
+import butterknife.InjectView;
+import butterknife.OnClick;
+
+/**
+ * Presents movie details
+ */
```
SyncedSynapse: Comment

app/src/main/java/org/xbmc/kore/utils/SelectionBuilder.java
```diff
@@ -83,6 +85,11 @@ public SelectionBuilder where(String selection, String... selectionArgs) {
         return this;
     }
 
+    public SelectionBuilder join(String table) {
```
SyncedSynapse: No need for this

SyncedSynapse:
Have to go, more comments later...

app/src/main/java/org/xbmc/kore/ui/ArtistOverviewFragment.java
```diff
@@ -0,0 +1,484 @@
...
+                     .show();
+            }
+        }, callbackHandler);
+    }
+
+    @OnClick(R.id.add_to_playlist)
+    public void onAddToPlaylistClicked(View v) {
+        final PlaylistType.Item playListItem = new PlaylistType.Item();
+        playListItem.artistid = artistId;
+        MediaPlayerUtils.queueAudio(this, playListItem);
```
SyncedSynapse: Breaks the build. You changed the signature of this method in your last PR, so this should reflect that.

SyncedSynapse:
Looks good and is indeed better than it was.

My only question about this is that it introduces one more step to get to the artist's albums (one more click), which usually is what the user wants to do primarily. This could be fixed by making the "Albums" tab the default for the activity. Only downside is that we would loose the shared element animation, which i personally like a lot...

poisdeux:
I think the extra effort to get to the artist albums is a minor problem, as I believe most users will use a touchscreen device and it will only take a simple swipe to get the list of albums. Personally, I usually want to listen to an artist and not a specific album. If I do want to listen to a specific album, I go to the album list and use search.
But the real reason I wanted this, is that in my opinion it enhances the UX of listening to music. Just like we used to do (long, long ago....) reading the booklet or vinyl cover :D

app/src/main/java/org/xbmc/kore/provider/MediaProvider.java
poisdeux: Ah I didn't see that. Fixed now.

poisdeux:
Code updated

app/src/main/java/org/xbmc/kore/provider/MediaDatabase.java
```diff
@@ -88,7 +88,14 @@
...
                 " AND " +
                 ALBUM_GENRES + "." + MediaContract.AlbumGenres.GENREID + "=" + AUDIO_GENRES + "." + MediaContract.AudioGenres.GENREID;
-	}
+
+        /**
+         * Join to get Songs for an Artist
+         */
+        String SONGS_FOR_ARTIST_JOIN =
+                SONGS + " JOIN " + ALBUM_ARTISTS + " ON " +
+                SONGS + "." + MediaContract.Songs.ALBUMID + "=" + ALBUM_ARTISTS + "." + MediaContract.AlbumArtists.ALBUMID;
```
SyncedSynapse: Add HOST_ID to the join to be unique.

poisdeux: I thought it was redundant as host id was added to the where clause in the builder. But after taking a closer look I noticed it isn't the same. Sorry about that.

Issue #196, opened 2016-02-05
Item lists (Directory listings, playlist, ...) cut long names


BenBE:
When file information are not present and thus for browsing e.g. a directory of your video list, typical names like "[Foo] Beautiful Series S05E23 [08F47B23].mkv" are inadvertedly shortened to something like "[Foo] Beautiful Series S...].mkv". Which kinda make it useless in a directory of your favorite series ;-)

Also switching from portrait to landscape doesn't help much to view more of the filename, as a second column is displayed and the displayed portion of the filename gets even smaller.

poisdeux:
I agree and would like to see an action button added to the toolbar to allow the user to switch between single and multi-column mode. I do not think we should remove shortening long filenames though.
If @SyncedSynapse agrees I'll implement it.

BenBE:
What might be viable is having a short menu at the right of each item (filename shortened) to play and enqueue an item with one click, yet clicking the item normally unfolds it (unshortened name) with several more options (enqueue first, enqueue last, rename, change metadata, ...).

This also solves another UX issue if you accidentally touch an item instead of hit the quite small menu.

SyncedSynapse:
@poisdeux I have no objection to having a setting that changes between single/multi column mode, as long as it it beneath the overflow menu. Furthermore, give that it isn't much helpful in a phone in portrait, it probably should only be available to tablets or in landscape (not sure if res/menu can be specified for sw600dp / land, need to check)

SyncedSynapse:
2 lines are now visible on the files list to help with this.

PR #202, opened 2016-02-11
Implemented single/multi column option for lists

poisdeux:
Added a menu item in the overflow menu to allow the user
to switch the gridview between single column and multi column mode

This fixes issue #196


SyncedSynapse:
Merging this as is, but i'll change the order of the items on the menu. I think that "Refresh" should be last.

PR #203, opened 2016-02-13
Added home button

tomerf:


SyncedSynapse:
In which situations does this help? There's already a Home button on the bottom bar that does the same, as far as i could tell.
Asking this because i'm not a fan of adding long press actions to the d-pad, select or back buttons.

tomerf:
Missed the home button (the entire bar actually).
But I still think it's logical to add it there, since pressing the back button multiple times will eventually bring the user to home view.

SyncedSynapse:
I understand your position but in this case i don't agree with it. 

I think that adding long presses on the remote buttons isn't something that should be done just because we can (it is a hidden feature and can confuse users), but should only be done if there's no alternative. Furthermore, it should be avoided on the "main" buttons (the directional buttons, the select and back), and only used on the secondary buttons. We also might want to change the handler of "back" in the future, for instance to allow repetition, and introducing this will cause conflits.

I'm not extremely opposed to it, and i agree that my reasons are not rock solid, but as i said, i also don't think there's a good reason for introducing this, given that it is available one row down.

PR #204, opened 2016-02-13
Cleanup unused methods

tomerf:


app/src/main/java/org/xbmc/kore/utils/NetUtils.java
```diff
@@ -194,23 +194,6 @@ public static boolean sendWolMagicPacket(String macAddress, String hostAddress,
...
-            for (int i = 0; i < 6; i++) {
-                bytes[i] = (byte) Integer.parseInt(hex[i], 16);
-            }
-        }
-        catch (NumberFormatException e) {
-            throw new IllegalArgumentException("Invalid hex digit in MAC address.");
-        }
-        return bytes;
-    }
-
```
SyncedSynapse: This one can be useful in the future for some reason that i can't think of right now... I would keep it.

tomerf: Restored.

PR #206, opened 2016-02-14
Remove Hard Coded Strings to allow translation

KowalskiOmniROM:
Screenshots https://plus.google.com/115610928059890728494/posts/egPEeCY4e8k


SyncedSynapse:
Agree, but this isn't building because the strings need to be added to strings.xml first.

SyncedSynapse:
See comment on #213

Issue #161, opened 2015-12-24
Last search should be remembered


jondo:
* Go to the 'albums' list.
* Search for a text string to filter the listed albums.
* Click one of the search results. This switches to the album's track list.
* Now the 'back' button should bring me back to the filtered list. However, the filter is lost and I am back at the full 'albums' list.

PR #210, opened 2016-02-15
Implemented restoring search query

poisdeux:
When user enters a search query for a *ListFragment the query will
now be restored on device configuration changes and when
*ListFragment is restored from the backstack.

Fixes issue #161


poisdeux:
Needs more testing.

PR #212, opened 2016-02-17
Add strings before commit to preferences.xml

KowalskiOmniROM:
Remove Hard Coded Strings to allow translation (1/2)


SyncedSynapse:
Same comment as on PR #213

Issue #61, opened 2015-04-03
Add option to disable lockscreen while remote is active


ThiefMaster:
Another useful feature from the old xbmc remote. While watching a movie it's quite annoying to unlock the phone using the lockscreen pattern (or whatever is used) just to access the remote.

For extra security I'd suggest to automatically re-enable the lockscreen once playback stops, so if you e.g. fall asleep while watching a movie your phone is only unlocked until it finished.

shtrom:
:+1: see also #109, for a widget on the lock screen.

sputtene:
I use Llama for that. It's an app that automatically performs actions when certain conditions are met. You can find it on Google Play for free.

My events in Llama:
- When the current app is Kore: disable screen timeout
- When charging: set screen timeout to 300 seconds
- When using battery: set screen timeout to 60 seconds

The first event keeps the screen on when Kore is the active app, so the lock screen isn't displayed either.
One of the two other events take care of reinstating the screen timeout when Kore is no longer the active app.

shtrom:
Not available on F-Droid ):

sputtene:
Unfortunately not. I've searched F-Droid first, but didn't find any apps that allow you to automate stuff. If you find any, please let me/us know :)

I've settled for Llama from the Play Store which is free, unlike Tasker for example.

SyncedSynapse:
@jtdor has added a commit that helps with this issue. It doesn't keep the phone awake, but i'm not a fan of that either (and i believe it requires additional permissions, which i dislike even more), but is a good solution imo.

PR #215, opened 2016-02-19
Added option to keep the remote above the lockscreen (issue #61)

jtdor:


SyncedSynapse:
Nice. I'm just not sure about the naming of the option (in English), but i can't come up with anything better so...

I'll probably move this option to a "Advanced settings" section in the future.

jtdor:
Thanks for merging. I'm not really happy about the wording (both English and German) either, but did not find anything less clumsy.

PR #222, opened 2016-03-17
Implemented restoring search query

### LINKED ISSUE #161: Last search should be remembered

poisdeux:
When user enters a search query for a ListFragment the query will
now be restored on device configuration changes and when
ListFragment is restored from the backstack.

I hit the limit of manually testing this and implemented unit tests to automate all cases I could think of. See #223 for the the PR implementing unit tests.

I created tests for two different list fragment setups. One using a view pager and one that does not.
See 
https://github.com/poisdeux/Kore/blob/unittest/restoresearchquery/app/src/androidTest/java/org/xbmc/kore/RestoreSearchQueryViewPagerTest.java
for the view pager tests.

See https://github.com/poisdeux/Kore/blob/unittest/restoresearchquery/app/src/androidTest/java/org/xbmc/kore/RestoreSearchQueryListFragmentTest.java
for the single list fragment tests

Btw this fixes #161

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/AbstractCursorListFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AbstractCursorListFragment.java
@@ -17,6 +17,7 @@
 package org.xbmc.kore.ui;
 
 import android.annotation.TargetApi;
+import android.app.Activity;
 import android.content.Intent;
 import android.content.ServiceConnection;
 import android.database.Cursor;
@@ -25,15 +26,20 @@ import android.support.annotation.Nullable;
 import android.support.v4.app.LoaderManager;
 import android.support.v4.content.CursorLoader;
 import android.support.v4.content.Loader;
+import android.support.v4.view.MenuItemCompat;
 import android.support.v4.widget.SwipeRefreshLayout;
 import android.support.v7.widget.SearchView;
+import android.text.TextUtils;
 import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.AdapterView;
+import android.widget.BaseAdapter;
 import android.widget.CursorAdapter;
+import android.widget.EditText;
 import android.widget.Toast;
 
 import org.xbmc.kore.R;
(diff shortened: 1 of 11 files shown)


poisdeux:
Hmmm, I missed a scenario:

1. User enters search query
2. User removes search query
3. User switches tabs
4. User returns to original tab
5. Result: search query should be cleared. But it is restored to the last entered search query...

I will fix/add it ASAP.

poisdeux:
Well, that was fun. Implementing above mentioned scenario resulted in a complete redesign of my previous solution....

SyncedSynapse:
Did you fix the scenario you identified in a previous comment? I tried it and it still happens.

If it can be fixed it would be nice, but it's not something i would require to merge this, as it only happens on the music section and isn't a deal breaker.

app/src/main/java/org/xbmc/kore/ui/AbstractListFragment.java
```diff
@@ -43,6 +46,7 @@
 import android.widget.TextView;
 import android.widget.Toast;
 
+import org.w3c.dom.Text;
```
SyncedSynapse: Not needed. If you resubmit this PR please remove it.

poisdeux:
I thought I did, however I noticed I did not test clearing the text using the X button...
So apparently, I fixed above mentioned scenario for the cases the user clears the text using backspace or presses the back key (either the hardware key or the one in the toolbar).

I do not have time right now to fix this. I will add a comment in the unittest to test the different ways of clearing the text as well.

SyncedSynapse:
Do you want me to merge this as is and later fix it, or should it wait?

poisdeux:
Just leave it open and I'll add a comment when I updated this PR with the
fix.

poisdeux:
I've updated this PR and now includes the fix to handle the close button.

SyncedSynapse:
I applied #228 first and now this doesn't apply directly (applying this one first and then #228 would also result in conflicts). Mind checking the conflicts against the current codebase.

poisdeux:
Updated

Issue #233, opened 2016-04-24
Tapping a preference toggles other preferences


jtdor:
Hi,

the settings in Kore are almost unusable for me. When I tap on one of the preferences other preferences are toggled in some arbitrary order until the preference I'm actually tapping on is toggled. The only description of a similar problem I found was on Stack Overflow ([Android SwitchPreferences change together in PreferenceActivity](https://stackoverflow.com/questions/20906338/android-switchpreferences-change-together-in-preferenceactivity)). A solution seems to be to use CheckBoxPreference instead of SwitchPreference or use a custom SwitchPreference as a workaround as described [here](https://code.google.com/p/android/issues/detail?id=26194#c4).

Would it be ok for you to switch to CheckBoxPreference? This would even leave a little more space to display the description of the preferences.

poisdeux:
I am unable to reproduce this. Which version of Android are you using and on what kind of device?

jtdor:
I'm using an HTC One mini 2 phone with the preinstalled Android 4.4.2.

SyncedSynapse:
I was able to reproduce this on 4.3.

I really think that switch preferences are more appropriate here, but given that this can potentially affect every device prior to lollipop (about 30% of the installed base) using checkboxes seems prudent.

PR #234, opened 2016-04-24
Replace SwitchPreference with CheckBoxPreference to avoid weird toggling behaviour of multiple SwitchPreferences on the same activity (issue #233)

jtdor:


SyncedSynapse:
Merging this because this issue seems to affect versions pre-lollipop.

Issue #219, opened 2016-03-08
Incomplete library


juup:
Sorry, I'm not sure where the right place is to post the issue. I found similar post in the [forum](http://forum.kodi.tv/showthread.php?tid=251897&pid=2273589#pid2273589), but issue tracker from F-Droid linked here.

Since update to Jarvis the music library is incomplete.
I've already tried different servers (Raspberry Pi and Linux Mint), and deleted the list and rescan it on server. There are lot of albums missing in the remote app after update the list. 
The "Browser Library" seems to be complete, so I guess the remote app have some trouble.

Version v2.0.0

poisdeux:
I can confirm I have the same issue with my OSMC setup. I have not yet been able to reproduce this on my dev/test setup. Still working on it.
@SyncedSynapse do you have any idea what might be causing this?

SyncedSynapse:
@poisdeux No, but it would be useful if you could run a debug version of Kore connected to your OSMC, do a library sync and check the log. The relevant code is in `LibrarySyncService.java` and it contains some log messages that might be useful.

IT syncs albums/artists/songs in chunks, 300 for artist/albums, 600 for songs, so i would keep an eye on the log to check if it reached the end. If the issue is on Kore's side it should just blow up, but maybe it just isn't receiving all the data from Kodi. To check that, sending a HTTP request to OSMC would help. Something like this (didn't test it, you will get some examples on the debug log):
`{
    "jsonrpc": "2.0",
    "method": "AudioLibrary.GetAlbums",
    "id": 32,
    "params": {
        "limits": {"start": 0, "end": 300}
        "properties": ["title", "description", "artist", "genre"]
    }
}
`

SyncedSynapse:
Should be fixed (in development version). If not, please reopen.

PR #236, opened 2016-04-26
Issue/#219

poisdeux:
This PR includes PR #237. You probably want to review and merge that first.

This should fix issue #219 for music items. It should be implemented for other media items as well. But I want to show/discuss my proposed solution first.

The problem with the current way of determining if there are more items available on Kodi, is that it uses the size of the list of returned elements. When requesting media items from Kodi through JSON RPCs the `start` and `end` index are given for some media. To determine if there are more items available on Kodi, the size of the returned list is used (which should equal `end - start`). If Kodi returns less items than the requested size, it is assumed that all items have been retrieved.
I noticed that with Jarvis, Kodi may return a list smaller than `end - start`. This will make Kore stop recursing and thus only retrieve a small part of the available items.
Fortunately, Kodi returns the limit with the actual returned `start` and `end` indices. My solution uses this, to determine if we should recurse and get more items. I have tested my solution with Isengard and Jarvis and AFAICT it works.
(I'm thinking of adding a unit test for this in the future)


SyncedSynapse:
Lot of changes... I wish you would have run this by me before, because i don't think that introducing `JsonResult` this way is the way to do it...

So, let's take it step by step:

1. Regarding #237 i haven't reviewed it, but i'm assuming that this is just spliting the various inner classes on the `LibrarySyncService` to different files, right? If so i agree.

2. 
Regarding introducing `JsonResult`, it shouldn't be introduced as the base return type of all json-rpc methods. The `limits` info is only returned on responses that have lists, so it should only be available on those methods.
So, assuming we need the `LimitsReturned` info from Kodi (i'll get to this later), and taking for instance `GetArtists`, we currently have:
`public static class GetArtists extends ApiMethod<List<AudioType.DetailsArtist>>`
We should only substitute these type of classes with something like:
`public static class GetArtists extends ApiMethod<ApiList<AudioType.DetailsArtist>>`
`ApiList` (maybe not the best name) should have the `LimitsReturned` and `Items`. It could probably be a subclass of `List`, adding the `LimitsReturned` (haven't thought this through, so this might not be the best way, but at first sight makes sense).
Now, this is more work than just introducing `JsonResult` to all methods/classes, but it's the right way to handle it, and it's consistent with the way the API is constructed. Ideally it should be done for all methods, but it doesn't need to be so, we could do it on a needed basis.

3. You've added a `total` member to the `List.Limits` class, but it shouldn't be done there. This might be confusing, but the API is mimicking the json specification (available on the introspect files), and there's `List.Limits`, which is used to specify the limits we want returned on a call, and there's `List.LimitsReturned`, which is used to identify which limits are returned. `List.LimitsReturned` isn't yet available on the `ListType` class, but it's the one we want to implement and use on the return side (with the start, end and total members).
Not sure if this is clear. One of the design principles of this api layer is to be as close and consistent with the json-rpc specification as possible, and it's important to always check the specification before creating new types. It's a little confusing at first but it's the only way to keep it consistent.

4. Now, regarding the original issue, i agree with the solution, but i want to confirm with other team members that this started happening on Jarvis. Pinging @Montellese (just read from this point on): Can you confirm that from Jarvis on, a call to for instance `GetArtists` that specify `Limits` can return less than the requested number, even though there are more items than those (e.g request 300 results, get back 250 even though there are 500 results)? Is this related to a webserver buffer size limit or something similar that i saw mentioned somewhere?

poisdeux:
On Thu, Apr 28, 2016 at 8:30 PM, Synced Synapse <notifications@github.com>
wrote:

> Lot of changes... I wish you would have run this by me before, because i
> don't think that introducing JsonResult this way is the way to do it...
>
I believe you mean "run this by me first"? If so, I had to write some kind
of code to test my solution. This was the quickest way for me to add the
limits I needed.
I believe PRs are meant to discuss the solution/code. I prefer to just
write the code and have something to actually talk about. Instead of
discussing it up-front when
I don't have a clear view of what the problem is and what might be a
solution.

So, let's take it step by step:
>
>    1.
>
>    Regarding #237 <https://github.com/xbmc/Kore/pull/237> i haven't
>    reviewed it, but i'm assuming that this is just spliting the various inner
>    classes on the LibrarySyncService to different files, right? If so i
>    agree.
>
> Yup, that's it. It made it easier for me to get an overview and get a
glimpse of the code design.


>    1.
>
>    Regarding introducing JsonResult, it shouldn't be introduced as the
>    base return type of all json-rpc methods. The limits info is only
>    returned on responses that have lists, so it should only be available on
>    those methods.
>    So, assuming we need the LimitsReturned info from Kodi (i'll get to
>    this later), and taking for instance GetArtists, we currently have:
>    public static class GetArtists extends
>    ApiMethod<List<AudioType.DetailsArtist>>
>    We should only substitute these type of classes with something like:
>    public static class GetArtists extends
>    ApiMethod<ApiList<AudioType.DetailsArtist>>
>    ApiList (maybe not the best name) should have the LimitsReturned and
>    Items. It could probably be a subclass of List, adding the
>    LimitsReturned (haven't thought this through, so this might not be the
>    best way, but at first sight makes sense).
>    Now, this is more work than just introducing JsonResult to all
>    methods/classes, but it's the right way to handle it, and it's consistent
>    with the way the API is constructed. Ideally it should be done for all
>    methods, but it doesn't need to be so, we could do it on a needed basis.
>
> Thanks for explaining. I don't mind the extra work, I just didn't see
where else to put the returned limits


>    1.
>
>    You've added a total member to the List.Limits class, but it shouldn't
>    be done there. This might be confusing, but the API is mimicking the json
>    specification (available on the introspect files), and there's
>    List.Limits, which is used to specify the limits we want returned on a
>    call, and there's List.LimitsReturned, which is used to identify which
>    limits are returned. List.LimitsReturned isn't yet available on the
>    ListType class, but it's the one we want to implement and use on the
>    return side (with the start, end and total members).
>    Not sure if this is clear. One of the design principles of this api
>    layer is to be as close and consistent with the json-rpc specification as
>    possible, and it's important to always check the specification before
>    creating new types. It's a little confusing at first but it's the only way
>    to keep it consistent.
>
> Indeed, this wasn't clear to me. I will update the code, probably
somewhere next week.

SyncedSynapse:
Ok, if you need some opinions or have any doubts post them here.

poisdeux:
Using your suggestions I refactored the code. See 
https://github.com/poisdeux/Kore/compare/refactor/librarysyncservice...poisdeux:issue/%23219
for a quick diff between the refactoring of LibrarySyncService and the fix for #219

SyncedSynapse:
I can't comment on the diff, so i'll comment here. Hopefully it will be clear from context what i'm commenting on.

1. Class `ApiList` looks good, though i wasn't thinking on using it the way you do.

2. Class `ApiListMethod`. Why do you need this? Just use `ApiMethod`. A method that returns a list is currently instantiated as `ApiMethod<List<Something>>`, with this it should be `ApiMethod<ApiList<Something>>` and the `ApiList` result should be constructed on his `resultFromJson`. For instance `GetArtists` should extend `ApiMethod<ApiList<AudioType.DetailsArtist>>` and have a `public ApiList<AudioType.DetailsArtist> resultFromJson(ObjectNode jsonObject)` that constructs the ApiList (just calling the `LimitsReturned` constructor with the correct json node, and using the current code for the rest of the list).  Goal: take a look at the `SyncMusic` class, line 104. This shouldn't be `public void onSuccess(ApiList<List<AudioType.DetailsArtist>> result)`, it should be `public void onSuccess(ApiList<AudioType.DetailsArtist> result)`.

3. You don't need class `ApiReturnParameter`. All classes on jsonrpc/type are return parameters, and `LimitsReturned` is just one more that should follow existing conventions. `LimitsReturned` should have a constructor that does what you did in its `parseJsonNode`. Note: i get that you looked at `ApiParameter` and noticed that it inherited from `ApiParameter` and replicated the logic to this one, but the way the API is constructed classes that are possible inputs to a json method should implement this, but outputs don't need to implement anything, just rely on its constructor. I admit that it might make sense doing what you did, but to keep it consistent, it would have to be done to all the existing types...

Logic on `SyncMusic` looks good. Maybe the function `moreItemsAvailable` should be a member function on `LimitsReturned` so that it could be used on other places (the name should be changed though, to something like `lastChunk` or something).

poisdeux:
2. I used inheritance to minimize code refactoring and reduce code duplication. I changed it and use composition now.

3. Changed.

I'm not sure about putting `moreItemsAvailable` in `LimitsReturned`. I think the current setup of ListType seems to be converting data to/from JSON. I think moving it to a class in the utils package would be more logical.

app/src/main/java/org/xbmc/kore/jsonrpc/method/AudioLibrary.java
```diff
@@ -85,7 +85,7 @@ public String resultFromJson(ObjectNode jsonObject) throws ApiException {
     /**
      * Retrieve all artists
      */
-    public static class GetArtists extends ApiListMethod<List<AudioType.DetailsArtist>> {
+    public static class GetArtists extends ApiMethod<ApiList<List<AudioType.DetailsArtist>>> {
```
SyncedSynapse: No, the goal is to have signature:
`ApiMethod<ApiList<AudioType.DetailsArtist>>`

`ApiList` substitutes `List` and encapsulates the `List` and `LimitsReturned`

SyncedSynapse:
Please check the latest line comments. `ApiList` is meant to substitute the use of `List` on methods that returns lists.

app/src/main/java/org/xbmc/kore/jsonrpc/method/AudioLibrary.java
poisdeux: But of course... no need for items to support more than List

SyncedSynapse:
I merged the songs PR first, so now this doesn't merge cleanly. `SongsListFragment.java` uses `org.xbmc.kore.service.LibrarySyncService` which is changed in this PR.

Can you check it.

SyncedSynapse:
And by the way, #219 isn't needed anymore, right?

poisdeux:
Fixed and squashed.

Indeed #219 can be closed.

Issue #72, opened 2015-04-24
[feature request] Full control of add-ons


axelsimon:
Hi,

Ok, first of all, I realise this is likely to require a large amount of work, but I think it would be a very neat enhancement.
Instead of simply showing installed add-ons and taking you to the add-on's screen, it would be great if we could actually control the add-on from Kore.

The main problem I see is that there is no standardisation of how add-ons work and what they do. One solution could be to get add-ons to come with a mapping of their (main?) functions, which could be used by Kore to build the menus and send the events/requests.

As Kore is now the official Kodi remote, this might give us enough weight to suggest that add-ons need to now include this description which could be used to control them. That should probably be discussed with the main Kodi team.

As an example, here is an actual use case. I use the DI.fm app a fair bit, it's pretty simple: open add-on, add-on loads list of DI radios, I click radio, music magically appears.
This same could be done on Kore (go to add-ons, DI, see list of radio, select radio) avoiding having to turn the TV on just to listen to music.

Another use case is adding items to the playlist (say Vimeo or Youtube videos) while a video is already playing and thus using up the main display.

Anyway, I'd be interested in discussing this enhancement and your opinion on how feasible it is.

Cheers

axel

SyncedSynapse:
With all the variety in add-ons i'm not sure it would be feasible to standardize a simple set of functions, but i'm not the best person to evaluate this.
In any case, i think that the primary site for discussing this shouldn't be here. You probably should post on the forum to gauge interest and feasibility from add-on developers.

bromix:
I would just enable the navigation of an adoon like Yatse did. In Yatse you can navigate the current content of an addon. Nothing more - everything else would complicate things.

SyncedSynapse:
@bromix what do you mean by navigation? Using the d-pad on add-ons? That's possible (if there are any issues with this it might have something to do with threads and TCP on Kodi, and the solution is to disable TCP in Kore).

Razzeee:
![screenshot_2015-06-21-20-02-48](https://cloud.githubusercontent.com/assets/5943908/8272770/e97c824a-1850-11e5-91ab-2f0416832d1b.png)
Here is an example from the YouTube app. It seems to fetch the list items

wp9015362:
> @axelsimon wrote:
> 
> Instead of simply showing installed add-ons and taking you to the add-on's screen, it would be great if we could actually control the add-on from Kore.

+1

That would be really nice.

Regards

eisnerd:
[...]

PR #239, opened 2016-05-04
Plugin/addon content browser

eisnerd:
See #72.

This adds a tab beside the addon details view for browsing the content that's presented by the addon through its virtual file tree.

This also adds a "Pin/Unpin" button to the addon details view which allows the content tab for a particular addon to be pinned as a tab in the addons activity for quick access. This gives addons that present a good content tree ease of use comparable to the Files browser, once you've picked them out.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/AddonsActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AddonsActivity.java
@@ -206,7 +206,7 @@ public class AddonsActivity extends BaseActivity
         selectedAddonTitle = vh.addonName;
 
         // Replace list fragment
-        final AddonDetailsFragment addonDetailsFragment = AddonDetailsFragment.newInstance(vh);
+        final AddonOverviewFragment addonDetailsFragment = AddonOverviewFragment.newInstance(vh);
         FragmentTransaction fragTrans = getSupportFragmentManager().beginTransaction();
 
         // Set up transitions
--- a/app/src/main/java/org/xbmc/kore/ui/AddonsActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AddonsActivity.java
@@ -206,7 +206,11 @@ public class AddonsActivity extends BaseActivity
         selectedAddonTitle = vh.addonName;
 
         // Replace list fragment
-        final AddonOverviewFragment addonDetailsFragment = AddonOverviewFragment.newInstance(vh);
+        final SharedElementFragment addonDetailsFragment =
+            vh.browsable
+            ? AddonOverviewFragment.newInstance(vh)
+            : AddonDetailsFragment.newInstance(vh)
+            ;
         FragmentTransaction fragTrans = getSupportFragmentManager().beginTransaction();
 
         // Set up transitions
--- a/app/src/main/java/org/xbmc/kore/ui/AddonsActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AddonsActivity.java
@@ -70,7 +70,7 @@ public class AddonsActivity extends BaseActivity
         navigationDrawerFragment.setUp(R.id.navigation_drawer, (DrawerLayout) findViewById(R.id.drawer_layout));
 
         if (savedInstanceState == null) {
-            AddonsOverviewFragment addonListFragment = new AddonsOverviewFragment();
+            AddonListContainerFragment addonListFragment = new AddonListContainerFragment();
 
             // Setup animations
             if (Utils.isLollipopOrLater()) {
--- a/app/src/main/java/org/xbmc/kore/ui/AddonsActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AddonsActivity.java
@@ -70,7 +70,7 @@ public class AddonsActivity extends BaseActivity
         navigationDrawerFragment.setUp(R.id.navigation_drawer, (DrawerLayout) findViewById(R.id.drawer_layout));
 
         if (savedInstanceState == null) {
-            AddonListFragment addonListFragment = new AddonListFragment();
+            AddonsOverviewFragment addonListFragment = new AddonsOverviewFragment();
 
             // Setup animations
             if (Utils.isLollipopOrLater()) {
--- a/app/src/main/java/org/xbmc/kore/ui/AddonsOverviewFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AddonsOverviewFragment.java
@@ -15,7 +15,6 @@
  */
 package org.xbmc.kore.ui;
 
-import android.annotation.TargetApi;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.os.Bundle;
@@ -31,10 +30,8 @@ import org.xbmc.kore.R;
 import org.xbmc.kore.utils.LogUtils;
 import org.xbmc.kore.utils.TabsAdapter;
 import org.xbmc.kore.utils.UIUtils;
-import org.xbmc.kore.utils.Utils;
 
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.Set;
 
 import butterknife.ButterKnife;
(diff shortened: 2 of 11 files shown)


app/src/main/java/org/xbmc/kore/ui/AddonsOverviewFragment.java
```diff
@@ -0,0 +1,112 @@
...
+        Set<String> bookmarked = prefs.getStringSet("bookmarked", Collections.<String>emptySet());
+        long baseFragmentId = 70 + bookmarked.size() * 100;
+        for (String path: bookmarked) {
+            String name = prefs.getString("name_" + path, "Content");
+            Bundle addon = new Bundle();
+            addon.putString(AddonDetailsFragment.BUNDLE_KEY_NAME, name);
+            addon.putParcelable(MediaFileListFragment.ROOT_PATH, new MediaFileListFragment.FileLocation(name, "plugin://" + path, true));
+            tabsAdapter.addTab(MediaFileListFragment.class, addon, name, baseFragmentId++);
+        }
+        tabsAdapter.addTab(AddonListFragment.class, new Bundle(), R.string.addons, baseFragmentId);
```
SyncedSynapse: This line should be 10 lines upper, as the addons tab should always be the first one.

app/src/main/java/org/xbmc/kore/ui/AddonsOverviewFragment.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Collections;
+import java.util.Set;
+
+import butterknife.ButterKnife;
+import butterknife.InjectView;
+
+/**
+ * Container for the TV Show overview and Episodes list
+ */
+public class AddonsOverviewFragment extends Fragment {
```
SyncedSynapse: You've added a `AddonOverviewFragment` and a `AddonsOverviewFragment`... This is confusing and can lead to mistakes. We should use a more distinct name for this.

SyncedSynapse:
Good one, though i have some comments.

I'll preface this by saying that i'm not a heavy user of addons, so my opinion might no be the best one, but here it goes.

* As i've said before i'm not 100% sure about putting the contents of addons on tabs in the details activity (I don't really like tabs on the details view, even though they are used in the TV Shows for instance - i plan to remove them though). It also has the side effect that the content fragment gets loaded as soon as you enter the details view, which can "start" the addon which is not intended (take a look at the global search addon for instance). That being said, i agree that it's an obvious place to put the navigation, and it is easily discoverable.

* I think that the idea of pinning/unpinning is a clever one, but I'm not 100% sold on it. I think that it might be confusing to the user what it is doing (at least the first time), and it also has the side effect of "starting" the addons as soon as the user gets to the list (the first 2 pinned at least, as that is the default number of fragments that get started). I think it is (extremely?) useful for power users, but i can't quantify it, as i'm not one...

I'd like to have some more opinions on the way this was implemented. I'll ask some team members who have more experience with addons to weight in.

SyncedSynapse:
One more thing: browsing on the file list assumes that we're browsing directories, so when browsing upwards it just finds the previous / and navigates there. Browsing on addons works like this most of the time, but i believe that not always. For instance, on Vimeo (youtube also?) there's no way to get back to the "root" of the addon because the URI is different (haven't checked but the error it gives is consistent with this).

Not sure if this is a general problem on other addons. Navigating should really use a stack...

app/src/main/java/org/xbmc/kore/ui/AddonsOverviewFragment.java
eisnerd: Moving this 10 lines up would indeed ensure that the Addons tab would always be the first one. Once you've pinned a couple of useful Addons, this will probably become the least likely one for you to want to visit, however, which is why I deliberately made it the rightmost. I'll change this if you disagree, but it certainly wasn't an accident.

app/src/main/java/org/xbmc/kore/ui/AddonsOverviewFragment.java
eisnerd: I really puzzled as to what to call these. The existing classes didn't seem to offer a convention that would help distinguish these two. I'll have another think about it, but suggestions would be welcome.

eisnerd:
The focus here is definitely on plugins that present a good virtual file tree. Some don't really show anything at all and some, as you point out, start with a search. Pinning plugins like that would be little use. That is why I added this as an opt-in "pin it if you like" feature instead of trying to load all plugins.

I was not aware that you don't like putting the contents of addons on tabs in the details activity. Have I missed out on a discussion elsewhere? Unlike the "pin" feature, I haven't really given the option for this to be not used for plugins where is doesn't work well. If I can make sure just for this pair of tabs that they're only loaded when first displayed, this will, in my view, fix the problem. Then, selecting the content tab becomes the opt-in action and, for plugins that work better the other way, you can just press launch instead.

eisnerd:
I'll investigate the problem you describe with the ".." parent entry. I didn't look closely at how it's implemented and have only tried a couple of plugins.

app/src/main/java/org/xbmc/kore/ui/AddonsOverviewFragment.java
SyncedSynapse: Yeah, i knew it wasn't an accident, but i really think that the addons list should be the default tab to be shown. It's consistent in every situation. If you pin a couple of addons, only the first one pinned will be immediately available, the rest will also require one more touch to access.

app/src/main/java/org/xbmc/kore/ui/AddonsOverviewFragment.java
SyncedSynapse: How about `AddonListContainerFragment`. Not ideal but at least i think it's less confusing.

SyncedSynapse:
Regarding the tabs automatically starting the addon, i was expecting that a `viewPager.setOffscreenPageLimit(1)` would work but it doesn't (it's the default in fact).

There might be a better way. Addons have `types` (take a look at the `Addon.Details` class that is returned on `GetAddons`) and probably the only addons that are worth navigating are of type `xbmc.addon.video`, `xbmc.addon.music` or `xbmc.addon.pictures`. Will try to confirm this.

eisnerd:
It would be `setOffscreenPageLimit(0)` because the number indicates the off-screen pages to preload and retain, not the total number of pages, but the usual `ViewPager` doesn't support this anyway.

Instead, I am now delaying `browseDirectory` based on `get/setUserVisibleHint`.

SyncedSynapse:
@eisnerd 
`type` info on `AddonDetails` can be used to know which addons are browsable. If this is equal to `xbmc.python.pluginsource` (defined as `XBMC_PYTHON_PLUGINSOURCE` on `AddonType.java` it can be browsed, otherwise it can't.

Ideally this would be checked on `AddonsActivity` and if it is browsable it would load the fragment with the tabs or the Details fragment if it is not.

(if this isn't clear please say so).

eisnerd:
I'm not quite sure what the other addon types might be for, but all the addons that I want to browse at the moment are `xbmc.python.pluginsource`, so I have followed your suggestion.

app/src/main/java/org/xbmc/kore/ui/AddonsActivity.java
```diff
@@ -206,7 +206,11 @@ public void onAddonSelected(AddonListFragment.ViewHolder vh) {
         selectedAddonTitle = vh.addonName;
 
         // Replace list fragment
-        final AddonOverviewFragment addonDetailsFragment = AddonOverviewFragment.newInstance(vh);
+        final SharedElementFragment addonDetailsFragment =
+            vh.browsable
+            ? AddonOverviewFragment.newInstance(vh)
+            : AddonDetailsFragment.newInstance(vh)
+            ;
```
SyncedSynapse: Nice.

eisnerd:
Cool. Thanks for the time you put into advising and improving this with me. There's definitely some more to be done, but this will do for me and hopefully some of the participants to #72 for now.

SyncedSynapse:
Thanks for the contribution. Feel free to hop in anytime.

MartijnKaijser:
screenshot_20160510-231515


You might want to remove the Play button there as it doesn't really have a function I think.

SyncedSynapse:
@MartijnKaijser It starts the addon on Kodi, so it can be useful (and consistent with the rest of the interface).

PR #250, opened 2016-05-31
Fixed issue showing songs without an album or artist

poisdeux:
This should provide a fix for issue #7 

Note that this introduces a database update to include the artist identifier in the songs table. This is needed to be able to retrieve the artist when there is no album associated with the song.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/provider/MediaDatabase.java
+++ b/app/src/main/java/org/xbmc/kore/provider/MediaDatabase.java
@@ -32,7 +32,7 @@ public class MediaDatabase extends SQLiteOpenHelper {
 
 	private static final String DB_NAME = "xbmc.sqlite";
     private static final int DB_VERSION_PRE_EVENT_SERVER = 4,
-            DB_VERSION = 5;
+            DB_VERSION_PRE_SONG_ARTISTS = 5, DB_VERSION = 6;
 
 	/**
 	 * Tables exposed
@@ -48,6 +48,7 @@ public class MediaDatabase extends SQLiteOpenHelper {
         String ARTISTS = "artists";
         String ALBUMS = "albums";
         String SONGS = "songs";
+        String SONG_ARTISTS = "song_artists";
         String AUDIO_GENRES = "audio_genres";
         String ALBUM_ARTISTS = "album_artists";
         String ALBUM_GENRES = "album_genres";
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/xbmc/kore/jsonrpc/type/AudioType.java
```diff
@@ -279,6 +279,7 @@ public DetailsAlbum(JsonNode node) {
         public static final String ALBUMARTIST = "albumartist";
         public static final String ALBUMARTISTID = "albumartistid";
         public static final String ALBUMID = "albumid";
+        public static final String ARTISTID = "artistid";
```
SyncedSynapse: This is not needed. `DetailsAlbum` inherits from `DetailsMedia` which already has `artistid` (populated).

app/src/main/java/org/xbmc/kore/jsonrpc/type/AudioType.java
```diff
@@ -294,6 +295,7 @@ public DetailsAlbum(JsonNode node) {
         public final String album;
         public final List<String> albumartist;
         public final List<Integer> albumartistid;
+        public final List<Integer> artistid;
```
SyncedSynapse: Same comment.

app/src/main/java/org/xbmc/kore/jsonrpc/type/AudioType.java
```diff
@@ -317,6 +319,7 @@ public DetailsSong(JsonNode node) {
             albumid = JsonUtils.intFromJsonNode(node, ALBUMID);
             albumartist = JsonUtils.stringListFromJsonNode(node, ALBUMARTIST);
             albumartistid = JsonUtils.integerListFromJsonNode(node, ALBUMARTISTID);
+            artistid = JsonUtils.integerListFromJsonNode(node, ARTISTID);
```
SyncedSynapse: Same comment

app/src/main/java/org/xbmc/kore/provider/MediaDatabase.java
```diff
@@ -471,6 +480,10 @@ public void onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion) {
                 db.execSQL("ALTER TABLE " + Tables.HOSTS +
                            " ADD COLUMN " + MediaContract.HostsColumns.EVENT_SERVER_PORT +
                            " INTEGER DEFAULT " + HostInfo.DEFAULT_EVENT_SERVER_PORT + ";");
+            case DB_VERSION_PRE_SONGS_NO_ALBUM:
+                db.execSQL("ALTER TABLE " + Tables.SONGS +
+                          " ADD COLUMN "+ MediaContract.SongsColumns.ARTISTID +
+                          " INTEGER DEFAULT 0;");
```
SyncedSynapse: Default to the `albumartistid` column would be better.

app/src/main/java/org/xbmc/kore/provider/MediaDatabase.java
```diff
@@ -346,6 +354,7 @@ public void onCreate(SQLiteDatabase db) {
                    MediaContract.SyncColumns.UPDATED + " INTEGER NOT NULL," +
                    MediaContract.SongsColumns.HOST_ID + " INTEGER NOT NULL " + References.HOST_ID + ", " +
                    MediaContract.SongsColumns.ALBUMID + " INTEGER NOT NULL, " +
+                   MediaContract.SongsColumns.ARTISTID + " INTEGER NOT NULL, " +
```
SyncedSynapse: Maybe i'm missing something, but you're creating an Integer column, and them you're updating it with a string, which is a concatenation of the artistid's for the song. If there's more than one, this will probably crash.

SyncedSynapse:
See my comments in the code.

The issues come from the fact that `artistid` at the song level should be treated the same way as at the album level: there can be more than one artist, so there should be a many to many relationship - creating a `song_artists` tables.

As is, i suspect that it might crash if there's more than one artist for a song (didn't test it, but unless i'm missing something, it's likely).

.travis.yml
```diff
@@ -6,11 +6,16 @@ env:
...
     # needed build tools
-    - build-tools-22.0.1
+    - build-tools-23.0.3
 
     # The SDK version used to compile your project
-    - android-22
+    - android-23
 
     # extra needed components
     - extra-android-m2repository
```
SyncedSynapse: How about making this a separate PR.

poisdeux: Done: #252

app/src/main/java/org/xbmc/kore/provider/MediaDatabase.java
poisdeux: Didn't explicitly test when a song has multiple artists. So, if I understand correctly, this can happen when a song is part of an album that has multiple artists, right?

app/src/main/java/org/xbmc/kore/jsonrpc/type/AudioType.java
poisdeux: Oops, overlooked the inheritance.

app/src/main/java/org/xbmc/kore/provider/MediaDatabase.java
SyncedSynapse: Probably, but the main issue is that in Kodi the connection between songs and artists is many to many and this model should respect that (see main comment)

poisdeux:
I refactored my solution according to your suggestion, introducing the song_artists table to implement the many-to-many relationship.

app/src/main/java/org/xbmc/kore/service/library/SyncMusic.java
```diff
@@ -347,12 +348,28 @@ public void onSuccess(ApiList<AudioType.DetailsSong> result) {
...
+                for (AudioType.DetailsSong song : items) {
+                    for (int artistId : song.artistid) {
+                        songArtistsValuesBatch[artistCount] = new ContentValues();
+                        songArtistsValuesBatch[artistCount].put(MediaContract.SongArtists.HOST_ID, hostId);
+                        songArtistsValuesBatch[artistCount].put(MediaContract.SongArtists.SONGID, song.songid);
+                        songArtistsValuesBatch[artistCount].put(MediaContract.SongArtists.ARTISTID, artistId);
+                        artistCount++;
+                    }
+                }
+
```
SyncedSynapse: 1. I suggest changing the `artistsCount`  (plural) variable to `totalArtistsCount` to make it distinct from `artistCount`. It was confusing me.

2. You're populating `songArtistsValuesBatch`, but you're not doing anything with it! The songs aren't getting saved to the DB, and no songs are appearing on the UI. You might have missed this because you have a version with the `songsArtists` table filled, but you should test the common scenario, which is someone who is upgrading from a DB without that table.

poisdeux: Oops. This was a left over from my previous solution. I removed it.

SyncedSynapse: Wait, i'm not following you... So where are you populating the `songsArtists` table?

poisdeux: OK, I messed up. I originally implemented the solution on #251 and extracted the solution without the unittest into this PR. As it has been a while since I looked at this code I need more time to get this right. I'll post a comment when (I think) I fixed the code. This also puts #251 on hold.

SyncedSynapse: I believe you just think needed a `contentResolver.bulkInsert` (before you removed the for that is)

poisdeux: You're right. I updated the code and added the bulkInsert. Sorry about the mess. 

(I don't have a lot of time at the moment. After your first comment I quickly compared this with #251. I overlooked the code refactoring in #251 which adds insertSongs so it can be used from the tests as well. This made me think the above code was a leftover from the previous solution)

PR #223, opened 2016-03-17
Implemented unittests to test restoring search query

poisdeux:
* Added perl tools to generate json files for music and movie data
* Upgraded build tools version, sdk compile version, and support libraries
  as this is required by the test packages.
* Added new product flavor unittest to make sure assets and settings required
  for testing do not also ship with a release version of Kore

PR #251, opened 2016-05-31
Unittest/mediaprovider

poisdeux:
In order to test PR #250 I created an integration test for mediaprovider. It uses the same data as #223


SyncedSynapse:
PR #250 is included here... Shouldn't.

poisdeux:
I know. It is kinda hard to keep them separated as both the fix and the test for the fix require the database upgrade. Do you want a separate PR that includes an integration test for the MediaProvider as it is currently implemented on master? Then, this PR could hold only the added tests for #250

SyncedSynapse:
@poisdeux Not relevant anymore, right?

poisdeux:
No, it is still relevant. I just had not updated this with the refactored solution from #250 as I still haven't found a good way to keep #250 and this mutually exclusive.
(But if you merge #250 first I can just rebase this and your first remark vanishes into thin air ;) )

MartijnKaijser:
can we merge this?

poisdeux:
AFAICT this can be merged. Just rebased it and ran the checks to be sure it still works.  But it would be nice if @SyncedSynapse could review this first. This PR does not really fix or add anything for end users so I guess this PR can wait until @SyncedSynapse has time to do the review.

app/src/main/java/org/xbmc/kore/service/library/SyncMusic.java
```diff
@@ -151,6 +146,8 @@ private void deleteMusicInfo(final ContentResolver contentResolver,
                                where, new String[]{String.valueOf(hostId)});
         contentResolver.delete(MediaContract.AlbumGenres.CONTENT_URI,
                                where, new String[]{String.valueOf(hostId)});
+        contentResolver.delete(MediaContract.SongArtists.CONTENT_URI,
+                               where, new String[]{String.valueOf(hostId)});
```
SyncedSynapse: errr....

This delete should have been included in the PR that introduced songsArtists...

Without it, every library sync will duplicate the values on that table. And this version is on prod... I haven't seen any complaints, but this sure looks like it will be duplicating entries.

app/src/main/java/org/xbmc/kore/service/library/SyncMusic.java
```diff
@@ -348,13 +301,7 @@ public void onSuccess(ApiList<AudioType.DetailsSong> result) {
...
                 int totalArtistsCount  = 0;
                 // Save partial results to DB
-                ContentValues songValuesBatch[] = new ContentValues[items.size()];
-                for (int i = 0; i < items.size(); i++) {
-                    AudioType.DetailsSong song = items.get(i);
-                    songValuesBatch[i] = SyncUtils.contentValuesFromSong(hostId, song);
-                    totalArtistsCount += song.artistid.size();
-                }
-                contentResolver.bulkInsert(MediaContract.Songs.CONTENT_URI, songValuesBatch);
+                insertSongsItems(items, contentResolver);
```
SyncedSynapse: I believe that if this is applied as is, there will be duplicate code.

After this line, there's currently code to insert the `songsArtists`, which will be kept, as this PR doesn't delete it. But that code is also included in the `insertSongsItems` implementation.

app/src/main/java/org/xbmc/kore/service/library/SyncMusic.java
```diff
@@ -384,12 +331,98 @@ public void onSuccess(ApiList<AudioType.DetailsSong> result) {
...
+            for (int artistId : song.artistid) {
+                songArtistsValuesBatch[artistCount] = new ContentValues();
+                songArtistsValuesBatch[artistCount].put(MediaContract.SongArtists.HOST_ID, hostId);
+                songArtistsValuesBatch[artistCount].put(MediaContract.SongArtists.SONGID, song.songid);
+                songArtistsValuesBatch[artistCount].put(MediaContract.SongArtists.ARTISTID, artistId);
+                artistCount++;
+            }
+        }
+
+        contentResolver.bulkInsert(MediaContract.SongArtists.CONTENT_URI, songArtistsValuesBatch);
```
SyncedSynapse: Possible duplicate code, see previous comment

SyncedSynapse:
Didn't look  at the test bits, but i have some questions regarding the parts that impact the Kore core. @poisdeux care to take a look.

app/src/main/java/org/xbmc/kore/service/library/SyncMusic.java
MartijnKaijser: Guess it should be patched and updated to prod?

poisdeux: Argh, yes this should have been in the original PR...
Fortunately, the table song_artists requires the combination host_id, songid, and artistid to be unique. 
( UNIQUE (host_id, songid, artistid) ). This means, during additional syncs the entries will simply replace the current entries.

app/src/main/java/org/xbmc/kore/service/library/SyncMusic.java
poisdeux: Ah, yes. Missed that one. Thx!

app/src/main/java/org/xbmc/kore/service/library/SyncMusic.java
poisdeux: Should be fixed now

poisdeux:
@SyncedSynapse I should be OK now. Can you take another look to see if I didn't miss anything (this time...)

SyncedSynapse:
Can't right now. Will be back next week but not sure if I'll have time to
take a look.

PR #253, opened 2016-06-04
Added file size to media file list view

tomerf:


SyncedSynapse:
I get a lot of nulls on the file size... Maybe present the file size only if `itemFileSize > 0`

PR #254, opened 2016-06-14
Sync delay steps

tomerf:
Sometime when I have a wrong version of subtitles it often gets out of sync and there's no easy way to adjust it. Since I don't know how much to delay I have to do several trial and error sessions and each one takes clicking through a long series of menus.

I'm not sure what's the best solution here, so I'll describe what I did/want.

The current implementation opens the Kodi slider where you can offset by a small step (0.1s) and only see the results once you click ok.
My idea to fix it is to add 2 buttons (+/- delay) on the Kore's subtitles submenu that will allow to quickly adjust the subtitles by a more significant step and have immediate feedback (the finer control is still available).
I added +0.5s and -0.5s buttons using the dialog's ok/cancel buttons.
I still want to keep the dialog visible for further adjustments after clicking, but even without calling dismiss() explicitly it still disappear (why?).

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/GenericSelectDialog.java
+++ b/app/src/main/java/org/xbmc/kore/ui/GenericSelectDialog.java
@@ -143,6 +143,21 @@ public class GenericSelectDialog
                                 mListener.onDialogSelect(token, which);
                             dialog.dismiss();
                         }
+                    })
+                    .setPositiveButton("+0.5s >", new DialogInterface.OnClickListener() {
+                        @Override
+                        public void onClick(DialogInterface dialog, int which) {
+                            if (mListener != null)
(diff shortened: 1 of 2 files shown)


SyncedSynapse:
> The current implementation opens the Kodi slider where you can offset by a small step (0.1s) and only see the results once you click ok.

Are you sure about this? I don't need to click ok to see the changes in the delay. They are not immediate, it can take a couple of seconds for Kodi to reflect the delay, but i don't have to click anything for the delay to be applied. 

I understand your goal here, i also strugle with subtitle/audio delays, but i'm not sure that this is the right solution (even if the dialog wasn't being dismissed). 

Check the behavior above, and maybe you won't need this, as there are several issues here that are somewhat tricky.

PR #257, opened 2016-06-20
Fixed printing false message during gradle configuration phase

poisdeux:
The println is invoked during gradle's configuration phase as well. During the configuration phase all projects are executed. This created some confusion, as it looked like the SET_ANIMATION_SCALE was granted for all flavors and not just the instrumentationTest flavor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,4 @@ android:
 before_script:
 
 script: 
-  - ./gradlew assembleRelease
+  - ./gradlew assembleFullRelease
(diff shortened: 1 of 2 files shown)


SyncedSynapse:
@poisdeux Want to check the travis build or can i safely merge this?

poisdeux:
The instrumentationTest flavor failed. Can't really tell from the travis log what's wrong. It's their script that crashes. As they still consider instrumentation testing experimental I don't wanna waste much time on it. Fixed the travis build by building the full flavor only.

PR #259, opened 2016-06-29
squid:S1854 - Dead stores should be removed

georgekankava:
This pull request is focused on resolving occurrence of Sonar rule
squid:S1854 - Dead stores should be removed.
You can find more information about the issue here:
https://dev.eclipse.org/sonar/rules/show/squid:S1854
Please let me know if you have any questions.
George Kankava

### DIFF
--- a/app/src/main/java/org/xbmc/kore/host/HostManager.java
+++ b/app/src/main/java/org/xbmc/kore/host/HostManager.java
@@ -132,7 +132,6 @@ public class HostManager {
                 while (cursor.moveToNext()) {
                     int idx = 0;
                     int id = cursor.getInt(idx++);
-                    long updated = cursor.getLong(idx++);
                     String name = cursor.getString(idx++);
                     String address = cursor.getString(idx++);
                     int protocol = cursor.getInt(idx++);
@@ -143,7 +142,7 @@ public class HostManager {
                     String macAddress = cursor.getString(idx++);
                     int wolPort = cursor.getInt(idx++);
                     boolean useEventServer = (cursor.getInt(idx++) != 0);
-                    int eventServerPort = cursor.getInt(idx++);
+                    int eventServerPort = cursor.getInt(idx);
 
                     hosts.add(new HostInfo(id, name, address, protocol, httpPort, tcpPort,
                             username, password, macAddress, wolPort, useEventServer, eventServerPort));
--- a/app/src/main/java/org/xbmc/kore/ui/AddonOverviewFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AddonOverviewFragment.java
@@ -110,7 +110,7 @@ public class AddonOverviewFragment extends SharedElementFragment {
         long baseFragmentId = 1000;
         tabsAdapter = new TabsAdapter(getActivity(), getChildFragmentManager())
                 .addTab(AddonDetailsFragment.class, args, R.string.addon_overview, baseFragmentId++)
-                .addTab(MediaFileListFragment.class, contentArgs(args), R.string.addon_content, baseFragmentId++)
+                .addTab(MediaFileListFragment.class, contentArgs(args), R.string.addon_content, baseFragmentId)
                 ;
         viewPager.setAdapter(tabsAdapter);
         pagerTabStrip.setViewPager(viewPager);


app/src/main/java/org/xbmc/kore/service/library/SyncMovies.java
```diff
@@ -163,7 +163,6 @@ public void onSuccess(List<VideoType.DetailsMovie> result) {
                     // As we're going to recurse, these result objects can add up, so
                     // let's help the GC and indicate that we don't need this memory
                     // (hopefully this works)
-                    result = null;
```
poisdeux: Did you read the comment above this statement?
Are you sure setting result to null is of no use at all?

georgekankava: @poisdeux Yes, there is no need to set it to null because it is out of scope after this method ends and candidate for GC removal.

georgekankava: I will remove comments as well about such fields in this pull request.

poisdeux: As long as the method is recursing it will be in scope. As it is used to sync items over a network recursing may take a while. I believe it was originally added to allow the GC to cleanup already handled results while the method is still recursing.

BTW: remove the comment as well, as it doesn't make sense anymore.

georgekankava:
@poisdeux PR updated.

app/src/main/java/org/xbmc/kore/service/library/SyncMovies.java
```diff
@@ -159,11 +159,6 @@ public void onSuccess(List<VideoType.DetailsMovie> result) {
 
                 LogUtils.LOGD(TAG, "syncAllMovies, movies gotten: " + result.size());
                 if (result.size() == LIMIT_SYNC_MOVIES) {
-                    // Max limit returned, there may be some more movies
-                    // As we're going to recurse, these result objects can add up, so
-                    // let's help the GC and indicate that we don't need this memory
-                    // (hopefully this works)
-                    result = null;
```
poisdeux: Unless you can clearly explain that setting this to null is useless I suggest we keep it.

georgekankava: @poisdeux Yes I am also for keeping this if you say that this is some special case. I will remove this changes.

georgekankava:
@poisdeux PR updated.

app/src/main/java/org/xbmc/kore/service/library/SyncMusic.java
```diff
@@ -284,7 +283,6 @@ public void onSuccess(ApiList<AudioType.DetailsAlbum> result) {
 
                 if (moreItemsAvailable(limitsReturned)) {
                     LogUtils.LOGD(TAG, "chainCallSyncAlbums: More results on media center, recursing.");
-                    result = null; // Help the GC?
```
poisdeux: See previous comment

app/src/main/java/org/xbmc/kore/service/library/SyncMusic.java
```diff
@@ -355,7 +353,6 @@ public void onSuccess(ApiList<AudioType.DetailsSong> result) {
 
                 if (moreItemsAvailable(limitsReturned)) {
                     LogUtils.LOGD(TAG, "chainCallSyncSongs: More results on media center, recursing.");
-                    result = null; // Help the GC?
```
poisdeux: See previous comment

georgekankava:
@poisdeux Updated.

app/src/main/java/org/xbmc/kore/host/HostManager.java
```diff
@@ -132,7 +132,6 @@ public static HostManager getInstance(Context context) {
                 while (cursor.moveToNext()) {
                     int idx = 0;
                     int id = cursor.getInt(idx++);
-                    long updated = cursor.getLong(idx++);
```
poisdeux: This should not be removed or you must make sure idx is incremented correctly to get the next element.

poisdeux:
In general, automated code analysis tools usually are not very good at parsing logic. So, you should always check if the code improvements do not actually break the code.

SyncedSynapse:
The `idx++` that are being removed are indeed unnecessary, but they might prevent bugs in the future, when someone comes to this piece of code and just copy pastes the previous line. I sacrifice some superfluous `incs` on a non critical code section for the possibility of avoiding a bug in the future.

PR #262, opened 2016-06-29
squid:S1125 - Literal boolean values should not be used in condition expressions

georgekankava:
This pull request is focused on resolving occurrences of Sonar rule
squid:S1125 - Literal boolean values should not be used in condition expressions.
You can find more information about the issue here:
https://dev.eclipse.org/sonar/rules/show/squid:S1125
Please let me know if you have any questions.
George Kankava


poisdeux:
In this particular case I do not agree. The ! is easily overlooked.

SyncedSynapse:
I agree with @poisdeux , and this checks aren't really very relevant IMO...

Issue #104, opened 2015-07-22
“Various Artists” albums don't show song artists


axelsimon:
Hi,

Right now, if an album is marked as Various Artists, Kore will show the following when showing the album: tracknumber, title and track length.

But not the artist.

It is often hard to know what song you're about to play without the artist's name ;)

I think putting the artist's name where the track length currently is and moving track length to the right of that should work nicely to solve the issue.

Thanks!

axel

SyncedSynapse:
Done in #281

PR #281, opened 2016-08-24
Implemented displaying artist for each song in album details view

poisdeux:
This should fix issue #104 

Artist is now displayed below song title in the list. Duration is displayed below artist.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/provider/MediaProvider.java
+++ b/app/src/main/java/org/xbmc/kore/provider/MediaProvider.java
@@ -23,7 +23,6 @@ import android.database.Cursor;
 import android.database.sqlite.SQLiteDatabase;
 import android.net.Uri;
 import android.provider.BaseColumns;
-import android.support.annotation.Nullable;
 
 import org.xbmc.kore.utils.LogUtils;
 import org.xbmc.kore.utils.SelectionBuilder;
@@ -671,8 +670,9 @@ public class MediaProvider extends ContentProvider {
                               .mapToTable(MediaContract.Songs.SONGID, MediaDatabase.Tables.SONGS)
                               .mapToTable(MediaContract.Songs.TITLE, MediaDatabase.Tables.SONGS)
                               .mapToTable(MediaContract.Songs.ALBUMID, MediaDatabase.Tables.SONGS)
-                              .mapToTable(MediaContract.Songs.UPDATED, MediaDatabase.Tables.SONGS)
                               .mapToTable(MediaContract.Songs.THUMBNAIL, MediaDatabase.Tables.SONGS)
+                              .mapToTable(MediaContract.Songs.DISPLAYARTIST, MediaDatabase.Tables.SONGS)
+                              .mapToTable(MediaContract.AlbumArtists.ARTISTID, MediaDatabase.Tables.ALBUM_ARTISTS)
                               .mapToTable(MediaContract.SongArtists.ARTISTID, MediaDatabase.Tables.SONG_ARTISTS)
                               .where(Qualified.SONGS_HOST_ID + "=?", hostId)
                               .groupBy(Qualified.SONGS_SONGID);
@@ -681,8 +681,8 @@ public class MediaProvider extends ContentProvider {
                 final String hostId = MediaContract.Hosts.getHostId(uri);
                 final String albumId = MediaContract.Albums.getAlbumId(uri);
                 return builder.table(MediaDatabase.Tables.SONGS)
-                              .where(MediaContract.Songs.HOST_ID + "=?", hostId)
-                              .where(MediaContract.Songs.ALBUMID + "=?", albumId);
+                              .where(Qualified.SONGS_HOST_ID + "=?", hostId)
+                              .where(Qualified.SONGS_ALBUMID + "=?", albumId);
             }
             case SONGS_ID: {
                 final String hostId = MediaContract.Hosts.getHostId(uri);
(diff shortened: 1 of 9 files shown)


SyncedSynapse:
Not sure about this one. In the most common case, the artist name will just be duplicated on each song, which is kind of wasteful (and needs 3 lines for each song).
Ideally this would just show in albums that have various artists. On a quick look i couldn't find any attribute on Album that indicates that, but i'm probably missing something.
A quick and dirty way to do it is to compare the song artist with the album artist and show the song artist only if it's different.
And maybe we could put the artist and the duration on the same line to save some space.

poisdeux:
I agree with your concern that the proposed solution is a bit wasteful.

As I see it we have three options to fix #104:

1. Always show artist for each song
2. Only show artist for songs that differ from the album artist
3. Only show artist for each song on albums with multiple artists

I think option 2 will be very confusing and therefore I do not consider that as a solution at all.
Option 3 would be best, however also more complicated to implement. 
I chose option 1 as I didn't find the redundant displaying of the artist annoying at all. Most devices have a high resolution screen nowadays and the artist is shown in a smaller size than the song title.
I wasn't too sure about combining the duration and artist on a single line. Didn't look right to me. 
But i feel different now. The following might just work:

1.  Kodi rocks!
     1:36 | Team Kodi
2. Kore rocks as well
     2:23 | Team Kodi
3. Yatse who?
     1:12 | Tolriq

Thumbs up for option 1 with duration and artist on a single line?

SyncedSynapse:
I think that combining the duration and artist on a single line is a good suggestion. Ideally, the artist should be on the same line as the title, but that won't fit for a lot of cases, so putting it on the second line is a sensible compromise.

But i don't share your view regarding option 2 (only showing the artist if its different). I don't think it will cause unnecessary confusion.

I don't have strong feelings about this, so i'm not opposed to always showing the artist, i just think that showing it only when it is different would be better.

@MartijnKaijser can we have your opinion here (we're talking about the album details screen).

DaveTBlake:
Martijn prompted me to comment, and please ping me for any future music stuff :)
I would be strongly in favour of showing song artist(s) for each track. It may just duplicate the album artist, but it is not only various artist compliation albums that have different song artist(s) to that of the album e.g. featured artists on just one track, duets etc.  also hsitorically many users tag classical music with album artist = conductor/ orchestra but song artist = composer name (was the only way to get players to show composer).

In JSON GetSongs "displayartist" is a string that gives the artist name(s) as they were tagged in the music file. I suggest using that.  

For interest from v16 onwards (and processing of ARTISTS tag) the "displayartist" is not always the same as the union of the list of artist names. For example we could have 
displayartist = "artist1 feat. artist2"
artist = {"artist1", "artist2"}

SyncedSynapse:
So be it :)
Let's show the display artist next to the duration on each song.

poisdeux:
Updated PR implementing displaying the artist next to the duration for each song.

 AFAICT we only include **artist** and not **display artist** in the database ([AudioType#DetailsArtist](https://github.com/xbmc/Kore/blob/master/app/src/main/java/org/xbmc/kore/jsonrpc/type/AudioType.java#L149)). So we can't show the display artist, only the artist. Correct?

Note, this now also includes a fix for the integration test.

DaveTBlake:
I don't know my way around the Kore code so my comments may not help, but "displayartist" is part of  Audio.Fields.Song not an artist property. Remember a song can have many artists too.

poisdeux:
We don't use it in songs as well ([AudioType#DetailsSong](https://github.com/xbmc/Kore/blob/master/app/src/main/java/org/xbmc/kore/jsonrpc/type/AudioType.java#L318)).
Just to be sure: we should use **displayartist** to show the artist to the user and we should use **artist**/**albumartist** in search filters?

DaveTBlake:
Strictly yes I believe that is correct. Before v16 displayartist was always a simple union of the array in artist, but that is no longer true. Of course they will almost always be closely related and a simple string search would produce similar results.

Out of interest, with multiple artists for a song what is Kore storing?

poisdeux:
We assume we get ALL artists from querying **AudioLibrary.GetArtists** and connect the song to the correct artist using its **artistid**. @SyncedSynapse please correct me if I'm wrong.

SyncedSynapse:
@poisdeux 
`displayartist` is available at the song level. It's on https://github.com/xbmc/Kore/blob/master/app/src/main/java/org/xbmc/kore/jsonrpc/type/AudioType.java#L74 which is inherited by `DetailsSong`

poisdeux:
Ah I see. Not pretty, but I guess we need to add display artist to the songs table now, to display the artist correctly for songs. Correct?

SyncedSynapse:
yeah, `displayartist` needs to be added to the `Songs` table. 

I've also noticed that, as this PR stands, if a song has more than one artist it will appear several times (it's getting data from `SONGS_FOR_ARTIST_AND_OR_ALBUM_JOIN`), which is not ideal...

poisdeux:
Updated PR. We now only show the display artist. Assuming that it will always be set if an artist is available. Is that a valid assumption? 
If not, we need to check if display artist is set, if not revert to album artist, if that is not set as well, revert to artist...

Added display artist column to songs table. This required a DB upgrade and users should
do a refresh for music items to be able to see the display artist in the song lists.

Updated SONGS_FOR_ARTIST_AND_OR_ALBUM_JOIN to also get songs that are part of an album for which the album artist matches the searched artist but the song has a different artist.

poisdeux:
Hmm, bumped into a problem. Song with an artist but without an album is not correctly returned.
Working on it...

poisdeux:
Fixed! @SyncedSynapse can you do a review one more time...

app/src/main/java/org/xbmc/kore/provider/MediaProvider.java
```diff
@@ -737,11 +737,14 @@ private SelectionBuilder buildQuerySelection(Uri uri, int match) {
                               .mapToTable(MediaContract.Songs.SONGID, MediaDatabase.Tables.SONGS)
                               .mapToTable(MediaContract.Songs.TITLE, MediaDatabase.Tables.SONGS)
                               .mapToTable(MediaContract.Songs.ALBUMID, MediaDatabase.Tables.SONGS)
-                              .mapToTable(MediaContract.Songs.UPDATED, MediaDatabase.Tables.SONGS)
-                              .mapToTable(MediaContract.Songs.THUMBNAIL, MediaDatabase.Tables.SONGS)
```
SyncedSynapse: Any reason for leaving this out?
Not really relevant i guess, but was wondering why.

SyncedSynapse:
Looks good, just did a comment because i didn't understand why you removed some lines, but i guess they are not relevant.

app/src/main/java/org/xbmc/kore/provider/MediaProvider.java
poisdeux: Wasn't really necessary, but while fixing #104 I also did some cleaning.

PR #283, opened 2016-08-26
Implemented recursion fix for movies and tv shows

poisdeux:
This applies the fix for issue #219 Incomplete libraries
to movies and tv shows.
For music items this was already fixed in PR #236 
See commit 7330f85241920e0ae9a5bec7719c497998fb070c


SyncedSynapse:
@poisdeux This is generating some conflicts. `git apply` this on the most recent branch won't work.
I've tried `patch -p1 < diff` and it was able to patch it, but i would like you to take a look first.

poisdeux:
Fixed.

Issue #285, opened 2016-09-12
Kodi v17 changed name of the player overlay showing codecs and such info 


MartijnKaijser:
For v17 a different window should be triggered for showing the codec/video info which is under longpress  button

SyncedSynapse:
b7ece61951590b6adbd0bcad04a8f61df6574566 and 18811371a7df8f9942d15d6a9c4bf94cada45bde should fix this.

MartijnKaijser:
Works. Thx

PR #286, opened 2016-09-14
Fix for long press Info button to display codec info in v17

BSunk:
Fix for issue #285.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/jsonrpc/method/Input.java
+++ b/app/src/main/java/org/xbmc/kore/jsonrpc/method/Input.java
@@ -59,6 +59,7 @@ public class Input {
         public final static String SHOWSUBTITLES = "showsubtitles";
         public final static String NEXTSUBTITLE = "nextsubtitle";
         public final static String CODECINFO = "codecinfo";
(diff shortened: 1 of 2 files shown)


SyncedSynapse:
Thanks. I'm not going to merge this right now because i'm thinking that this sort of issues will happen in the future, so Kodi's version should be present in the local database, to avoid always calling `GetProperties` (and it should be kept up to date seamlessly). Then this PR should just check the `host` version, not needing to call `GetProperties`.

MartijnKaijser:
It should be a one time check imo on initial connection and then it should save it until next connection.

SyncedSynapse:
Ideally yes, but on Kore you don't keep an active connection. Whenever you switch screens (from the remote to any other screen for instance), exit Kore, etc, it closes the connection, and we shouldn't update that info on every connection start.
Maybe keep a date reflecting when the info was last updated and update it on a daily basis.

poisdeux:
We might also naively use the old method and depend on receiving an error response whenever the method isn't implemented. Then update the version and try again. If it still fails we should notify the user.

DaveTBlake:
Why check Kodi version? Why not use the JSON API version instead? Isn't that part of why such version control exists?

PR #292, opened 2016-09-29
Support playback of shared plain http-video urls

noctux:
Currently Kore does only support the playback of youtube and vimeo video
urls when invoked via the "Play with Kodi" share menu.

This commit adds support for plain video urls in the most straightforward way
possible.
The fix is rather minimal (actually it more seems like removing an artificial limitation?),
but I would be happy to adapt the patch based on any feedback provided :)

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/RemoteActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/RemoteActivity.java
@@ -386,16 +386,17 @@ public class RemoteActivity extends BaseActivity
         }
 
         final String videoId = getVideoId(videoUri);
-        if (videoId == null) {
-            Toast.makeText(RemoteActivity.this,
-                    R.string.error_share_video,
-                    Toast.LENGTH_SHORT).show();
-            return;
-        }
-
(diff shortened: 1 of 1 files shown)


SyncedSynapse:
Hi,
Can you give a URL example to test this? Where do you find this useful?

noctux:
Example url: [http://ftp.nluug.nl/pub/graphics/blender/demo/movies/Sintel.2010.720p.mkv]()

More generally speaking, it works for all urls leading to html5 videos. This includes the mobile versions of the "mediatheken" of several german broadcasters. I open the video in firefox mobile and then share that url with core via "Play on Kodi".

My second (primary?) usecase is easy playback of videos from my computer. I tend to issue a quick "thttpd -D -d . -p 8080" over a directory containing videos related to a certain event (e.g. holiday videos, drone flights, etc." and then use the procedure described above to select the specific video via the webbrowser and then share it with kodi for playback. This is intended as a sort of temporary sharing while I do not necessarily want to have those videos in my kodi collection etc...

SyncedSynapse:
Ok thanks

noctux:
Thanks for merging

Issue #199, opened 2016-02-07
Mute button hard to find and not highlighted


lebeno:
When Kodi has been muted and there is no media playing, I can find no (un)mute button or indication in Kore. When there is media playing, the (un)mute button is only visible on the "now playing" screen, not on the "remote" screen and it is not highlighted to indicate the muted state when Kodi was muted before the current playback started.

I did spend some time going through different screens to find the (un)mute button, only on one of the three "remote" screens. Although Kodi displays a very visible red "muted" icon in the top right corner of the screen, I regularly use Kore as a headless remote when playing music. A more prominent mute state button would be handy, certainly in the case where Kore is used to control Kodi without a screen.

Steps to reproduce: Play some music, mute Kodi in the "now playing"screen of Kore, stop all playback. Go to music, start playback of any artist or album. Go to Menu > Remote and there is no (un)mute button or indication. When you swipe left the (un)mute button is visible but not highlighted to indicate the currently muted state.

RogerTheShrubber1:
F8 key will toggle mute on/off

PR #294, opened 2016-10-03
Check the status of the mute button in NowPlaying

tomerf:
Check the status of the mute button when moving to Now Playing screen, fixing issue #199

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/NowPlayingFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/NowPlayingFragment.java
@@ -222,6 +222,16 @@ public class NowPlayingFragment extends Fragment
     public void onResume() {
         super.onResume();
         hostConnectionObserver.registerPlayerObserver(this, true);
+        Application.GetProperties action = new Application.GetProperties(Application.GetProperties.MUTED);
+        action.execute(hostManager.getConnection(), new ApiCallback<ApplicationType.PropertyValue>() {
+            @Override
+            public void onSuccess(ApplicationType.PropertyValue result) {
+                setVolumeMuteButton(result.muted);
+            }
+
+            @Override
+            public void onError(int errorCode, String description) { }
+        }, callbackHandler);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/xbmc/kore/ui/NowPlayingFragment.java
```diff
@@ -311,18 +321,7 @@ public void onVolumeMuteClicked(View v) {
...
-                    TypedArray styledAttributes = theme.obtainStyledAttributes(new int[] {
-                            R.attr.colorAccent});
-                    volumeMuteButton.setColorFilter(
-                            styledAttributes.getColor(0,
-                                    getActivity().getResources().getColor(R.color.accent_default)));
-                    styledAttributes.recycle();
-                } else {
-                    volumeMuteButton.clearColorFilter();
-                }
+                setVolumeMuteButton(!result);
```
SyncedSynapse: I haven't tested this yet, but something looks odd...
Why are you passing `!result` to `setVolumeMuteButton` when the code inside it is exactly the same that was removed from here (which relied on `result`)?

SyncedSynapse: Hmmm, this is working the opposite for me (which is consistent with the previous comment).
When you click the mute button on Kore it mutes Kodi, but in Kore's interface it appears as not muted (the button isn't highlighted). Maybe you were assuming that if the button is highlighted it means that it's not muted but it's the other way around.

Looking into this i was wondering if this check (the call to application.GetProperties to check the muted status) should be done in the HostConnectionObserver, so that it gets done regularly, to always keep Kore in sync with Kodi (this PR only keeps it in sync in the `OnResume` call). But it's one more call to Kodi (there are 3 already) being done regularly (if connecting through HTTP, through TCP we needed to implement the `Application.OnVolumeChanged` which isn't being acted upon right now), so i'm not sure if it should be done... People might have noticed that the interface isn't in sync but i haven't heard any complains so it might not be worth doing it regularly. 
Right now, doing it in the `OnResume` handler looks ok.

poisdeux: @SyncedSynapse yup, it is. I use the same solution @tomerf implemented in this PR to set the volume state on resume. But I also use the HostConnectionObserver to set the volume state as soon as it changes. BTW, I did not have to take the complement of the mute state. Strange that @tomerf does need it...
 
@tomerf I'm currently working on showing the volume level on the NowPlayingFragment. It also includes setting the mute state. 
See https://github.com/poisdeux/Kore/tree/feature/%2330
I can rebase my branch on this PR when you fix it.
But I'm not going to wait for it...

tomerf:
Fixed.

PR #301, opened 2016-10-18
Local play: Support in-device video playing as 2nd screen

yytsui:
# Feature
1. Add a new play button in Movie detail view.
2. When user click the button, if the video is accessible , a video playing intent is initiated. 
   A list of other in-device video player apps pops up, then user can choose one of them to play video.
# Comment
1. A different button may be needed to give user cue that it's playing in-device instead of the Kodi  host.
2. The same feature will be implemented for TV Show later.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
@@ -436,17 +436,10 @@ public class MovieDetailsFragment extends AbstractDetailsFragment
             return;
         }
 
-        HostInfo hostinfo = getHostInfo();
-        String pathforUrl = movieDownloadInfo.fileName.replaceAll(" ", "%20").replaceAll("'","%27");
-        String videoUrl = String.format("http://%s:%d/vfs/%s", hostinfo.getAddress(),
-                    hostinfo.getHttpPort(), pathforUrl);
-
+        String videoUrl =  movieDownloadInfo.getMediaUrl(getHostInfo());
         Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(videoUrl));
         intent.setDataAndType(Uri.parse(videoUrl), "video/*");
         startActivity(intent);
-
-
-
     }
 
     @Override
--- a/app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/MovieDetailsFragment.java
@@ -18,6 +18,7 @@ package org.xbmc.kore.ui;
 import android.annotation.TargetApi;
 import android.app.AlertDialog;
 import android.content.DialogInterface;
+import android.content.Intent;
 import android.content.res.Resources;
 import android.content.res.TypedArray;
 import android.database.Cursor;
@@ -48,6 +49,7 @@ import com.melnykov.fab.ObservableScrollView;
 
 import org.xbmc.kore.R;
 import org.xbmc.kore.Settings;
+import org.xbmc.kore.host.HostInfo;
 import org.xbmc.kore.jsonrpc.ApiCallback;
 import org.xbmc.kore.jsonrpc.event.MediaSyncEvent;
 import org.xbmc.kore.jsonrpc.method.Player;
@@ -114,6 +116,7 @@ public class MovieDetailsFragment extends AbstractDetailsFragment
     @InjectView(R.id.go_to_imdb) ImageButton imdbButton;
     @InjectView(R.id.download) ImageButton downloadButton;
     @InjectView(R.id.seen) ImageButton seenButton;
+    @InjectView(R.id.local_play) ImageButton localPlayButton;
 
     // Detail views
     @InjectView(R.id.media_panel) ScrollView mediaPanel;
@@ -423,6 +426,27 @@ public class MovieDetailsFragment extends AbstractDetailsFragment
         // Change the button, to provide imeddiate feedback, even if it isn't yet stored in the db
         // (will be properly updated and refreshed after the refresh callback ends)
         setupSeenButton(newPlaycount);
+    }
+
+    @OnClick(R.id.local_play)
+    public void onLocalPlayClicked(View v) {
+        if (movieDownloadInfo == null) {
+            // Nothing to play on local
+            Toast.makeText(getActivity(), R.string.no_files_to_play, Toast.LENGTH_SHORT).show();
+            return;
+        }
+
+        HostInfo hostinfo = getHostInfo();
+        String pathforUrl = movieDownloadInfo.fileName.replaceAll(" ", "%20").replaceAll("'","%27");
+        String videoUrl = String.format("http://%s:%d/vfs/%s", hostinfo.getAddress(),
+                    hostinfo.getHttpPort(), pathforUrl);
+
+        Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse(videoUrl));
+        intent.setDataAndType(Uri.parse(videoUrl), "video/*");
+        startActivity(intent);
+
+
+
     }
 
     @Override
--- a/app/src/main/java/org/xbmc/kore/utils/FileDownloadHelper.java
+++ b/app/src/main/java/org/xbmc/kore/utils/FileDownloadHelper.java
@@ -100,6 +100,13 @@ public class FileDownloadHelper {
         public String getDownloadDescrition(Context context) {
             return context.getString(R.string.download_file_description);
         }
+
+        public String getMediaUrl(HostInfo hostInfo) {
+            String pathforUrl = fileName.replaceAll(" ", "%20").replaceAll("'","%27");
+            String videoUrl = String.format("http://%s:%d/vfs/%s", hostInfo.getAddress(),
+                    hostInfo.getHttpPort(), pathforUrl);
+            return videoUrl;
+        }
     }
 
     /**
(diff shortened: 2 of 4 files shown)


poisdeux:
Thanks! This will be an awesome addition.

One thing that I am not sure of, is adding an extra button to the interface. I think "local play" is confusing. I'd rather have the download button support this feature by providing a "Stream play" or "Local play" option in the popup dialog. This would make it evident it will not play on Kodi but on the device Kore is running.

yytsui:
Yes, the wold "Streaming" is more proper one to describe this feature.  It's another way of downloading, but will users expect the media file to be kept even after playing is finished if the ui interaction was stated from "Download" button?

Another option is having a new different icon that is obvious cue for streaming, but help from someone good at graphic icon making is needed :)

poisdeux:
I think your right. Using the download button is confusing as well.

How about adding the functionality to the Play button?
Tapping the play button will play the item on Kodi. If the user long presses the play button a context menu pops up giving the user two choices. Something like "Play on Kodi"  and "Stream on device"?

yytsui:
That's a good idea!
Just to clarify, the "Play on Kodi" you mentioned in the menu pops up:
Say I have a Raspi with Kodi installed, and a cellphone with both Kore and Kodi installed. The cellphone::Kore is configured as a remote control of the Raspi::Kodi. So the "Play on Kodi" in the cellphone::Kore menu pops up  means playing video on Raspi::Kodi, not on cellphone::Kodi, right?

SyncedSynapse:
Haven't tested this or looked at it, but thanks for this.

My concern is exactly the UI for this, and i was thinking of using the FAB, with a "speed dial" transition:

https://material.google.com/components/buttons-floating-action-button.html#buttons-floating-action-button-transitions

That shows the 2 options: Play on Kodi and Play on device (or Stream on device. Stream is more accurate, but i think that Play is more user-friendly, so not sure which is better).

One more thing, i think that this should be optional, meaning that there should be an option in settings to disable the "Play on device" option, and just keep the FAB working as currently. I think that the majority of users use the remote to start playing on other devices, so this action should be quick. Those users won't be interested in local play, so they can go into settings and "disable"it. Implementing this is just a matter of checking the setting and showing one FAB or the other.

One more thing to consider is that the current FAB isn't the standard, it moves out of the screen when the user scrolls, and i'm not sure if this FAB supports the speed dial function.

@yytsui Regarding your last comment, yes you're correct.

poisdeux:
The speed dial option for the FAB button looks great. I think that will be best from the UX perspective.
Long-press is not something everyone tries to see if there are more options available.
However, I don't think it is available in the Android SDK. We could use android-arsenal's version:

https://android-arsenal.com/details/1/3062

Regarding the "Play on Kodi" option. A better option would be to mention the name of the mediacenter Kore is currently connected too. So instead of "Play on Kodi" use "Play on OSMC" or "Play on RaspPi" or whatever the user named its Kodi instance when configuring Kore.

yytsui:
So I tried to implement the speed-dial play by using yavski/fab-speed-dial as mentioned by @poisdeux , it's on my "fab" branch 
https://github.com/yytsui/Kore/tree/fab
The video playing works both way, but unfortunately  after user clicking "Play on Kodi" on speed dial menu item, although the video did start playing on Kodi host, the Kore crashed at the point of trying to start CircularReveal animation.

here is the Runtime Error message:
=======================================
/AndroidRuntime: FATAL EXCEPTION: main
                  Process: org.xbmc.kore, PID: 16554
                  java.lang.IllegalStateException: Cannot start this animator on a detached view!
                      at android.view.RenderNode.addAnimator(RenderNode.java:817)
                      at android.view.RenderNodeAnimator.setTarget(RenderNodeAnimator.java:277)
                      at android.view.RenderNodeAnimator.setTarget(RenderNodeAnimator.java:261)
                      at android.animation.RevealAnimator.<init>(RevealAnimator.java:37)
                      at android.view.ViewAnimationUtils.createCircularReveal(ViewAnimationUtils.java:48)
                      at org.xbmc.kore.utils.UIUtils.switchToRemoteWithAnimation(UIUtils.java:425)
                      at org.xbmc.kore.ui.MovieDetailsFragment$2.onSuccess(MovieDetailsFragment.java:330)
                      at org.xbmc.kore.ui.MovieDetailsFragment$2.onSuccess(MovieDetailsFragment.java:316)
                      at org.xbmc.kore.jsonrpc.HostConnection$2.run(HostConnection.java:439)
                      at android.os.Handler.handleCallback(Handler.java:739)
                      at android.os.Handler.dispatchMessage(Handler.java:95)
                      at android.os.Looper.loop(Looper.java:135)
                      at android.app.ActivityThread.main(ActivityThread.java:5221)
                      at java.lang.reflect.Method.invoke(Native Method)
                      at java.lang.reflect.Method.invoke(Method.java:372)
                      at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:899)
                      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:694)
======================================

I will see what I can do to solved this "Cannot start this animator on a detached view" issue.

poisdeux:
Did some quick debugging and it looks like touchGuard is causing problems. You can disable it by adding 

app:touchGuard="false"

to your FabSpeedDial widget in fragment_movie_details.xml

I don't have time right now to do some a thorough analysis. Maybe @yavski can be of more help?

SyncedSynapse:
Not sure about this, but setting `app:touchGuard="false"` doesn't prevent the FAB from disappearing? 

The current behavior - the FAB disappears on scrolling down - should ideally be kept, because the FAB gets in the way when the user is just browsing the movie/tv show. If it's not possible to keep this behavior then the UI needs to be adjusted, at least to increase the bottom margin, so that it's possible to see all the info and the FAB at the same time.

yytsui:
@SyncedSynapse what version of Android were you testing on? Because the "FAB + menu" did stick around on my Nexus 7 Lollipop 5.0 while scrolling?

SyncedSynapse:
@yytsui Sorry, maybe i wasn't clear. I haven't tested this PR yet (no time atm, and i'm having some trouble with my computer...).
What i meant is that the FAB should disappear on scrolling, and i was asking if `app:touchGuard="false"` wouldn't prevent that.

poisdeux:
I believe the current FAB button is from https://github.com/makovkastar/FloatingActionButton
Which is currently deprecated. The official FAB button does not support the attachToScrollView which is currently used to remove the FAB when the user scrolls the grid/listview.
We'll need to add support for this ourselves using a CoordinatorLayout. See https://guides.codepath.com/android/Floating-Action-Buttons for instructions.

yavski:
Hi, I had a look at @yytsui 's fab branch and more particularly at commit `8a20713014445c30e39bcd2a5972e9dea5f11b4f` however I couldn't reproduce the issue. Can you confirm if the issue is on the screen below? If you can provide me with steps to reproduce, I'll look into it and release a fix. Cheers.

device-2016-10-25-213558

poisdeux:
@yavski that is correct. I was able to reproduce the issue on an android virtual device running API 23.
When you select "Play on OSMC" it should crash.

yavski:
@yytsui @poisdeux I've fixed the crash in 1.0.7.

poisdeux:
@yytsui it's been quiet for a while here. I was wondering if you could use any help.

yytsui:
@poisdeux  I did read through https://guides.codepath.com/android/Floating-Action-Buttons and agreed that the behaviours while scrolling should be kept  . Just have to find time to implement it ... I have been quite busy on my daily work  recently.

poisdeux:
@yytsui thx for the update. Looking forward to the final implementation.

primaeval:
Snap. I've just implemented Play Locally too. :)
https://github.com/primaeval/Kore/tree/local
I used a small play icon at the start of the buttons row for Movies and TV Episodes, and a Play Locally menu entry for Addons. You can see if this might work for you. I think it is fairly descrete and intuitive.
I hope this functionality gets into the main app as I tried to make it simple enough for my kids to play movies on their tablets and phones. 
device-2017-02-15-142247
device-2017-02-15-142342

poisdeux:
I still think the FAB speed dial is the best solution to incorporate this feature.

poisdeux:
@primaeval I think you should incorporate @yytsui implementation of the speed dialer and create a new PR. There doesn't seem to be a way for you to continue working on this PR as I've suggested in #351.

poisdeux:
No, your PR of course needs the same solution. If you want to wait on someone else to implement the speed dialer, that's fine. But then you shouldn't bother creating a PR as it probably will get denied.
I'll probably start working on this when I'm done with #320.

primaeval:
I had a think about the buttons and had another thought.
As SyncedSynapse mentioned, the primary function of Kore is as a remote control so I agree that the big blue Play button should really do just mainly or only that on the device it is connected to. That is what users expect.
The Play Local functionality is really only part of all the other things Android can do with a video/media url: play, queue, cast etc. I think it makes more sense to have a seperate Share button instead of a Local Play icon. That would encompass Local Play, queuing, casting and everything else you can do with media urls.
Share Icon:
Share Icon
This is what I get for a movie url:
device-2017-02-15-184541

poisdeux:
Good suggestion. IMO using the share icon is way better then using the play icon in the button bar.
And I like that the share feature includes all other sharing options.
However, I still don't like the extra button and I believe the FAB button functions as the main button for "Play this item". Even if that means playing it on the android device.
So, I'm still in favour of adding the feature to the FAB button.

poisdeux:
@yytsui are you still working on this?

yytsui:
@poisdeux  I will not have to time to implement the UI per our discussion above in the near future :(.
Anyone can help?

poisdeux:
If I recall correctly the touchGuard was the only problem with the current solution, right?
If so, can't we just disable it, so only the FAB button can be used to close the speeddialer?
IMO that will be fine from a UX point a view.

poisdeux:
Closing this because development has been continued in #504

PR #302, opened 2016-10-21
Refactored integration tests to use robolectric

poisdeux:
Refactored our integration tests to use robolectric as much as possible. This allows us to run the tests without using an android device.
We now have three test dirs under **app/src**
- **instrumentationTest**: this holds all tests that should run on an android device and will mostly be 
                                       used to test UI interaction.
- **test**: this holds all tests that can run locally and will mostly be anything that does not test UI 
             interactions.
- **testUtils**: this holds all resources required by either forms of testing. All specific resources should 
                    be included in their respective testing dirs.


SyncedSynapse:
@poisdeux If you do a `build` or `assemble` with this PR, the build fails with proguard errors. Can you please check it out.

poisdeux:
Fixed the proguard errors. Proguard was removing the junit classes...
Also added the assets to the sourceset for `fullDebug` to enable running the local unit tests using the **full** flavor as well. Changed the travis build to use the **full** flavor for local tests as I think **instrumentationTest** might be misleading.

SyncedSynapse:
@poisdeux Can you please check if the tests are ok in this PR. I'm getting a lot of
`org.xbmc.kore.provider.mediaprovider.MediaProviderMusicTest > queryAllArtistsTest FAILED
    java.io.FileNotFoundException`
errors

poisdeux:
The tests are OK. They only work using the `fullDebug` flavor and not `fullRelease`. That is on purpose as I don't want to package the json assets required for testing in the actual release version.

I haven't looked into it but it might be possible to exclude running the tests for the `fullRelease` flavor.
If you really really really want to be able to run the `build` task.

Issue #101, opened 2015-07-20
[Feature Request]Option to sort albums by artist name an year rather than album title


DaveTBlake:
Would be useful to be able to control the sort criteria when listing albums. Currently albums are listed by title, but listing by artist name and year would be more natural - how I would put CDs on the shelf. Best as an option in case some people prefer the current approach.

axelsimon:
Agreed, listing albums by year is what seems most obvious to me, but I can understand if some people prefer in alphabetical order. A way to chose this would be great.

gvkt:
+1 for this suggestion

TheYetiWakes:
+1 from me too.

MartijnKaijser:
@SyncedSynapse @poisdeux this would be great option to have.
Sort by Artist/album  or sort by artist/album(year)    just like in Kodi itself

poisdeux:
See #309. Note that I didn't implement @MartijnKaijser suggestion to sort by Artist/Album. Not sure if that is really useful as you can use the Artist screen to get the albums for a specific artist

MartijnKaijser:
Problem is my library has about 6000 artists and only a handful has albums added.

poisdeux:
So you'll only get the handful listed in the album list. Then sorting on artist should be enough right? Or do these artists have a kazillion albums?

poisdeux:
Should be fixed in the next release. See #309

PR #309, opened 2016-11-04
Added sort options to albumlistfragment

poisdeux:
List is sorted by default on album title. User
can change this to artist or artist/year.

This should fix request #101


DaveTBlake:
Thanks for looking at this feature request.

Issue #30, opened 2015-02-25
[Feature Request] Visual volume indicator


2Zero:
As a user, it would be useful for me to have a visual indicator in Kore of the volume level in Kodi on volume change.

Use case, I often play / queue music using Kore as a control whilst I have the Kodi screen off (sometimes from the next room). An indicator of Kodi volume in Kore would act as visual feedback that changing the volume (via icon or hardware keys) was having an effect.

Personal preference, a slider overlay of some kind which is only displayed for a few seconds on volume change but which could also be directly manipulated by touch.

poisdeux:
Implemented in version 2.2.0.
https://github.com/xbmc/Kore/releases/tag/v2.2.0

PR #312, opened 2016-11-16
Implemented showing volume level on NowPlayingFragment

poisdeux:
This should implement feature request #30

Volume level is shown using a seekbar and a textview. The seekbar can also be used to change the volume. I kept the increment and decrement volume buttons to allow fine grained volume control.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/NowPlayingFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/NowPlayingFragment.java
@@ -50,7 +50,6 @@ import org.xbmc.kore.jsonrpc.method.Application;
 import org.xbmc.kore.jsonrpc.method.GUI;
 import org.xbmc.kore.jsonrpc.method.Input;
 import org.xbmc.kore.jsonrpc.method.Player;
-import org.xbmc.kore.jsonrpc.type.ApplicationType;
 import org.xbmc.kore.jsonrpc.type.GlobalType;
 import org.xbmc.kore.jsonrpc.type.ListType;
 import org.xbmc.kore.jsonrpc.type.PlayerType;
@@ -72,7 +71,8 @@ import butterknife.OnClick;
  */
 public class NowPlayingFragment extends Fragment
         implements HostConnectionObserver.PlayerEventsObserver,
-        GenericSelectDialog.GenericSelectDialogListener {
+                   HostConnectionObserver.ApplicationEventsObserver,
+                   GenericSelectDialog.GenericSelectDialogListener {
     private static final String TAG = LogUtils.makeLogTag(NowPlayingFragment.class);
 
     /**
@@ -132,8 +132,6 @@ public class NowPlayingFragment extends Fragment
     @InjectView(R.id.rewind) ImageButton rewindButton;
     @InjectView(R.id.fast_forward) ImageButton fastForwardButton;
 
-    @InjectView(R.id.volume_down) ImageButton volumeDownButton;
-    @InjectView(R.id.volume_up) ImageButton volumeUpButton;
     @InjectView(R.id.volume_mute) ImageButton volumeMuteButton;
     @InjectView(R.id.shuffle) ImageButton shuffleButton;
     @InjectView(R.id.repeat) ImageButton repeatButton;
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/xbmc/kore/ui/NowPlayingFragment.java
```diff
@@ -471,12 +469,28 @@ public void onError(int errorCode, String description) { }
...
+
+                            @Override
+                            public void onError(int errorCode, String description) {
+                                if (!isAdded()) return;
+                                // Something went wrong
+                                Toast.makeText(getActivity(),
+                                               String.format(getString(R.string.error_getting_properties), description),
+                                               Toast.LENGTH_SHORT).show();
+                            }
+                        }, callbackHandler);
```
SyncedSynapse: Please rebase this PR on the latest version, as the code that is being removed is more recent.

SyncedSynapse:
Well... it looks nice, but you know what i'm going to complain about, don't you :)

There's very little space to include the seek bar... I've tried it on a 4.7" screen and it's difficult and odd to use. It's not very user friendly and i don't have particular large thumbs.

Not sure how this could be improved, but i feel that this isn't the best UI solution. Possible solutions:
1. remove the seek bar and the up/down volume buttons and include just one button that opens a dialog box with the seek bar and maybe up/down;
2. Remove the 2 up/down buttons, move the mute button to the left, move the seek bar to the right of that button, occupying the space freed by the 2 buttons.
other?

I'm also not sure that the volume text on top of the seek bar is necessary. At least on Android 4.3 it looks odd and misaligns the seek bar.

I just had a look at the code and couldn't find anything odd (apart from the comment i left), but there seems to be a bug. When i rotate the device (killing the activity) it always showed "muted". Looks like something on the observer isn't saving the previous state, but i would need to dig deeper to check this out.

poisdeux:
Initially I only used the seekbar to display the volume level. But when testing I noticed that even on lowres screens it was large enough to actually use. But I kinda forgot that I'm not the only one using Kore (I have skinny fingers). 
I don't think we should remove the volume buttons. I personally don't like using the seekbar to set the volume as it does not provide fine grained control. As I see it we have the following options:

1. Use the current proposed setup in the PR and just disable changing the volume with the seekbar
2. Remove the seekbar and only display the volume in percentage between the decrement and 
    increment volume buttons.
3. Keep the current media bar and use a Toast or AlertDialog to show the volume level when it changes


Although I've tested on small lowres devices (virtual devices that is)  I am not sure that option 1 will provide a good volume level indication on all setups. We could try it and just wait until someone starts complaining... 
Option 2 would be a good solution in a functional sense but not very aesthetic.
As the seekbar in the media bar will probably not be a viable solution I prefer option 3.
It is consistent with the main volume control on Android, although that might be confusing for some.

So, thumbs up for option 3?

SyncedSynapse:
I tried removing the up/down buttons to check how it looks and it looks kind of nice. To try it, just try this remote layout: http://pastebin.com/17qfZuHv

https://drive.google.com/file/d/0B8b8P_pnImuPaHpzT0NCOWhlVWM/view?usp=sharing

It is more informative than the up/down buttons, but it can be more cumbersome to use, particularly for fine-grained control. Then again, there's always the option to use the hardware volume keys, which i think is more user friendly anyway.

Regarding you 3 suggestions, 1) and 2) break the UI symmetry, and i'm not seeing how they can be made to look good. 3) is innocuous, but is it that much helpful? Furthermore, it can't show a toast for each volume change, otherwise a lot of toasts would possibly be shown.

poisdeux:
Ah yes, you're right that the up/down volume buttons functionality is duplicated by the hardware volume keys. Although I've noticed that it sometimes skips a level (e.g. going from 49 to 51). I think I'll add a delay to prevent repeating too soon.
I tested your design and I also think it looks good. I even noticed that in landscape mode it's large enough for fine-grained volume control. So, mute button - seekbar combo it is!

poisdeux:
Updated the code. 

Regarding my remark about the volume level sometimes skipping a level. This has nothing to do with keys repeating but Kodi returns the volume as a float instead of an integer. So these are rounding errors. This is something that needs to be fixed on Kodi's side, as the JSON API specifies an integer as return value.

http://kodi.wiki/view/JSON-RPC_API/v6#Application.OnVolumeChanged

PR #315, opened 2016-11-21
change "title" from itemFile.title to itemFile.label

sebastian-de:
change "title" from itemFile.title to itemFile.label for TYPE_ALBUM and TYPE_SONG.

This fixes the View of some Addons for me (for example the Shoutcast2 addon, where only current playing tracks were displayed instead of the Station name).
It also aligns with ListType.Sort.SORT_METHOD_LABEL.


SyncedSynapse:
Honestly i don't know if this will impact something else, but i'll merge it for now. If there are complaints this will have to be revisited.

PR #320, opened 2016-12-02
Implemented showing mute state in action bar

poisdeux:
When Kodi is muted a mute menu item is added to the action bar which can
be used to unmute Kodi. The mute icon is only displayed when Kodi is muted.

I'm not yet convinced this is the best solution. Another solution could be using a snackbar at the bottom to show the mute state. For instance like the example at [material.google.com](https://material.google.com/components/snackbars-toasts.html#snackbars-toasts-specs) shows.
However, I'm afraid some users might find the snackbar solution annoying and don't consider the mute state to be that important.

This should fix issue #199

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/widgets/NowPlayingPanel.java
+++ b/app/src/main/java/org/xbmc/kore/ui/widgets/NowPlayingPanel.java
@@ -1,0 +1,271 @@
+/*
+ * Copyright 2017 Martijn Brekhof. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.xbmc.kore.ui.widgets;
+
+
+import android.content.Context;
+import android.util.AttributeSet;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.ImageButton;
+import android.widget.ImageView;
+import android.widget.TextView;
+
+import com.sothree.slidinguppanel.SlidingUpPanelLayout;
+
+import org.xbmc.kore.R;
+import org.xbmc.kore.jsonrpc.type.GlobalType;
+import org.xbmc.kore.utils.UIUtils;
+
+public class NowPlayingPanel extends SlidingUpPanelLayout {
+
+    public interface OnPanelButtonsClickListener {
+        void onPlayClicked();
+        void onPreviousClicked();
+        void onNextClicked();
+        void onVolumeMuteClicked();
+        void onShuffleClicked();
+        void onRepeatClicked();
+        void onVolumeMutedIndicatorClicked();
+    }
+
+    private OnPanelButtonsClickListener onPanelButtonsClickListener;
+
+    private TextView title;
+    private TextView details;
+    private ImageView poster;
+    private ImageButton previousButton;
+    private ImageButton nextButton;
+    private ImageButton playButton;
+    private MediaProgressIndicator mediaProgressIndicator;
+    private VolumeLevelIndicator volumeLevelIndicator;
+    private HighlightButton volumeMuteButton;
+    private HighlightButton volumeMutedIndicatorButton;
+    private RepeatModeButton repeatModeButton;
+    private HighlightButton shuffleButton;
+
+    public NowPlayingPanel(Context context) {
+        super(context);
+        initializeView(context);
+    }
+    public NowPlayingPanel(Context context, AttributeSet attributeSet) {
+        super(context, attributeSet);
+        initializeView(context);
+    }
+
+    public NowPlayingPanel(Context context, AttributeSet attributeSet, int defStyle) {
+        super(context, attributeSet, defStyle);
+        initializeView(context);
+    }
+
+    private void initializeView(Context context) {
+        LayoutInflater inflater = (LayoutInflater) context.getSystemService(Context.LAYOUT_INFLATER_SERVICE);
+        inflater.inflate(R.layout.now_playing_panel, this);
+    }
+
+    @Override
+    protected void onFinishInflate() {
+        super.onFinishInflate();
+
+        title = (TextView) findViewById(R.id.npp_title);
+        details = (TextView) findViewById(R.id.npp_details);
+        poster = (ImageView) findViewById(R.id.npp_poster);
+        previousButton = (ImageButton) findViewById(R.id.npp_previous);
+        nextButton = (ImageButton) findViewById(R.id.npp_next);
+        playButton = (ImageButton) findViewById(R.id.npp_play);
+        mediaProgressIndicator = (MediaProgressIndicator) findViewById(R.id.npp_progress_indicator);
+        volumeLevelIndicator = (VolumeLevelIndicator) findViewById(R.id.npp_volume_level_indicator);
+        volumeMuteButton = (HighlightButton) findViewById(R.id.npp_volume_mute);
+        repeatModeButton = (RepeatModeButton) findViewById(R.id.npp_repeat);
+        shuffleButton = (HighlightButton) findViewById(R.id.npp_shuffle);
+        volumeMutedIndicatorButton = (HighlightButton) findViewById(R.id.npp_volume_muted_indicator);
+
+        setupButtonClickListeners();
+    }
+
+    public void setOnPanelButtonsClickListener(OnPanelButtonsClickListener listener) {
+        onPanelButtonsClickListener = listener;
+    }
+
+    public void setOnVolumeChangeListener(VolumeLevelIndicator.OnVolumeChangeListener listener) {
+        volumeLevelIndicator.setOnVolumeChangeListener(listener);
+    }
+
+    public void setOnProgressChangeListener(MediaProgressIndicator.OnProgressChangeListener listener) {
+        mediaProgressIndicator.setOnProgressChangeListener(listener);
+    }
+
+    public void setVolume(int volume, boolean muted) {
+        volumeLevelIndicator.setVolume(muted, volume);
+
+        if (muted) {
+            volumeMutedIndicatorButton.setVisibility(View.VISIBLE);
+        } else {
+            volumeMutedIndicatorButton.setVisibility(View.GONE);
+        }
+
+        volumeMutedIndicatorButton.setHighlight(muted);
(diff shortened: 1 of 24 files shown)


app/src/main/res/values/themes.xml
```diff
@@ -120,6 +120,7 @@
         <item name="iconRepeatOne">@drawable/ic_repeat_one_white_24dp</item>
         <item name="iconShuffle">@drawable/ic_shuffle_white_24dp</item>
         <item name="iconVolumeMute">@drawable/ic_volume_off_white_24dp</item>
+        <item name="iconVolumeMutedToolbar">@drawable/ic_volume_off_black_24dp</item>
```
SyncedSynapse: white

poisdeux:
Well I wasn't really sure this is the best solution and I've taken a good look at other players and think adding a playbar at the bottom while something is playing on Kodi is a much better solution. This allows the user to browse the library and be able to control what's playing and change other things like volume. IMO this improves the UX as the current implementation requires switching between library and remote. So yeah, only showing the mute state seems a bit awkward. I just see this as a first step towards including the playbar...

Regarding your remark that I included this in the BaseActivity. I see it's also used by *HostActivity.java, I guess we should separate this and add a new BaseActivity in the host package.

You say you don't like to apply the observer mechanism to the whole app. Are you worried about resources? If so, we could make users disable the feature in the settings screen.

Other stuff remarks:
  1. I wanted it to look different than normal action items. But couldn't find any good colors.
  2. I did this to make sure the mute state is always shown as the leftmost item. I agree this makes 
      adding/changing the menu items shown as actions more difficult.

SyncedSynapse:
Adding a playbar at the botton is probably a good idea. I personally don't feel the need for it (particularly considering that there are always notifications), but i can appreciate it's usefulness.
There are some things that must be discussed about that before commiting to one:
1. Which actions to show. At first sight i would say the same as the ones present in the notification (play/pause/ff/rw), and probably introduce the mute/unmute, but it must be done in a way that looks good.
2. What happens to the "play" buttons on the details screen. They have to migrate upwards, and that might make them a little bit too high...
3. This shouldn't really be in `BaseActivity`. Something like `PlayerControlActivity` or similar would be better (and it should inject the interface on the root view). Ideally this would be done via composition instead of inheritance, but that might be a little clumsy, if we have to "forward" all the lifecycle events to it (haven't thought this through).

Regarding the other questions:
1. Yes, my worry about the observer mechanism being in the whole app is about resources. It's not that i think that it consumes a lot of resources, it doesn't, but i feel that it wasn't justified... A playbar on the other hand justifies it.

poisdeux:
1. I've created a mockup showing a volume, previous, play/pause, and next button. 
   The image at the left (the kodi image is just an example) is used to show the poster of what is 
   currently playing. The volume button will show the mute state and will popup a dialog showing a 
   slide bar representing volume level and allows the user to change the volume. It will disappear 
   after a few seconds (3?).  
image
  
2. I think we can just let it move up. Haven't tested it, but I think there is enough room even on lowres devices.

3. I've been tempted to refactor the org.xbmc.kore.ui package a few times before. I think we should now just go for it and redesign the ui package. And I think we should do this before implementing the playbar. I will create a new PR that will move the activities and fragments into their own packages and we can use that PR to discuss/develop a workable solution.

SyncedSynapse:
Ok, alternatively can you look at, for instance, how Google Play Music does it:

device-2016-12-27-165801


The control bar is simpler, with an image, the currently playing item, and just a Play/pause button. I prefer this version as it looks better to me, and i'm not sure that the fwd/rwd/volume buttons are essential in this context.

It doesn't solve the original issue on this PR (showing the mute status), but that could be a button that appears on the right side of play/pause, only if the mute status is on.

Some other random thoughts:
1. Just to make sure that we're on the same page, this bar would only appear if something is playing, right.
2. The FAB can indeed be pulled up, but only if the bar is showing up, and should return to its normal position if there's nothing playing.
3. Technically how do you see this being implemented? Specifically, the "injection" of the control bar in the view?

poisdeux:
I do think the playbar should include the most used buttons for the given media type currently playing. Especially playing music I find next and prev handy while browsing my music collection. Now I have to switch to the remote and hit back again to continue browsing.

For me the preferred buttons for the different media types would be:

* **Music**: volume, play/pause, next, previous
* **Music video**: volume, play/pause, next, previous
* **TV show**: volume, play/pause
* **Movie**: volume, play/pause
* **PVR**: volume, play/pause

Including the current playing info as the google music player does might not leave enough room to include all my preferred buttons. However, besides changing the buttons depending on what is playing, we could also include the current playing item info depending on the available space. For instance, we could use the following criteria:

* **sw320dp**: Show only the buttons for music. show media info as well for other types
* **sw400dp**: Show poster and buttons for music. Show poster, media info and buttons for other types
* **sw600dp**: Show poster, media info and buttons for all types


Random answers ;):

1. Yup
2. Indeed
3. That's why we need to refactor the ui package first. Using a common activity (PlaybarActivity?) for all activities that need to display the playbar, we can easily add it as hidden in the layout.

SyncedSynapse:
Do you use the volume on Kodi? I tend to leave it as is and just use the tv/home theater remote for sound. Maybe it's just me, but my usage pattern doesn't really rely on Kodi's volume.

On another note, we could brainstorm some other ideas for this, like making it more like the Navigation drawer, including the now playing info, control buttons, and remote buttons:
1. either a new full navigation drawer to the right. This has the advantage that would be technically easy, just another nav drawer, but has the disadvantage that it would be hard to discover by new users.
2.a control bar on the bottom with minimum buttons, that would open to reveal more actions on a swipe up action. This would be more works, as i believe that the existing nav drawer mechanism wouldn't fit here, but would be more discoverable.

Not sure if this was clear.

poisdeux:
I primarily use Kodi to control my AV volume as I do not want to use a separate remote or physically access to the receiver (which is kinda hard as I put it out of view, (yes because of my spouse and kids). So for me, volume control is essential.

I don't like your idea of having another navigation bar on the right. I don't think it is a common UX pattern. I do like the idea of being able to pull up the control bar on the bottom. That would definitely be a nice UX pattern. ( This might eventually even replace the remote activity completely and free up some room in the action bar :) )

Btw, i intended this PR to discuss this. I prefer using PRs for this, as we can more easily discuss implementations as well. Just to be clear, when I finish refactoring the code and we have some agreement on how to implement this, I will update this PR.

SyncedSynapse:
Yeah, a navigation bar on the right isn't common and would be somewhat confusing. I just brought it up because it would be straightforward to implement, but it's not worth it.

A bottom control bar that could be pulled up is nicer but more work. It can be made in stages nevertheless. By the way, i don't think it should replace the remote screen, because it would always miss the "Now playing" fragment, which i find useful. I was thinking that the bar shouldn't even slide up to occupy all the screen, just to the middle or so, and just show the remote buttons, the now playing info and control buttons.

In any case, the question remains: what should be shown on the "collapsed" bar. Looking at your proposal 3 posts up, i think we only diverge on the "volume" button (i agree with showing up fw/rwd if music is playing). What about just showing the volume button if the sound is muted, and touching it would unmute the sound and make it disappear? This would cover the most "urgent" need regarding volume control and in normal conditions there would be space for the usual now playing info.

poisdeux:
I think showing the volume button only when muted is a good solution. We can then show the volume (as we do on the NowPlayingFragment) when the user pulls the control bar up.

poisdeux:
I found this library that implements a sliding panel. https://github.com/umano/AndroidSlidingUpPanel

I've created a simple version using the library. See commit 16f0f300c09b7fe370b1529d4a3c1ea3a0ebe68e

Below image shows the panel collapsed:
image

Here you see it pulled up:
image

I do think showing the mute status is confusing in the above setup. 
I use the volume muted icon for indicating the volume is muted and also for allowing the user to mute the volume. One option would be to use a different color for the indicator. I would prefer RED but that looks horrible on the day theme. Any suggestions are appreciated.

SyncedSynapse:
I think it looks great.
Regarding the mute button, i feel that what we discussed ealier is the best solution - only show the mute button in the collapsed panel if volume is muted, otherwise hide it. As for the color, i would go with the `R.color.accent_default`, just like in the Now Playing fragment (note that under this proposal, the button would always be colored). It's enough to catch the user's attention IMO.

poisdeux:
Yay, 215 days and 17 PRs later, I finally finished the now playing panel.

* The slideup panel is only displayed when something is
   playing. It starts collapsed showing the media poster
   and title of what is currently playing.

* Media controls implemented are volume, progress, shuffle,
         repeat and play/pause for all items. Next and previous are only
         available when a music item is playing.

* In collapsed mode the panel will display the mute button only
         if Kodi is muted. The mute button in expanded mode is always
         visible.

* Panel is enabled by default. Users can disable the panel
         in Settings

* Implemented listening to Player.OnPropertyChanged notifications
         to update shuffle and repeat button states.

SyncedSynapse:
Nice! From a quick test, everything looks to be working. Just think that the collapsed bar size could be bigger. I believe that Google Play's bar is somewhat bigger, but i'll check it.

I'll test it for a few more days, and take a look at the code and merge it afterwards.

Guess it's time for a new release, what do you think?

poisdeux:
Yes, a new release would be nice.

app/src/main/java/org/xbmc/kore/ui/widgets/NowPlayingPanel.java
```diff
@@ -0,0 +1,271 @@
...
+            poster.setLayoutParams(layoutParams);
+        }
+    }
+
+    public ImageView getPoster() {
+        return poster;
+    }
+
+    private void setupButtonClickListeners() {
+        playButton.setOnClickListener(new OnClickListener() {
```
SyncedSynapse: Why not create only one `OnClickListener` and set it as the handle for all these buttons.

SyncedSynapse:
From the tests and looking at the code everything seems good.

I have some doubts about the layout of the notification, namely:
1. In the collapsed state it should be bigger and with more margin between the widgets;
2. The expanded state is somewhat cluttered... Honestly i'm not a fan of having the volume slider there as i believe it isn't that relevant, but i believe that you find it particularly useful, so i'll leave at that. But if including the progress and the volume is relevant, i'd argue that having the full play/stop/ff/rwd, etc is as relevant and the full row of controls should be present (so that the expanded bar mimics the control section on the now playing fragment).

I'll merge this now and will do a few experiments in the next few days to it.

app/src/main/java/org/xbmc/kore/ui/widgets/NowPlayingPanel.java
poisdeux: Didn't really think about it, but you are right. This is creating a fair amount of objects which is not needed.

poisdeux:
Regarding your doubts:

   1. I guess you're right. I tried to keep it small to keep as much room available for the lists. But people with large hands will probably find the buttons too small.
   2. Again you are right that we should actually add more controls. In fact, I want the NowPlayingPanel to include everything from the NowPlayingFragment (replacing it). I think that it will improve UX considerably if we add the NowPlayingFragment and PlaylistFragment to the NowPlayingPanel and have the RemoteActivity only show the RemoteFragment.

PR #328, opened 2016-12-27
Refactor/uipackage

poisdeux:
To make the org.xbmc.kore.ui package more manageable I want to subdivide it into multiple packages.

I came up with the following division:

* **generic**: fragments/activities used by multiple sections
* **sections**: these are the different screens we show (I was also thinking of naming this **screens**)
  * **addon**
  * **audio**
  * **file**
  * **hosts**
  * **remote**
  * **settings**
  * **video**
* **viewgroups**: our custom viewgroups
* **views**: our custom views

All abstract and base classes will remain under **org.xbmc.kore.ui**.

I'm not sure what to do with the OnBackPressedListener. Should this be under generic or move it outside  **org.xbmc.kore.ui** and place it under *utils*?

There are more issues to resolve but I think we first need to come up with a good grouping of all the classes under **org.xbmc.kore.ui**.


SyncedSynapse:
I agree, but i think that it might be cleaner if you would get rid of the *sections* and just keep *addons*, *music*, etc at the base level. 

Regarding *OnBackPressedListener*, why not leave it at the base level, like the abstract classes?

PR #330, opened 2016-12-29
Materialized dialogs and preferences

monzee:
- changed usages of platform `AlertDialog`s to support version to enable material-themed dialogs before Lollipop.
- refactored `Settings{Fragment,Activity}` to use support preference classes
- reorganized and reworded preferences to be more usable in small screens


app/src/main/res/values/strings.xml
```diff
@@ -328,13 +328,13 @@
...
-    <string name="keep_remote_above_lockscreen">Keep remote above lockscreen</string>
-    <string name="pref_keep_screen_on">Keep screen on when using remote</string>
-    <string name="show_notification">Show notification while playing</string>
-    <string name="pause_during_calls">Pause playing while phone in a call</string>
-    <string name="use_hardware_volume_keys">Use volume keys to control volume</string>
-    <string name="vibrate_on_remote">Vibrate on remote button press</string>
+    <string name="switch_to_remote">Show on media start</string>
+    <string name="keep_remote_above_lockscreen">Show over lockscreen</string>
+    <string name="pref_keep_screen_on">Stay awake</string>
+    <string name="show_notification">Notifications</string>
```
SyncedSynapse: I prefer the previous version of this one. 
Note that all preferences now start with a Verb, except this one (or ones that lead to dialogs), and it would be nice if this rule applied to all of them.

app/src/main/res/values/strings.xml
```diff
@@ -328,13 +328,13 @@
...
+    <string name="pause_during_calls">Pause during phone call</string>
+    <string name="use_hardware_volume_keys">Device volume keys</string>
```
SyncedSynapse: Likewise "Use device volume keys" sound better to me...

SyncedSynapse:
Just had some comments regarding the wording on `strings.xml` (which i think is overall better than it was, but i guess it can still be improved).

monzee:
Changed. My main reason for shortening the titles was because they were being ellipsized in portrait mode in my old phone. When I applied the v14.material theme the font became smaller so it's less of an issue now. The original `use_hardware_volume_keys` text was still too long though.

PR #331, opened 2016-12-29
Fix failing Travis-CI builds

monzee:
issuing a pull request to see if the build succeeds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,4 +28,14 @@ android:
 before_script:
 
 script: 
-  - ./gradlew testFullDebugUnitTest assembleFullRelease
+  - ./gradlew assembleFullDebug testFullDebugUnitTest
+
+before_cache:
+  - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
+  - rm -rf $HOME/.gradle/caches/*/plugin-resolution/
+
+cache:
+  directories:
+    - $HOME/.gradle/caches/
+    - $HOME/.gradle/wrapper/
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ android:
 before_script:
 
 script: 
-  - ./gradlew assembleFullDebug testFullDebugUnitTest
+  - ./gradlew --info assembleFullDebug testFullDebugUnitTest
 
 before_cache:
   - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: false
 
 env:
   matrix:
-    - ANDROID_TARGET=android-22  
+    - ANDROID_TARGET=android-25
 
 android:
   components:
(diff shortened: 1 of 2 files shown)


monzee:
It now fails for a different reason. Looks like a test classpath issue.

poisdeux:
Running the **testFullDebugUnitTest** locally from the commandline it seems the fullDebug sourceset does not include the assets anymore... Can't really see what is wrong at the moment.

monzee:
It is odd. Adding `assembleFullDebug` before `testFullDebugUnitTest` seems to fix it here locally, but it's weird that the test task does not depend on the assemble task. Then travis-ci can now find the assets but then it downloads things from sonatype __while__ executing the tests which causes it to fail every test. I have no idea what's going on.

poisdeux:
Moving `assembleFullDebug` before `testFullDebugUnitTest` doesn't fix it for me. The problem is that the `fullDebug` sourceset is not used (anymore). Changing the sourceset to `full` fixes it. However, we use `full` to create release builds and I used `fullDebug` to make sure the testing assets are not included in a release build. To fix this we need to create a new product flavor `localTest` so we can include the testing assets for `localTest` only.

monzee:
The build passed once then failed again. I think Robolectric is doing funky things. It's not really needed though, is it? The CI environment has the actual SDK, the tests could just be run by support AndroidJUnitRunner.

poisdeux:
Disregard my previous comment. Sorry, wasn't running it properly. Apparently the `assembleFulleDebugUnitTest` which is called by `testFullDebugUnitTest` doesn't include the sourceset anymore. 
The AndroidJunitRunner needs to run on an actual android device. So, AFAICT we need robolectric to provide an android environment without needing an actual android device.

poisdeux:
Woohoo... found the culprit. It's the upgrade of gradle 2.1.3 to 2.2.3.

https://github.com/xbmc/Kore/commit/c966e52ee49647a7149fd5252dec3cb8df118852#diff-c197962302397baf3a4cc36463dce5ea

Downgrading to 2.1.3  makes Kore build again. Both locally and on travis.

@SyncedSynapse any reason, other than google recommending it, you upgraded the gradle plugin?

SyncedSynapse:
@poisdeux No, no particular reason. If it's causing trouble let's downgrade it.

SyncedSynapse:
eea3a69d3d97f2a4a6011114022631054d322ead has fixed this, right?

poisdeux:
@SyncedSynapse  Indeed. 

@monzee I was already working on a fix for this on PR #328. When I saw your PR this morning I thought we might get Travis-CI working again together. @SyncedSynapse merged PR #328 just now and it also includes the fix. Making this PR obsolete :(. Sorry about that.
Anyway many thanks for your help. Your remark about running `assembleFullDebug` first, pointed me into the right direction.

monzee:
All good as long as the red X's are gone. Closing this now.

PR #332, opened 2016-12-30
Added detailsfragment mockup

poisdeux:
Next stop. Refactor the detail fragments.

I want to move as much control (logic) as possible to the AbstractDetailsFragment (maybe afterwards refactor that as well into multiple components). This has the benefit that it will make the UI/UX more consistent, allow more easily to change or add new UI elements, reduce bugs, etc. etc..
I've already managed to do this for the list fragments. However, the detail fragments vary a lot more which makes it more difficult.

The general idea is that the concrete detail fragments only provide the information (model) that needs to be displayed. The abstract fragment contains the logic (control, view) to actually display the information.

To be able to move the logic up we need to determine on what in general the detail fragment should show. I've created a mockup showing the common elements in my opinion:

detailsfragment svg

We then use getters that each concrete detail fragment needs to implement. If the getter returns null we can remove the element from the UI if it is not mandatory. This way we can accomodate all detail fragments.
I consider the following elements mandatory: **Art**, **Poster**, **Title**.


SyncedSynapse:
Please be careful with this, and don't go too far with it. Going too far might make it too unflexible to accommodate future needs.
UI code is messy by nature, so i don't try to structure it too much, and i tolerate a (fair) amount of duplication in it. That being said, i haven't though about it in detail.

poisdeux:
My idea is to try to get as much as possible into components (using composition). The abstract class should only mandate the common elements to adhere to an overall UX design and reduce regression. For instance, the shared element transitions really enhances the UX imo but is really hard to get right for each screen separately and keep it that way. Look at the TV show redesign that broke the return transition...

I intend to use a Fragment for the additional info section to allow each screen to add its own elements. Just to be clear, I'll use this PR to refactor the detail fragments.

SyncedSynapse:
I'm getting 2 FCs with this as it stands:
1. In the music section, going into an artist detail;
2. In a tv show, when trying to "abuse" the sync. It's difficult to replicate, but usually triggering a sync on the tv show details, then quickly going into one episode or season, trigger a sync, get back, trigger a sync, etc. gives the following error:
```
    Process: org.xbmc.kore, PID: 30984
    de.greenrobot.event.EventBusException: Subscriber class org.xbmc.kore.ui.generic.RefreshItem already registered to event class org.xbmc.kore.jsonrpc.event.MediaSyncEvent
        at de.greenrobot.event.EventBus.subscribe(EventBus.java:179)
        at de.greenrobot.event.EventBus.register(EventBus.java:165)
        at de.greenrobot.event.EventBus.register(EventBus.java:133)
        at org.xbmc.kore.ui.generic.RefreshItem.startSync(RefreshItem.java:108)
        at org.xbmc.kore.ui.AbstractRefreshDetailsFragment.onRefresh(AbstractRefreshDetailsFragment.java:118)
        at android.support.v4.widget.SwipeRefreshLayout$1.onAnimationEnd(SwipeRefreshLayout.java:188)
        at android.support.v4.widget.CircleImageView.onAnimationEnd(CircleImageView.java:105)

```

Didn't look at this in details, but it doesn't seem to happen in the current version.

poisdeux:
Ok, I'll look into it. 

Currently working hard on a major refactoring on a different branch (https://github.com/poisdeux/Kore/tree/tmp/refactor/detailsfragment). I'll merge it onto this branch when it is in a working state.

poisdeux:
Finally had a good look at the new TV show design. It occurred to me that on a small screen it is not evident that you need to scroll vertically to get to the episodes/seasons. I like the new setup with the progression and next episodes, but I want to reintroduce the tabs at the top to enhance the UX. 
I would like to redesign the TV show details page with two tabs; Info and Progress. Move the next episodes and seasons list to the Progress tab and have the Cast right below the general TV show info. 
In my opinion this makes it coherent with the audio section and improves the UX.
@SyncedSynapse hopefully you agree :).

SyncedSynapse:
@poisdeux Sorry, but i specifically wanted to get rid of the tabs. I think that they waste vertical space and i generally don't like having tabs on a mobile UI. Sure, there are scenarios where they are the best solution, but this isn't one IMO.

I think that having the info, next episodes and seasons in the same screen is the most logical and useful thing to do. Regarding your worry - not being evident that you need to scroll vertically - i don't particularly agree with it, given that it's a pattern used throughout the rest of the app, so the user should be used to it (and if he is not, i believe he will catch it pretty soon). Nevertheless, to address that, notice that the description text has been shortened, to free some vertical space.

Regarding this section, i feel that there might be too much going on in this screen (and too little in others). I'm considering removing the cast info from this screen, and just showing it at the episode level. I'm not sure that this info is particularly important at the tv show level, so it might be sensible to move it.

poisdeux:
I've finally have a setup we can further discuss. Pinging @monzee as he/she might have some fresh ideas regarding design.

What I actually have done is apply the MVC design pattern to details fragments. 
The general idea is that the abstract classes are responsible for showing and updating the information and the concrete classes just tell what to show/update.
To clarify the setup I've added UML class and sequence diagrams (using [PlantUML](http://plantuml.com/)) which have been added under [doc/diagrams](https://github.com/poisdeux/Kore/tree/fe98fb0536186b7dafbbe5220ebaf78896f5c2f8/doc/diagrams).

I've created a few new abstract classes in the [ui](https://github.com/poisdeux/Kore/tree/fe98fb0536186b7dafbbe5220ebaf78896f5c2f8/app/src/main/java/org/xbmc/kore/ui) section
   * `AbstractFragment` class to hold the required information using the `DataHolder`
   * `AbstractInfoFragment` class to display media information
   * `AbstractAddtionalInfoFragment` class to allow `*InfoFragments` to add additional UI elements and propagate refresh requests. See for an example [`TVShowInfoFragment`](https://github.com/poisdeux/Kore/blob/fe98fb0536186b7dafbbe5220ebaf78896f5c2f8/app/src/main/java/org/xbmc/kore/ui/sections/video/TVShowInfoFragment.java) which adds [`TVShowProgressFragment`](https://github.com/poisdeux/Kore/blob/fe98fb0536186b7dafbbe5220ebaf78896f5c2f8/app/src/main/java/org/xbmc/kore/ui/sections/video/TVShowProgressFragment.java) to display next episodes and season progression.

I've also moved a lot of code into its own class to improve reusing code and encapsulate complexity
   * `RefreshItem` class to encapsulate all refresh functionality from AbstractDetailsFragment
   * `SharedElementTransition` utility class to encapsulate all shared element transition code
   * `CastFragment` class to encapsulate all code for displaying casts reducing code duplication
   * `DataHolder` class so we don't have to pass the ViewHolder from the `*ListFragment` to the `*DetailsFragment` or `*InfoFragment`

Now Kore uses `*DetailsFragments` for both fragments that show generic
info about some media item and fragments that hold all details for some media item.
For example, artist details shows artist info and used tabs to show artist albums and songs as well. 
IMO this is confusing and I've refactored AbstractDetailsFragment into two classes:
   * `AbstractDetailsFragment`: for fragments requiring a tab bar
   * `AbstractInfoFragment`:      for fragments showing media information

SyncedSynapse:
Sorry for the delay... And i still haven't got time to fully look into this.

But anyway, from a quick look it looks good, but i've found some issues that might be relevant:
1. The cast label disappeared. Was this on purpose?
2. Refresh isn't implemented in all items, so the message `Refreshing not implemented for this item` shows up. Additionally, when this happens the refresh animation isn't being stopped. This occured to me in the Music section (artist and album) and the addon details section. I think that, if a refresh is possible, something should happen. In the addon section i think it should just fetch the details again. In the music section, there's no Sync single artist/album, so it's a bummer that a refresh triggers a sync of all the music database, but i think this is better than nothing happening.
3. I'm getting a FC when triggering a refresh on a TV show episode details fragment:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.setVisibility(int)' on a null object reference
                                                                 at org.xbmc.kore.ui.sections.video.TVShowProgressFragment.displaySeasonList(TVShowProgressFragment.java:264)
                                                                 at org.xbmc.kore.ui.sections.video.TVShowProgressFragment.onLoadFinished(TVShowProgressFragment.java:162)
                                                                 at org.xbmc.kore.ui.sections.video.TVShowProgressFragment.onLoadFinished(TVShowProgressFragment.java:51)
   ```

Haven't looked into it, but looks like `TVShowProgressFragment` is still around, so its refresh methods get called.

Finally, the UML diagrams are a nice touch and help a lot, but it would also be nice if there were some comments at the class level (on the `abstract` fragments), just so that someone that looks there gets part of the picture. This is just a suggestion obviously.

SyncedSynapse:
I think there's a more profound issue with this than the NPE that you fixed.

Consider the flow of the tv shows activity: 
- It creates the list of tv shows;
- Selecting one the `TvShowInfoFragment` is created, which inherits from `AbstractInfoFragment` and creates a `AdditionalInfoFragment` (which gets added to the `FragmentManager` with a `ADDITIONAL_FRAGMENT_TAG` tag. Note that this fragment is never removed from the fragment manager, ever);
- Selecting a episode creates a `EpisodeDetailsFragment`, which nowadays doesn't create another `AdditionalInfoFragment` but it could and if it did, it would be added to the same `FragmentManager` with the same tag. And the `Refresh` method would behave strangely (it would get only one of the fragments and refresh it).

See what i mean?

poisdeux:
Regarding your concern about searching the nested fragments, I don't think this is a problem as the fragment manager first searches through the current layout. This means it will always return the one active and not any that are also available on the backstack.
However, I did change the way I handle the additionalInfoFragment. I'm now using the child's fragment manager which enables reusing existing additionalInfoFragments when returning to a fragment from the backstack, instead of always creating a new one.

SyncedSynapse:
Using the child fragment manager looks better. Now the lifecycle events of the added fragments (`onPause` and `onStop`) are correctly being called when the "parent" fragment is replaced on the activity.

Is there something more you want to add, or can i merge this?

poisdeux:
I want to test this a bit more and create an overview of which screens deviate from the current setup.

poisdeux:
AFAICT I squashed all issues. Note that I deviated from the current layout in the following screens:

  * Movie info: directors moved to right details section
  * TVShow episode info: directors moved to right details section
  * Addon info: moved authors and version to right details section

The PVR section still needs to be refactored. But I want to use a separate PR for that as I need to write some tests first. The reason is that I do not have a PVR setup and I'll need to use the PVR json responses under doc/json_responses to test the refactoring.

SyncedSynapse:
Regarding the layout changes, i hadn't noticed them, so no big deal (though now that you pointed them out, i might take a stab at making it look somewhat more balanced:)).

I'll take care of the PVR section, next week or so. But FYI take a look at the "Hardware request thread" on the Internal forums and/or talk to @MartijnKaijser about it.

Ok to merge for you?

poisdeux:
Yes, this can be merge.

Regarding the hardware, I don't have cable TV. I only watch movies and TV programs through internet services. Nowadays these services are pretty decent.

PR #335, opened 2017-01-03
Fixed issue with sharing code between local and instrumentation tests

poisdeux:
Using the **fullDebug** build variant, building from Android failed. I didn't notice this before as I was constantly using the **instrumentationDebug** build variant.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -40,14 +40,16 @@ android {
 
     sourceSets {
         fullDebug {
-            assets.srcDirs = ['src/testUtils/assets']
+            assets.srcDirs += 'src/testUtils/assets'
         }
         test {
-            java.srcDirs = ['src/test/java', 'src/testUtils/java']
+            java.srcDirs += 'src/testUtils/java'
+        }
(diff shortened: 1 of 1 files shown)


SyncedSynapse:
This breaks the `assembleRelease` target, which is used by jenkins.

app/build.gradle
```diff
@@ -40,14 +40,16 @@ android {
...
         }
         test {
-            java.srcDirs = ['src/test/java', 'src/testUtils/java']
+            java.srcDirs += 'src/testUtils/java'
+        }
+        androidTest {
+            java.srcDirs += 'src/testUtils/java'
         }
         instrumentationTest {
-            java.srcDirs = ['src/testUtils/java']
```
SyncedSynapse: This probably

poisdeux:
Hmm, strange. I can execute `assembleRelease` just fine. Where does it fail?

SyncedSynapse:
I swear i did get an error, but it's building fine now, after a `clean`... Sorry.

Issue #279, opened 2016-07-30
Sort TV shows by "last played" Status 


andyx1975:
Hi, 

One Feature that i miss in kore is to sort TV shows by last played date. This makes really for users... Because you can have every show that you are Currently watching on top. 

Thanks 
Andy

poisdeux:
Closing as this was fixed in #340

PR #340, opened 2017-01-20
Added option to sort by last played

tomerf:
Added option to sort movies and tv shows by last watched date, fixing issue #279


SyncedSynapse:
A previous PR changed `MediaDatabase.java`, so this needs to take that into account (just change the DB version code).
Can you please check it.

PR #343, opened 2017-01-28
Add support for sharing SVTplay urls to "Play on Kodi".

yohell:
SVTplay is a free streaming service provided by the Swedish
national television SVT.

This feature requires the svtplay video plugin on Kodi.


SyncedSynapse:
I've merged this without testing it (don't know SVT), but it has occurred to me that you didn't declare the intent filter on `manifest.xml`. Shouldn't these URLs be defined there?
I suppose that they are being caught by the generic intent filter, but it would be nice if they were defined explicitly.

yohell:
I expect you are completely right; I flat out forgot :-)

Fix submitted in new pull request.

PR #344, opened 2017-01-29
Refactored handling of shared urls, added handler for twitch.tv streams.

monzee:
I moved all intent-handling code out of the activity into a fragment. It's pretty much a rewrite, need testing to see if the behavior was preserved. I have not tested Vimeo shares. YouTube shares from the app seem ok. Also added handler for Twitch live streams from the app.


app/src/main/java/org/xbmc/kore/ui/sections/remote/ShareHandlingFragment.java
```diff
@@ -0,0 +1,370 @@
...
+            @Override
+            public void got(Boolean isPlaying) {
+                if (isPlaying) {
+                    new Seq<>(enqueue(h, file))
+                            .then(hostNotify(h, getString(R.string.shared_video_added)))
+                            .start(refreshPlaylist());
+                } else {
+                    new Seq<>(clearPlaylist(h))
+                            .then(enqueue(h, file))
+                            .then(play(h))
```
poisdeux: If I understand the Seq correctly, `play` will be started even if `enqueue` fails. This is different from the original chaining setup. I haven't tested this so I'm not sure, but this might result in some unexpected situations for the user.

app/src/main/java/org/xbmc/kore/ui/sections/remote/ShareHandlingFragment.java
```diff
@@ -0,0 +1,370 @@
...
+
+    interface OnFinish<T> {
+        void got(T result);
+    }
+
+    interface Task<T> {
+        void start(@NonNull OnFinish<? super T> then);
+    }
+
+    static class Seq<T> implements Task<T> {
```
poisdeux: This should probably be moved into its own Class under org.xbmc.kore.utils as I think it will be applicable to other parts of Kore as well. Also `Seq` is a bit cryptic, how about naming this `Sequencer`?

app/src/main/java/org/xbmc/kore/ui/sections/remote/ShareHandlingFragment.java
```diff
@@ -0,0 +1,370 @@
...
+            this.task = task;
+        }
+
+        <U> Seq<U> then(final Task<U> block) {
+            return new Seq<>(new Task<U>() {
+                @Override
+                public void start(@NonNull final OnFinish<? super U> then) {
+                    task.start(new OnFinish<T>() {
+                        @Override
+                        public void got(T unused) {
```
poisdeux: Can't we add a boolean argument here to make it possible to start the next task only if the previous succeeded?

app/src/main/java/org/xbmc/kore/ui/sections/remote/ShareHandlingFragment.java
monzee: How about `Task`, `Task.OnFinish` and `Task.Sequence`? It kinda feels bad to write single-method interfaces in their own file.

app/src/main/java/org/xbmc/kore/ui/sections/remote/ShareHandlingFragment.java
monzee: That's how it works currently. The downstream tasks aren't executed when a task doesn't invoke its `OnFinish`. I'll admit the code is really confusing as short as it is, but in my log tests before I removed them, it appears correct. I'll try to write tests to confirm.

app/src/main/java/org/xbmc/kore/ui/sections/remote/ShareHandlingFragment.java
monzee: If the error method of the API callback gets called, the task's `OnFinish` doesn't get called and the chain stops there

monzee:
Thanks for the feedback! I'll get back to this in a couple of days, got something else on the plate.

SyncedSynapse:
Right on time! I was actually thinking that a refactor was needed on that part of the code. What started as a quick and dirty way to share youtube urls has become a lot of code. Notice that i just merged a PR that also handles some other urls, so i would appreciate if you rebase this against the current master.

I agree with the feedback that @poisdeux posted, and i have some more comments:

1. Does this need to be a fragment? I mean, a fragment probably has some overhead that isn't necessary here. Doing this in a fragment also "obscures" the intent filter handling - in the manifest the intent filter is defined as being handled on the `RemoteActivity`, and now it isn't clear where it is being handled. I would prefer to keep `handleStartIntent` on the activity, and just refactor the code to somewhere else. But i don't think a full fledged fragment is needed.

(to be continued)

app/src/main/java/org/xbmc/kore/ui/sections/remote/ShareHandlingFragment.java
poisdeux: Ah yeah, Task is the main part here.

app/src/main/java/org/xbmc/kore/ui/sections/remote/ShareHandlingFragment.java
poisdeux: Thanks for explaining. However, I'm still having a hard time manually reproducing the flow.
As I understand it now this pattern is a combination of the builder and chain pattern?

SyncedSynapse: I agree with @poisdeux  here. I kind of like it, but i can't decide if it's elegant/useful or not because it is not simple.
At first sight there's too much going on on that pattern/idiom, and everything would be acceptable if it would lead to a usage that was obvious and simple, but looking at the previous 10 lines, i don't think it is particularly clear...
Maybe i need to digest this more.

monzee: This is an incomplete form of the idiom from functional languages called the continuation monad. It's missing the type for `Task<U> from(T value)` and the corresponding handler for it in `Seq` because I have no use for it here. I think the usage is simple, just the implementation is tricky. If it has docs and tests it should be more usable.

monzee:
Yes, I do think this should be in a fragment. IMO any lifecycle-sensitve code should be in activities or fragments. This sharing business is different enough from everything else happening in `RemoteActivity` that it should be pulled out from it. If it's not in a fragment, it would lead to something like described [in this talk](https://youtu.be/k3IT-IJ0J98?t=1487) and the `PlaylistFragment` interaction would be tricky.

monzee:
@yohell I'd appreciate it if you can check if the svtplay.se videos are handled correctly. I don't know where to get the plugin and the links to their videos.

yohell:
Sure, however I'm presently on the move and cannot readily check.

If I recall correctly I did nothing special to install it, but just got openelec and then got the svtplay plugin from the kodi menus, from under "video plugins". If I'm not completely misremembering, I believe kodi might at some point have even asked me if I'd like to install plugins such as svtplay on request?

Here's an url that should work if you want to try it yourself:

http://www.svtplay.se/video/6707361/bo-kaspers-vi-kommer-aldrig-att-do/bo-kaspers-vi-kommer-aldrig-att-do-avsnitt-1

Visit it in your web browser, then share it to "Play on Kodi", have some Swedish fine culture :)

Cheers!
/Joel



⁣Sent from TypeApp ​

yohell:
Ok, now I managed to test it with a few svtplay urls. Seems to work just fine!

monzee:
Wow, I didn't know it does that. I only had to click on the link in Gmail and it prompted to download the plugin. Pretty awesome.

app/src/main/java/org/xbmc/kore/utils/Task.java
```diff
@@ -0,0 +1,81 @@
...
+         * @param <U> The type of the value generated by the next task
+         * @return a sequence
+         */
+        public <U> Sequence<U> then(final Task<U> next) {
+            return new Sequence<>(new Task<U>() {
+                @Override
+                public void start(@NonNull final OnFinish<? super U> then) {
+                    task.start(new OnFinish<T>() {
+                        @Override
+                        public void got(T unused) {
```
poisdeux: As I understand it, within a Monad the bind operator ensures the next operation is not performed when the previous failed. Therefore a programmer using a Monad does not need to worry about the logic and can simply assume if his operation fails the next in line won't be called. This is not the same as how it is implemented here. Here, the programmer has to know that he/she should call got(...) to make sure the next in line is executed or should NOT call got(...) to stop the chain. 
So we need something to indicate failure/success and I think the easiest way to do this, is by changing the `got(...)` method to `got(T result, boolean status)` and use the status to determine if `then.start(then)` should be called.

app/src/main/java/org/xbmc/kore/utils/Task.java
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * A sequence of tasks chained together.
+     *
+     * The values generated by the individual tasks are ignored except for the
+     * last one which may be accessed in the consumer if all of the tasks in
+     * the chain completed successfully.
+     *
+     * @param <T> The type of the value generated by the last task.
+     */
+    class Sequence<T> implements Task<T> {
```
poisdeux: Shouldn't Task be decoupled from Sequence? Now I always get Sequence when I create a Task right? I would think that creating a single Task without running it in a Sequence should also be possible.

app/src/test/java/org/xbmc/kore/utils/TaskSequenceTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Test
+    public void tasks_are_called_in_order() {
+        final StringBuilder sb = new StringBuilder();
+        final AtomicBoolean done = new AtomicBoolean(false);
+        new Task.Sequence<>(new Task<Object>() {
+            @Override
+            public void start(OnFinish<? super Object> then) {
+                sb.append(1);
+                then.got(null);
```
poisdeux: Here you clearly demonstrate a nice usage of passing data from one task to another. So why not change the Sequence and Task and have it pass the argument to the next task instead of throwing it away?

app/src/test/java/org/xbmc/kore/ui/sections/remote/SharedUrlConverterTest.java
```diff
@@ -0,0 +1,197 @@
...
+    private static final String TWITCH_ID = "arteezy";
+    private static final String GARBAGE_BEFORE = "asdf sdf sdf ";
+    private static final String GARBAGE_AFTER = " asdf sdf sdf";
+
+    private String twitchUrl(String id) {
+        return "plugin://plugin.video.twitch/playLive/" + id + "/";
+    }
+
+    @Test
+    public void twitch_actual_shared_text() {
```
poisdeux: Any particular reason for using underscores instead of CamelCase?

poisdeux:
In general I like the Sequence pattern (based on the continuation monad) and in my opinion it improves the readability of the current chaining setup. 
I do think that the way it is currently setup is hard to understand for people assuming a OO design. At least that's what got me confused. So hopefully we can come to some kind of setup that is more intuitive to use for programmers with no functional programming experience. Please see my inline comments for further discussion.

app/src/main/java/org/xbmc/kore/utils/Task.java
```diff
@@ -0,0 +1,81 @@
+package org.xbmc.kore.utils;
+
+import android.support.annotation.NonNull;
+
+/**
+ * Represents a computation that will eventually complete.
```
poisdeux: What's a computation? What does eventually complete mean? (Try to place yourself inside a Java only programmer)
I would just stick to the term Task as something that needs to be done.

app/src/main/java/org/xbmc/kore/utils/Task.java
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * @param <T> The type of the value being computed or awaited.
+ */
+public interface Task<T> {
+
+    /**
+     * Starts the computation, invokes a procedure when done.
```
poisdeux: which procedure will be invoked? From this comment I'm assuming something other then the argument will be invoked as well.

app/src/main/java/org/xbmc/kore/utils/Task.java
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @param then The consumer of the value being computed.
+     */
+    void start(@NonNull OnFinish<? super T> then);
+
+    /**
+     * A procedure that awaits a value.
+     *
+     * @param <T> The type of the value generated by a previous computation.
```
poisdeux: Shouldn't this be the same type of the Task that implements the `OnFinish`? If so, the comment can be removed or should make that clear.

app/src/main/java/org/xbmc/kore/utils/Task.java
```diff
@@ -0,0 +1,81 @@
...
+ * Represents a computation that will eventually complete.
+ *
+ * @param <T> The type of the value being computed or awaited.
+ */
+public interface Task<T> {
+
+    /**
+     * Starts the computation, invokes a procedure when done.
+     *
+     * @param then The consumer of the value being computed.
```
poisdeux: In the Sequence the value being computed is not passed to the consumer, right? If so, this seems to contradict its main usage.

SyncedSynapse:
Going back to the fragment issue:
The justification you present is applicable to components that 1) are intended to be reusable and 2) are wired up to an activity lifecyle. Neither is applicable here: 1) This isn't going to be reused somewhere else (nevertheless, there's no harm in designing it as if it would), and 2) It really doesn't need to be wired up to the activity lifecycle, except in the sense that it needs to have access to the intent that started the activity, but there's no "management" to be done afterwards.

And like i said previously, it has a severe drawback: it obfuscates the place where the intent is handled. In the manifest it is explicitly stated that those intents are handled on the `RemoteActivity`, and now they aren't... they are handled on a fragment that gets to be instantiated in the activity, which isn't at all obvious for someone looking at it for the first time.

IMO either the intent gets handled at the `RemoteActivity` (and if it is too much code it can be partitioned to somewhere else) or a new activity that is explicitly specified in the manifest, and that handles control to the `RemoteActivity` after handling the intent.

Not really relevant given the above, but the `PlaylistFragment` registering with the intent handling fragment so that it can be refreshed isn't a good pattern. The fragments should be independent and all communication/synchronization should be done via the activity. Better yet, the `EventBus` should be used to pass appropriate messages (this would also help in synchronizing if the intent is handled in a different activity than the Remote activity).

Finally, this PR is including some duplicate commits that are already on master, so it isn't applying cleanly. @poisdeux can you help out here.

poisdeux:
@monzee If you run `git log --graph --decorate --oneline` on this branch (twitch-share) you see there was another twitch-share branch merged that was probably not up-to-date. Maybe the result of working on different machines?

```
* 135ff67 (HEAD -> monzee-twitch-share) Use RobolectricTestRunner on shared url tests
*   721c518 Merge branch 'twitch-share' of github.com:monzee/Kore into twitch-share
|\  
| * 96fe27a Now shows a notification when queuing a shared url while playing
| * 76ae02c Removed unused imports in RemoteActivity
| * 816131b Refactored handling of shared urls, added handler for twitch.tv streams.
* | 5145165 Added svtplay; merged recognizers into 1 static method, fixed tests
* | b64baf2 renamed a few things; preparing for rebase
* | 5c25fdb Extracted Task and related types into utils package; added Seq tests
* | 44ee0a9 Now shows a notification when queuing a shared url while playing
* | ca1fe57 Removed unused imports in RemoteActivity
* | ca99ebb Refactored handling of shared urls, added handler for twitch.tv streams.
* | 13abb4e (tag: v2.2.0, upstream/master, origin/master, origin/HEAD, master) Bump version for release Update release notes
* | 4c82094 Add www.svtplay.se host to "Play on Kodi" intent filter. (#345)
* | e72b7f4 Update Portuguese translation
* | 7a2ed77 Add support for sharing SVTplay urls to "Play on Kodi". (#343)
* | d5f8ee4 Added option to sort by last played (#340)
|/  
* c914723 Fix for empty titles on playlist
```

You can fix this by rebasing on `5145165` while working on branch `twitch-share` that is pointing to `135ff67`:

```
git rebase 5145165
```

Then you'll probably need to fix some conflicts...

The end result should be:

```
* 9e68b17 (HEAD -> monzee-twitch-share) Fixed NPE when calling intent#getAction() after changing themes
* e6ac8d2 Use RobolectricTestRunner on shared url tests
* bb649d1 Now shows a notification when queuing a shared url while playing
* cc7a10d Removed unused imports in RemoteActivity
* b6773fe Refactored handling of shared urls, added handler for twitch.tv streams.
* 5145165 Added svtplay; merged recognizers into 1 static method, fixed tests
* b64baf2 renamed a few things; preparing for rebase
* 5c25fdb Extracted Task and related types into utils package; added Seq tests
* 44ee0a9 Now shows a notification when queuing a shared url while playing
* ca1fe57 Removed unused imports in RemoteActivity
* ca99ebb Refactored handling of shared urls, added handler for twitch.tv streams.
* 13abb4e (tag: v2.2.0, upstream/master, origin/master, origin/HEAD, master) Bump version for release Update release notes
```

While you're at it, it wouldn't hurt to squash the commit history to a single commit...

app/src/main/java/org/xbmc/kore/utils/Task.java
monzee: I think that should be encoded in the `T` instead, but this implementation is incomplete like I said before. It doesn't have a type for `T -> Task<U>`. I'm hesitant to add it because 1) it's painful to read and use without retrolambda and 2) it's probably better to use a real library for this like RxJava or bolts-tasks than reimplement them.

app/src/test/java/org/xbmc/kore/utils/TaskSequenceTest.java
monzee: Yeah, it's incomplete. Will something like that be useful elsewhere in the code? I find it unpleasant to use without lambdas. I actually have a branch with something like that and I'm looking for places where it could be used.

app/src/test/java/org/xbmc/kore/ui/sections/remote/SharedUrlConverterTest.java
monzee: Just a habit. I don't see tests as methods, they're more like specs or documentation to me so they tend to have long names. SnakeCase gets really weird to read when there's a dozen words especially when there's the article 'a' somewhere.

app/src/main/java/org/xbmc/kore/utils/Task.java
monzee: I only implemented the kind of bind which drops the result, but in the last step the result does get passed. It's also the only place where you write a consumer (`OnFinish#got`) in a sequence, so it still makes sense.

app/src/main/java/org/xbmc/kore/utils/Task.java
monzee: Yes, same `T` as the `Task` that takes the `OnFinish`

monzee:
Sorry about that, not sure what happened when I rebased. Thanks for the instructions,  I have done them and the history looks straight now. How do I squash?

app/src/main/java/org/xbmc/kore/utils/Task.java
monzee: I wouldn't say it's coupled to Sequence. Ideally you shouldn't care about whether something is a Sequence or a Task. A Sequence is also a Task. You can pass it to `#then(Task)` or even do `new Sequence<>(sequence)`. It is possible to run a Task that's not in a sequence. I do exactly that in `#isPlayingVideo(handler).start(...)` in the fragment code. You can also write a named class that implements Task and use an instance of that in a Sequence.

monzee:
@SyncedSynapse I don't know, I feel pretty much the opposite. To me, it's way clearer to get rid of integration boilerplate in the activity and just attach a fragment. The name of the fragment says what it does so I don't think it's obfuscated.

Agreed the refreshing logic feels dirty. I tried to stick close to the original code without adding extra dependencies. The original refresh code is way worse IMO. One should never rely on private framework code like that. I actually didn't notice there's already a dependency on EventBus. I'll see if I can use that instead.

poisdeux:
@monzee regarding the squashing see https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

SyncedSynapse:
No need to squash, I can do it when it is merged in GitHub.

monzee:
Sorry for leaving this hanging, it doesn't really work anymore since the Twitch plugin changed some months ago.

PR #355, opened 2017-02-18
Make remote bottom bar configurable

tomerf:
Allow the user to configure what shortcuts to put in the bottom bar of the remote screen.
Also fixed the music link (broken in v17?)

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteFragment.java
@@ -25,6 +25,7 @@ import android.graphics.PorterDuffColorFilter;
 import android.graphics.drawable.BitmapDrawable;
 import android.os.Bundle;
 import android.os.Handler;
+import android.preference.PreferenceManager;
 import android.support.v4.app.Fragment;
 import android.view.LayoutInflater;
 import android.view.MotionEvent;
@@ -39,6 +40,7 @@ import android.widget.RelativeLayout;
 import android.widget.TextView;
 
 import org.xbmc.kore.R;
+import org.xbmc.kore.Settings;
 import org.xbmc.kore.eventclient.ButtonCodes;
 import org.xbmc.kore.eventclient.EventServerConnection;
 import org.xbmc.kore.eventclient.Packet;
@@ -59,9 +61,14 @@ import org.xbmc.kore.utils.RepeatListener;
 import org.xbmc.kore.utils.UIUtils;
 import org.xbmc.kore.utils.Utils;
 
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
 import butterknife.ButterKnife;
 import butterknife.InjectView;
 import butterknife.OnClick;
+import butterknife.Optional;
 
 /**
  * Remote view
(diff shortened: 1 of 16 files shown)


app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteFragment.java
```diff
@@ -227,6 +240,19 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
         skipPreviousIcon = styledAttributes.getResourceId(styledAttributes.getIndex(3), R.drawable.ic_skip_previous_white_24dp);
         styledAttributes.recycle();
 
+        Set<String> shownItems = PreferenceManager
+                .getDefaultSharedPreferences(getActivity())
+                .getStringSet(Settings.getRemoteBarItemsPrefKey(hostInfo.getId()),
+                        new HashSet<>(Arrays.asList(getResources().getStringArray(R.array.entry_values_remote_bar_items))));
```
SyncedSynapse: `default_values_remote_bar_items` as the default

SyncedSynapse:
Thanks for this, but it has some issues that should be addressed:

1. See my line comment. Without it, when the user first installs the version that has this enabled, it will show up all 11 icons, which is not what we want;
2. It is not saving the selection in the settings. This is because the setting has the `host_id` appended as its name, but when you save it, you're saving the setting without the `host_id` appended. Look into https://github.com/tomerf/Kore/blob/f826f7fac042c3795fe0f195d924e1a9b34a54ef/app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java#L59 to see how that is handled in the Navigation Drawer items. A similar solution should be applied to this setting;
3. I'm not sure about all the items that you included in the bar, namely the Videos, Weather and Favorites options.
3.1 Videos is kind of redundant IMO, given that Movies and TV Shows is already there. I might be wrong, but personally i don't see the need for it. In any case, at a minimum, the icon should be changed to distinguish it from the Movies one;
3.2 IMO the Weather is dispensable. Maybe it's just me, after all it is present by default on Kodi...
3.3 Favorites is one of the most requested features for Kore, but there's a big issue with it, and that's why it hasn't been included. Opening the Favorites window via JSON RPC seems to freeze Kodi's response to JSON RPC commands (both through TCP and through HTTP), so the user can get stuck easily. EventServer commands still work, but you can't dismiss the window, because "Back" is sent through JSON RPC. This doesn't happen when the Favorites window is opened through Kodi though - json rpc is responsive. @Montellese Is this related to the issue of modal dialogs hanging up the TCP listener thread (though on this case, it only happens if the favorites window is opened through json rpc, and it also blocks json rpc commands sent through HTTP)?

tomerf:
Thanks for the input. I noticed some things could use more refining.
1. Will fix.
2. Thanks, I'm sure I did that, weird.
3. I just copied the side bar items from the Estuary. Videos depends if you count files view. Icon idea? Favorites makes a lot of sense, but that Kodi bug is a problem (I used my TV remote to get out). Still, it's a Kodi issue, not Kore.

SyncedSynapse:
As for the videos icon, i would suggest either (from [here](https://material.io/icons/)) the "Local movies" from the "Maps" section, the "Video label" from the "AV" section or the "ondemand video" from the "Notification" section. These should be available on Android studio.
I'm also not sure about the "Weather" icon, given that a cloud is currently strongly associated with computing... Maybe the "wb sunnny" would be better (not sure if it looks good though).

As for the favorites, it doesn't matter if it's a Kodi or a Kore issue. It can leave the user in a unrecoverable situation, and that shouldn't be done.

tomerf:
All issues have been fixed and controversial items have been commented out.
I will check the icons tomorrow.

SyncedSynapse:
It's saving the preferences now, but when you return to the settings page, the saved options aren't selected. It always revert to the defaults.
This is because of what is explained here: https://github.com/tomerf/Kore/blob/f826f7fac042c3795fe0f195d924e1a9b34a54ef/app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java#L65
Take a look at that comment and the code that follows it.

One more thing, i think that this setting should be in the `Remote` section, not on the `Application`, and then you can drop the `Remote` from its name (just `Bottom bar shortcuts`).

PR #361, opened 2017-02-22
Move hard-coded strings (group titles Remote and Application) to strings.xml

akc-kyiv:


SyncedSynapse:
Thanks, not sure how i missed those...

PR #372, opened 2017-03-14
Fixes NPE in ConnectionObserversManagerService.onDestroy

poisdeux:


SyncedSynapse:
How about not letting the collection be null, instead.

Moving line https://github.com/poisdeux/Kore/blob/f0aef75b9ac0eb619ab50d87c124e856c7aca11e/app/src/main/java/org/xbmc/kore/service/ConnectionObserversManagerService.java#L97
to the declaration of `mConnectionObservers` should also do the trick (hopefully).

(also, the same check could be removed in `onTaskRemoved`)

poisdeux:
I guess so. Code updated.
However, AFAICT the NPE should never occur according to the Service lifecycle. The only situation where onStartCommand is not executed is when binding to the service. But we never bind.

SyncedSynapse:
Thanks.

Yeah, i also noticed that it shouldn't happen... Maybe it's specific vendor code

PR #374, opened 2017-03-22
Introduces BaseDrawerActivity

poisdeux:
* BaseDrawerActivity enforces a common layout which allows us to more
     easily add new common UI elements
   * All activities incorporating a drawer should extend BaseDrawerActivity
     Currently CastActivity is the only exception.

Note that I also refactored PVRActivity although I am unable to test the PVR section. Could someone test this for me? Especially the drawer icon should be tested, to see if it changes correctly from hamburger to arrow and back.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/addon/AddonsActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/addon/AddonsActivity.java
@@ -19,77 +19,47 @@ import android.annotation.TargetApi;
 import android.content.Intent;
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
-import android.support.v4.app.FragmentTransaction;
-import android.support.v4.widget.DrawerLayout;
-import android.support.v7.app.ActionBar;
-import android.support.v7.widget.Toolbar;
-import android.view.Menu;
+import android.text.TextUtils;
 import android.view.MenuItem;
-import android.view.Window;
 
 import org.xbmc.kore.R;
 import org.xbmc.kore.ui.AbstractFragment;
-import org.xbmc.kore.ui.BaseActivity;
-import org.xbmc.kore.ui.generic.NavigationDrawerFragment;
+import org.xbmc.kore.ui.BaseMediaActivity;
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/xbmc/kore/ui/sections/addon/AddonsActivity.java
```diff
@@ -20,76 +20,48 @@
...
-        super.onCreate(savedInstanceState);
-        setContentView(R.layout.activity_generic_media);
-
-        // Set up the drawer.
-        navigationDrawerFragment = (NavigationDrawerFragment)getSupportFragmentManager()
-                .findFragmentById(R.id.navigation_drawer);
-        navigationDrawerFragment.setUp(R.id.navigation_drawer, (DrawerLayout) findViewById(R.id.drawer_layout));
-
-        Fragment fragment;
         if (savedInstanceState == null) {
```
SyncedSynapse: oops
!=

poisdeux: I don't understand. If the bundle is not null this activity is recreated because of a configuration change. In which the fragment does not need to be recreated, right?

poisdeux:
I noticed that setting the navigation icon state is not preserved on device configuration changes. :( Debugging/fixing this now.

app/src/main/java/org/xbmc/kore/ui/sections/addon/AddonsActivity.java
SyncedSynapse: Take a look at the next line of code that remains. Better check in your local copy instead of here, which is probably confusing you.

poisdeux: Ah I see now. I already changed the code to fix the issues on device configuration changes and couldn't relate it to your comment.  Sorry about that.

poisdeux:
I think I fixed all issues with configuration changes. I renamed BaseDrawerActivity to BaseMediaActivity as we have some activities using a drawer that don't use the generic media layout.
Again, I was not able to test the PVR section. Hopefully that works as expected...

Issue #65, opened 2015-04-14
File browser shows wrong directory names


rigid:
It seems as if some directories are named like the scraper result of a file inside that directory.

I have this directory structure:
* 60s
 * foo.mp3
* 70s
 * bar.mp3
* 80s
 * baz.mp3
* 90s

Kore displays the directory like this:

60s
Bar - The Bar Song
80s
90s

DanhKE6D:
@SyncedSynapse It seems to me that if itemFile.filetype is a directory, itemFile.label should be used as the title for the directory. I don't have any song/video in my database so I did not see this behavior, but if someone already has lots of songs/videos in the database, the json rpc response may be different. In the code for newInstanceFromItemFile, title was changed to something different than itemFile.label before it was FileLocation object was instantiated.

SyncedSynapse:
This is not an issue as far as i'm concerned. If the scrapper has "indexed" those files, the info obtained by it is shown, instead of the simple file name. More useful imo.

rigid:
So what's the point of the "File" feature then and why not use the library in the first place? it can be argued if a "file browser" should display real filenames or metadata;
but naming the directory using scraper info from a file _inside_ this directory is wrong.

SyncedSynapse:
It's a conscious decision to shown all the metadata that is returned by Kodi.

Regarding the naming of the directory: Kore is only showing what is returned by Kodi. It is not digging into the directory to check if there is only one file and naming it accordingly. That issue is probably in the scrapper. If you go to the music->files section in Kodi, how does that directory appear?

rigid:
Then it's the same behaviour as in "Music", just using the real directory structure (which is incorrectly named by now).

Music -> Files shows only real filenames. In my example:

60s
70s
80s
90s

no metadata neither on directories nor on files. Even the file suffix is shown. (btw. i'm using kodi nightlies, please tell me if you need more info.)

SyncedSynapse:
In Kodi i also see the metadata when navigating through files. We probably have some different setting.

Anyway, Kore is showing the metadata for each dir/file and i think that's useful. If the metadata shown/obtained by the scrapper is incorrect, that should be dealt with on the scrapper.

rigid:
i wonder how "metadata for each directory" could work. There is not even a remote way any scraper could name the directory any better than I did already.

and still: no, Kore is not showing metadata for each directory but instead for some random file inside it.

SyncedSynapse:
[...]

PR #379, opened 2017-04-23
File view fixes

### LINKED ISSUE #119: Sorting: Files: List ist not sorted by filename
If I browse files via "Files" the list is not sorted by filename instead by "display name".

In case of music, if I have e.g. two files "01 Zong.mp3" and "02 Song.mp3" the second one displays first, because its ID3-Tag says "Song".

The same happens with files of episodes of tv series.

It would be nice to have the option to sort these lists by filename. :)

thx

sebastian-de:
Sort file lists by path and show filename in details
These patches fix #119 and #65
Screenshots: https://github.com/xbmc/Kore/issues/65#issuecomment-296399554

It also contains a fix for http://forum.kodi.tv/showthread.php?tid=278792

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/file/MediaFileListFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/file/MediaFileListFragment.java
@@ -54,6 +54,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Queue;
 import java.util.regex.Pattern;
+import java.io.File;
 
 /**
  * Presents a list of files of different types (Video/Music)
@@ -718,7 +719,7 @@ public class MediaFileListFragment extends AbstractListFragment {
                     break;
                 case ListType.ItemBase.TYPE_ALBUM:
                     title = itemFile.displayartist + " | " + itemFile.album;
-                    details = itemFile.file.substring(itemFile.file.lastIndexOf("/", itemFile.file.length()-2)+1, itemFile.file.length()-1);
+                    details = new File(itemFile.file).getName();
                     artUrl = itemFile.thumbnail;
                     sizeDuration = (itemFile.size > 0) && (itemFile.duration > 0) ?
                                    UIUtils.formatFileSize(itemFile.size) + " | " + UIUtils.formatTime(itemFile.duration) :
@@ -727,7 +728,7 @@ public class MediaFileListFragment extends AbstractListFragment {
                     break;
                 case ListType.ItemBase.TYPE_SONG:
                     title = itemFile.label;
-                    details = itemFile.file.substring(itemFile.file.lastIndexOf("/", itemFile.file.length())+1);
+                    details = new File(itemFile.file).getName();
                     artUrl = itemFile.thumbnail;
                     sizeDuration = (itemFile.size > 0) && (itemFile.duration > 0) ?
                                    UIUtils.formatFileSize(itemFile.size) + " | " + UIUtils.formatTime(itemFile.duration) :
--- a/app/src/main/java/org/xbmc/kore/ui/sections/file/MediaFileListFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/file/MediaFileListFragment.java
@@ -512,16 +512,15 @@ public class MediaFileListFragment extends AbstractListFragment {
                                     case R.id.action_play_from_this_item:
                                         mediaQueueFileLocation.clear();
                                         FileLocation fl;
+                                        // start playing the selected one, then queue the rest
+                                        mediaQueueFileLocation.add(loc);
+                                        playMediaFile(loc.file);
                                         for (int i = position + 1; i < fileLocationItems.size(); i++) {
                                             fl = fileLocationItems.get(i);
                                             if (!fl.isDirectory) {
                                                 mediaQueueFileLocation.add(fl);
                                             }
                                         }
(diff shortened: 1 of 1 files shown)


SyncedSynapse:
Thanks for this. I have some comments on it tough, so i would appreciate if you could review them.

app/src/main/java/org/xbmc/kore/ui/sections/file/MediaFileListFragment.java
```diff
@@ -323,7 +324,7 @@ private void browseDirectory(final FileLocation dir) {
 
         Files.GetDirectory action = new Files.GetDirectory(dir.file,
                 mediaType,
-                new ListType.Sort(ListType.Sort.SORT_METHOD_LABEL, true, true),
+                new ListType.Sort(ListType.Sort.SORT_METHOD_PATH, true, true),
```
SyncedSynapse: Ideally this would be an option on the overflow menu, like it is for instance in the movies section. Nvertheless, if you're not confortable implementing it, leave it and i'll merge it as is.

app/src/main/java/org/xbmc/kore/ui/sections/file/MediaFileListFragment.java
```diff
@@ -511,16 +512,15 @@ public boolean onMenuItemClick(MenuItem item) {
                                     case R.id.action_play_from_this_item:
                                         mediaQueueFileLocation.clear();
                                         FileLocation fl;
+                                        // start playing the selected one, then queue the rest
+                                        mediaQueueFileLocation.add(loc);
+                                        playMediaFile(loc.file);
```
SyncedSynapse: Slight possible bug here. `PlayMediaFile` needs to be the last thing on this block, as it also sends the queued items to Kodi, so `mediaQueueFileLocation` needs to be full before calling it.
Actually i don't think that `PlayMediaFile` should do this, but it's how it's implemented right now, so to avoid a possible bug this needs to be left as the last line before the return.

app/src/main/java/org/xbmc/kore/ui/sections/file/MediaFileListFragment.java
```diff
@@ -717,9 +717,17 @@ public static FileLocation newInstanceFromItemFile(Context context, ListType.Ite
                     artUrl = itemFile.thumbnail;
                     break;
                 case ListType.ItemBase.TYPE_ALBUM:
+                    title = itemFile.displayartist + " | " + itemFile.album;
+                    details = new File(itemFile.file).getName();
```
SyncedSynapse: I don't particularly like that we are creating a `File` object for each item on the directory. I get that it is the easiest way to get the filename all parsed up, but it is kind of wasteful.
Furthermore there's a probable bug here: take a look at the `getName` implementation and you'll notice that it used a `fileSystemSeparator` to get the last part of the name. The thing is, this always returns Android's file system separator (ext4 `\`), not the one where Kodi resides, so if Kodi is on windows, the path probably contains `/` and this won't work. Haven't tested this tough, it's just an hypothesis.

app/src/main/java/org/xbmc/kore/ui/sections/file/MediaFileListFragment.java
```diff
@@ -717,9 +717,17 @@ public static FileLocation newInstanceFromItemFile(Context context, ListType.Ite
...
+                    artUrl = itemFile.thumbnail;
+                    sizeDuration = (itemFile.size > 0) && (itemFile.duration > 0) ?
+                                   UIUtils.formatFileSize(itemFile.size) + " | " + UIUtils.formatTime(itemFile.duration) :
+                                   (itemFile.size > 0) ? UIUtils.formatFileSize(itemFile.size) :
+                                   (itemFile.duration > 0)? UIUtils.formatTime(itemFile.duration) : null;
+                    break;
                 case ListType.ItemBase.TYPE_SONG:
                     title = itemFile.label;
-                    details = itemFile.displayartist + " | " + itemFile.album;
+                    details = new File(itemFile.file).getName();
```
SyncedSynapse: Same as the last one

sebastian-de:
Thank you very much for reviewing my PR, I'll rework some of the things according to your comments in the coming days.

sebastian-de:
I only use Linux, so I can't check if getFilenameFromPath() works with Windows paths, but I copied most of it from getParentDirectory(), so this should work.

app/src/main/java/org/xbmc/kore/ui/sections/file/MediaFileListFragment.java
sebastian-de: I will need some time to implement this, so it would be nice if you could merge this change as is. I'll try to look into this though - if I succeed I'll send another PR.
I'm still not completely happy with SORT_METHOD_PATH anyway, as some files don't seem to get sorted by path/filename, but I can't figure out why.

bigretromike:
Will this fix problem with sorting episodes by episode number and not by episode title ? Is there a way to pick sort option like in webview ?

sebastian-de:
All I can say is that addons will be affected, since the addon section uses MediaFileListFragment. How exactly will depend on the virtual file tree of your addon. You can find details on  how Kore handles addons here: https://github.com/xbmc/Kore/pull/239
It'll be best if you try current Kore master and see if your issue is fixed.
With my changes you'll see the filename of each item in its second row (in this case the virtual filename returned by your addon).

Issue #81, opened 2015-05-25
[Feature Request] Need Favorites in menu


gvkt:
Don't know if this is possible.

It would be useful to have a Favorites menu item in the menu panel to bring up Kodi Favorites in one tap. If possible, displaying the Favorites list on Kore itself would be ideal (especially for no-display use for Audio) but this may be difficult to do.

Perhaps there is a workaround I am not aware of.

juup:
I'm interested in the same feature. Would be nice to control radio (plugin) with it.

micressor:
+1 for this feature!

jponzo:
+1!

jp438:
I would be very interested in this feature for the purposes of radio as well! +1

SyncedSynapse:
Unfortunately there are some issues with this. It is possible to open the favorites window on Kodi, but after being opened it isn't possible to dismiss it using the remote. I'm not sure, but i believe this is related to a known issue with threading on Kodi.

Given this, introducing a favorites button on the remote isn't really helpful, as it will lead to a situation where the user gets stuck.

jp438:
Thanks for the reply! Would it maybe be possible to sync the contents of the favourites menu with Kore so that it can start media directly outside the Kodi interface. (Just like you can browse your library entirely on Kore and select something to play without having to navigate Kodi's menus)

SyncedSynapse:
I guess something along those lines could be done, not sure how well if it would work though. Keeping this open.

dzsodzso63:
So +1 for this 
Sync favorites or if not possible, maybe a separate favorites menu for kore app separately 
Quite a problem if you use a deep menu item a lot (E. G. in a plugin)

wstephen:
Yes +1
The movies, tvshows, musics shortcuts on the Navigation Drawer are so great. But, the Kodi's Favorite Button (the Star Button) is so difficult to be reached by Kore Remote.

Understand that reasons for features such as TV Channel Switching, recording are strong and valid, developers may be hammering how to make a good room in the tiny screens of our mobile. However, I strongly suggest the team should consider making Kodi Favourite Shortcut available on the Navigation Drawer just like what they've done to movies and tvshows.

Kodi Favorites can store TV Channels with logos, and it can store a myriads of Kodi addons and programs. It allows renaming too. How can Kore not availing this feature in its remote function?

Please, please consider! Imagine the degree of convenience and how pleasant it will look.

unalterable:
[...]

PR #384, opened 2017-05-03
Support getting favourites in Kore (#81)

R4md4c:
This PR introduces getting favourites from Kodi to Kore. 
The favourite types that are supported are just the `window` and `media`.

Here is a demonstration of the favourites list that is retrieved from my Kodi box, please note that I'm using the same layout as in the PVRChannelsListFragment.

favourites list

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/generic/NavigationDrawerFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/generic/NavigationDrawerFragment.java
@@ -47,6 +47,7 @@ import org.xbmc.kore.host.HostInfo;
 import org.xbmc.kore.host.HostManager;
 import org.xbmc.kore.ui.sections.addon.AddonsActivity;
 import org.xbmc.kore.ui.sections.audio.MusicActivity;
+import org.xbmc.kore.ui.sections.favourites.FavouritesActivity;
 import org.xbmc.kore.ui.sections.file.FileActivity;
 import org.xbmc.kore.ui.sections.hosts.HostManagerActivity;
 import org.xbmc.kore.ui.sections.remote.RemoteActivity;
@@ -81,9 +82,11 @@ public class NavigationDrawerFragment extends Fragment {
             ACTIVITY_TVSHOWS = 3,
             ACTIVITY_MUSIC = 4,
             ACTIVITY_PVR = 5,
-            ACTIVITY_FILES = 6,
-            ACTIVITY_ADDONS = 7,
-            ACTIVITY_SETTINGS = 8;
+            ACTIVITY_FAVOURITES = 6,
+            ACTIVITY_FILES = 7,
+            ACTIVITY_ADDONS = 8,
+            ACTIVITY_SETTINGS = 9;
+
 
     // The current selected item id (based on the activity)
     private static int selectedItemId = -1;
@@ -143,6 +146,7 @@ public class NavigationDrawerFragment extends Fragment {
                 R.attr.iconTvShows,
                 R.attr.iconMusic,
                 R.attr.iconPVR,
+                R.attr.iconFavourites,
                 R.attr.iconFiles,
                 R.attr.iconAddons,
                 R.attr.iconSettings,
@@ -179,6 +183,10 @@ public class NavigationDrawerFragment extends Fragment {
             items.add(new DrawerItem(DrawerItem.TYPE_NORMAL_ITEM, ACTIVITY_PVR,
                                      getString(R.string.pvr),
                                      styledAttributes.getResourceId(styledAttributes.getIndex(ACTIVITY_PVR), 0)));
+        if (shownItems.contains(String.valueOf(ACTIVITY_FAVOURITES)))
+            items.add(new DrawerItem(DrawerItem.TYPE_NORMAL_ITEM, ACTIVITY_FAVOURITES,
+                    getString(R.string.favourites),
+                    styledAttributes.getResourceId(styledAttributes.getIndex(ACTIVITY_FAVOURITES), 0)));
         if (shownItems.contains(String.valueOf(ACTIVITY_FILES)))
             items.add(new DrawerItem(DrawerItem.TYPE_NORMAL_ITEM, ACTIVITY_FILES,
                                      getString(R.string.files),
@@ -344,6 +352,8 @@ public class NavigationDrawerFragment extends Fragment {
             return ACTIVITY_MUSIC;
         else if (activity instanceof PVRActivity)
             return ACTIVITY_PVR;
+        else if (activity instanceof FavouritesActivity)
+            return ACTIVITY_FAVOURITES;
         else if (activity instanceof FileActivity)
             return ACTIVITY_FILES;
         else if (activity instanceof AddonsActivity)
@@ -364,6 +374,7 @@ public class NavigationDrawerFragment extends Fragment {
         activityItemIdMap.put(ACTIVITY_REMOTE, RemoteActivity.class);
         activityItemIdMap.put(ACTIVITY_MOVIES, MoviesActivity.class);
         activityItemIdMap.put(ACTIVITY_MUSIC, MusicActivity.class);
+        activityItemIdMap.put(ACTIVITY_FAVOURITES, FavouritesActivity.class);
         activityItemIdMap.put(ACTIVITY_FILES, FileActivity.class);
         activityItemIdMap.put(ACTIVITY_TVSHOWS, TVShowsActivity.class);
         activityItemIdMap.put(ACTIVITY_PVR, PVRActivity.class);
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/xbmc/kore/ui/sections/favourites/FavouritesListFragment.java
```diff
@@ -0,0 +1,224 @@
...
+import org.xbmc.kore.jsonrpc.type.FavouriteType;
+import org.xbmc.kore.utils.LogUtils;
+import org.xbmc.kore.utils.UIUtils;
+
+import java.util.List;
+
+import butterknife.ButterKnife;
+import butterknife.InjectView;
+
+public class FavouritesListFragment extends Fragment implements SwipeRefreshLayout.OnRefreshListener {
```
poisdeux: Can't this extend AbstractListFragment?

R4md4c: Well, I was taking the PVRChannelsListFragment as a reference, and it was extending just Fragment, so I made it similar to PVRChannelsListFragment. Would you like me to extend from AbstractListFragment and do the necessary changes?

poisdeux: Yes please. The PVR section hasn't been refactored yet. It's on my TODO list...

SyncedSynapse: Actually it's on my todo list... sorry for the delay...

R4md4c: Done, But I noticed that AbstractListFragment doesn't handle pagination, should I even handle the pagination from the Favourites? or does `Favourites.GetFavourites` returns all the items when no Limit is specified?

app/src/main/java/org/xbmc/kore/ui/generic/NavigationDrawerFragment.java
```diff
@@ -81,9 +82,11 @@
...
             ACTIVITY_MUSIC = 4,
             ACTIVITY_PVR = 5,
-            ACTIVITY_FILES = 6,
-            ACTIVITY_ADDONS = 7,
-            ACTIVITY_SETTINGS = 8;
+            ACTIVITY_FAVOURITES = 6,
+            ACTIVITY_FILES = 7,
+            ACTIVITY_ADDONS = 8,
+            ACTIVITY_SETTINGS = 9;
+
```
SyncedSynapse: Slight bug over here. Actually it's just proof that this part of the code isn't very robust, but it is what it is right now, so we'll have to live with it...

You can't change the ids of the existing constants because these are already stored in the user's preferences (to show/hide the menu items). Meaning that what is stored in preferences is this actual number, so any change here will collide with that.
You'll have to keep the current activities as is and add ACTIVITY_FAVORITES as the 9. A side effect is that when the user receives the upgraded version, the favorites link won't be shown on the menu, because it won't be stored in the preferences... He will have to manually go into settings->side menu items and check the favorites. 

Alternatively (and ideally) we need to, ONCE, add this entry to the preference, which can be done on the remote activity. This is similar to the code that was removed in 39ee0b6634bffe5451912f63f67cf978632f92d7 (this was the check to introduce the PVR section, and was removed in the next version as it should no longer be necessary).
If this is not clear to you or you're not confortable with it, don't mind it, just don't change the previous IDs, add to them.

app/src/main/java/org/xbmc/kore/ui/sections/favourites/FavouritesListFragment.java
```diff
@@ -0,0 +1,224 @@
...
+            }
+
+            final FavouriteItemViewHolder vh = (FavouriteItemViewHolder) convertView.getTag();
+            final FavouriteType.DetailsFavourite favouriteDetail = getItem(position);
+
+            // We don't need the context menu here.
+            vh.contextMenu.setVisibility(View.GONE);
+
+            vh.titleView.setText(favouriteDetail.title);
+            vh.detailView.setText(favouriteDetail.type);
```
SyncedSynapse: Ideally the `favoriteDetail.type` shouldn't be used directly as the text. We should have the possible values on `strings.xml` and use those, so that they could be translated/made user friendly.

app/src/main/java/org/xbmc/kore/ui/generic/NavigationDrawerFragment.java
R4md4c: I tried to use the first suggestion, because I'm afraid that if I went with the second one, I may introduce more bugs.


So to be in the safe side I restored the previous order and assigned 9 to ACTIVITY_FAVOURITES, then added 9 to the `entry_values_nav_drawer_items` in arrays.xml. 

But I've noticed that the number of items in the `entries_nav_drawer_items` doesn't match the number of items in `entry_values_nav_drawer_items`. 

Do you know what is the reason behind that? Because I tried adding "Favourites" under `entries_nav_drawer_items` but noticed the mismatch between the index of that entry and the corresponding index at `entry_values_nav_drawer_items`.

R4md4c: I went with your first suggestion because it felt simpler, and I was afraid that I may introduce other bugs if I went with the second approach.

poisdeux:
Two more comments from me. First, your files are missing copyright statements. 
Second, when selecting a favourite item I would not expect it to start playing immediately. For movies, music, TV shows I would expect to be able to add them to the queue or play them using a context menu. For addons I would expect the addon to be opened on Kore so I can use browser functionality.
Also, the default is when an item is selected to play, Kore will switch to the remote fragment.

app/src/main/res/values/arrays.xml
```diff
@@ -64,6 +64,7 @@
         <item>@string/pvr</item>
         <item>@string/files</item>
         <item>@string/addons</item>
+        <item>@string/favourites</item>
```
SyncedSynapse: I think this should be in the same order as in the menu, so this entry should be after pvr, before files.

app/src/main/res/values/arrays.xml
```diff
@@ -112,7 +113,7 @@
         <item>5</item>
         <item>6</item>
         <item>7</item>
-        <item>8</item>
+        <item>9</item>
```
SyncedSynapse: Also here, after 5, before 6

SyncedSynapse:
I agree with @poisdeux regarding automatically switching to the remote when an item starts playing, but i don't agree with him regarding not expecting immediately start playing - that's what i expected.
A context menu on each item would be nice, but i don't see it as mandatory.

In any case @R4md4c thanks for this, and i think this is good to be merged as is.

R4md4c:
@poisdeux I could add a context menu to the media types that has Play or Queue Item, but I think this is not needed with the window type, also AFAIK there is no way to differentiate between media types that are coming from the favourites, that means I won't know if this item is a movie or a music or a TV Show.

Another thing I noticed is that in Krypton Addons can't be added to the favourites.

poisdeux:
@R4md4c Yes, I think you are right that the context menu only makes sense on media types.

Addons can be added in Krypton to favourites. You will need to get the file listing for the addons. From the main menu select *Add-ons*, then one of the categories for example *Video-addons*. Now , if you get the context menu for an addon you should be able to add it as favourite.

R4md4c:
@poisdeux I'm trying to queue a favourite item using this method[1], but it seems that it expects a playlist type for the media to get added to that type. The problem with that is, I don't have the media type of that favourite item, is there an API endpoint that you know of that takes the favourite item path it returns the file media type?

[1] https://github.com/xbmc/Kore/blob/master/app/src/main/java/org/xbmc/kore/utils/MediaPlayerUtils.java#L86

poisdeux:
Ah I see. AFAICT there is no way to get the current active playlist to determine which playlist we should add the item too. So, unfortunately I think queueing is not possible.

R4md4c:
Does that mean I should keep the context menu hidden and settle on just showing the remote after favourite item is clicked?

poisdeux:
Yes I think so. Unless @SyncedSynapse knows a way on how to implement queueing.

SyncedSynapse:
Yes for me. Changing the order of the items in preferences and switching to the remote after playing would be nice.

R4md4c:
I've done the requested changes, after I play the media the remote is shown as soon as the media starts playing in Kodi.

poisdeux:
Ok, I kinda messed up with the review feature. Sorry about that. I should have pressed the comment button. 
Anyway, the code as it currently stands has an issue where the Favourite screen crashes Kore when there are no favourites.

R4md4c:
Should I send the fix as a new PR? or what is the usual flow in these cases?

PR #396, opened 2017-06-02
Change position of play/pause and stop button

### LINKED ISSUE #393: Change postion of play/pause and stop button
On the current playback screen the order is:
prev back stop play forward next
however in every other (non-)kodi player the stop and play buttons are switched

mueller-ma:
fixes #393

Even kodi itself uses this order, so we should use it in the app, too

Signed-off-by: mueller-ma <mueller-ma@users.noreply.github.com>

### DIFF
--- a/app/src/main/res/layout-land/fragment_remote.xml
+++ b/app/src/main/res/layout-land/fragment_remote.xml
@@ -108,14 +108,6 @@
                 android:src="?attr/iconRewind"
                 android:contentDescription="@string/rewind"/>
             <ImageButton
-                android:id="@+id/stop"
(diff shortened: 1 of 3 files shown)


SyncedSynapse:
@MartijnKaijser Is there a guideline for the button order?

MartijnKaijser:
i did a quick search however could not find a definitive guide.
Maybe @phil65 knows as he works in UX design.

phil65:
no, there is no guideline. best is probably to align with the most popular players, at least thats what we tried with the default skin.
+1 for this PR.

PR #405, opened 2017-06-16
Implemented MediaProgressIndicator widget to allow for code sharing

poisdeux:
Created a compound view (widget) to show media progression information. This allows us to reuse the code in other parts. Such as the now playing panel, which is still under construction.
Refactored the NowPlayingFragment to use the new compound view.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/jsonrpc/type/PlayerType.java
+++ b/app/src/main/java/org/xbmc/kore/jsonrpc/type/PlayerType.java
@@ -283,6 +283,19 @@ public class PlayerType {
             this.milliseconds = milliseconds;
         }
 
+        /**
+         * Converts second duration into a hour, minutes, seconds duration
+         * @param seconds
+         */
+        public PositionTime(int seconds) {
+            this.hours = seconds / 3600;
+            int remainder = seconds - (this.hours * 3600);
+            this.minutes = remainder / 60;
+            remainder = remainder - (this.minutes * 60);
+            this.seconds = remainder;
+            this.milliseconds = 0;
+        }
+
         public JsonNode toJsonNode() {
             final ObjectNode node = objectMapper.createObjectNode();
             node.put(HOURS, hours);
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/xbmc/kore/ui/widgets/MediaProgressIndicator.java
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Sets the play speed for the progress indicator
+     * @param speed set to 0 to stop updating the progress indicator.
+     */
+    public void setSpeed(int speed) {
+        if( speed == this.speed )
+            return;
+
+        this.speed = speed;
+        this.progressIncrement = speed * (SEEK_BAR_UPDATE_INTERVAL/1000);
```
poisdeux: I'm using speed here to allow for different speeds. Although the original implementation only supported speed 1 I can't think of a reason why it can't be different. Is there a problem with speeds greater or smaller (besides 0) than 1?

app/src/main/java/org/xbmc/kore/jsonrpc/type/PlayerType.java
```diff
@@ -283,6 +283,19 @@ public PositionTime(int hours, int minutes, int seconds, int milliseconds) {
             this.milliseconds = milliseconds;
         }
 
+        /**
+         * Converts second duration into a hour, minutes, seconds duration
+         * @param seconds
+         */
+        public PositionTime(int seconds) {
+            this.hours = seconds / 3600;
+            int remainder = seconds - (this.hours * 3600);
```
SyncedSynapse: Just curious, why don't you use % operator?

app/src/main/java/org/xbmc/kore/ui/widgets/MediaProgressIndicator.java
SyncedSynapse: Can't remember, but IIRC in the original implementation, the update runnable was only active when the speed was 1. I believe that as soon as a FFWD or RWD was issued the updater was paused.

app/src/main/java/org/xbmc/kore/jsonrpc/type/PlayerType.java
poisdeux: Using modulo to get the remaining seconds would require an additional division, which is more expensive than multiplication and subtraction.

PR #406, opened 2017-06-16
Implemented VolumeLevelIndicator widget

poisdeux:
Refactored NowPlayingFragment to use new widget. Also refactored UIUtils.highlightImageView to reduce code duplication.


SyncedSynapse:
After merging #405 this isn't merging cleanly. Can you please take a look.

poisdeux:
fixed

PR #408, opened 2017-06-20
Implemented RepeatModeButton widget

poisdeux:
Its main purpose is to enable testing the repeat mode functionality.


SyncedSynapse:
Same here... Conflicts with #405 (and likely #406)

poisdeux:
Also fixed.

PR #410, opened 2017-06-20
Implemented a HighlightButton widget

poisdeux:
This can be used by buttons that need to be highlighted such as
the shuffle and the mute buttons.

I've refactored NowPlayingFragment to use the new widget.
The main reason for implementing this widget is to allow testing of mute and shuffle button states.


SyncedSynapse:
Same here.

poisdeux:
Last one... fixed.

Issue #402, opened 2017-06-11
[Bug] Sort is case sensitive


dale2507:
Sorting TV shows by name is case sensitive. This means that i (lowercase) appears after Z (uppercase). I would expect it to be case insensitive.

PR #411, opened 2017-06-23
Changed sort of names and titles to be case insensitive

tomerf:
Make the sort of movies/tv show titles and artists/album names case insensitive, fixing issue #402.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/audio/SongsListFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/audio/SongsListFragment.java
@@ -192,7 +192,7 @@ public class SongsListFragment extends AbstractCursorListFragment {
                 MediaProvider.Qualified.ALBUMS_THUMBNAIL
         };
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/xbmc/kore/ui/sections/audio/SongsListFragment.java
```diff
@@ -192,7 +192,7 @@ public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
                 MediaProvider.Qualified.ALBUMS_THUMBNAIL
         };
 
-        String SORT = MediaDatabase.sortCommonTokens(MediaProvider.Qualified.SONGS_TITLE) + " ASC";
+        String SORT = MediaDatabase.sortCommonTokens(MediaProvider.Qualified.SONGS_TITLE) + " COLLATE NOCASE ASC";
```
SyncedSynapse: Over here it might have some impact on performance, in big libraries. Let's keep it, but if someone complains we'll revisit this.

PR #427, opened 2017-07-31
Universal Subtitles Options Dialog

### LINKED ISSUE #375: [Feature Request] Dedicated subtitle button
Many remotes controls (real and software) have a dedicated subtitle button that we can use to quickly turn on/off the subtitle as well as select the language.

This is useful because some sections of a movie, we want to just check what they said then turn subtitles off again.

Right now the implementation is not very user friendly
requiring going through a hidden drop down menu then choose subtitle then we can choose  the subtitle language or turn on/off subtitle.

This is extremely cumbersome.

1. We need a dedicated subtitle button always showing.
2. It would also be good if we can hide the other unused buttons as well e.g. Play single, play all button,
3. Ideally we should be able to choose what buttons we want to show/hide e.g. subtitle, audio etc..

n123al:
this commit fixes #375

A class (app\src\main\java\org\xbmc\kore\ui\generic\SubtitlesOptionsDialog.java) was developed that handles the display of a dialog with options related to subtitles (Download subtitles, Sync subtitles, none). This class was used to add the subtitle button to the widget NowPlayingPanel and the RemoteFragment. Furthermore, any developer can use this class in order to implement a subtitle button to any activity.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
+++ b/app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
@@ -1,0 +1,313 @@
+package org.xbmc.kore.ui.generic;
+
+
+import android.app.Activity;
+
+import android.os.Handler;
+import android.support.v4.app.FragmentActivity;
+import android.text.TextUtils;
+import android.widget.Toast;
+
+import org.xbmc.kore.R;
+import org.xbmc.kore.host.HostConnectionObserver;
+import org.xbmc.kore.host.HostInfo;
+import org.xbmc.kore.host.HostManager;
+import org.xbmc.kore.jsonrpc.ApiCallback;
+import org.xbmc.kore.jsonrpc.ApiMethod;
+import org.xbmc.kore.jsonrpc.HostConnection;
+import org.xbmc.kore.jsonrpc.method.Addons;
+import org.xbmc.kore.jsonrpc.method.GUI;
+import org.xbmc.kore.jsonrpc.method.Input;
+import org.xbmc.kore.jsonrpc.method.Player;
+import org.xbmc.kore.jsonrpc.type.ListType;
+import org.xbmc.kore.jsonrpc.type.PlayerType;
+import org.xbmc.kore.utils.LogUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+/**
+ * Dialog that presents a list of options related to subtitles.
+ */
+public class SubtitlesOptionsDialog
+        implements HostConnectionObserver.PlayerEventsObserver,
+                    GenericSelectDialog.GenericSelectDialogListener{
+
+
+    // An interface to be implemented by everyone interested in RemoteRequired events
+    public interface SubtitlesOptionsDialogListener {
+        void onRemoteRequired();
+    }
+
+    private List<SubtitlesOptionsDialogListener> listeners = new ArrayList<SubtitlesOptionsDialogListener>();
+
+    private static final String TAG = LogUtils.makeLogTag(SubtitlesOptionsDialog.class);
+
+    /**
+     * List of available subtitles
+     */
+    private List<PlayerType.Subtitle> availableSubtitles;
+    private int currentSubtitleIndex = -1;
+
+    // Number of explicitly added options for subtitles (to subtract from the
+    // number of subtitles returned by Kodi)
+    static final int ADDED_SUBTITLE_OPTIONS = 3;
+
+    /**
+     * Constants for the general select dialog
+     */
+    private final static int SELECT_SUBTITLES = -111;
+
+    /**
+     * Host manager from which to get info about the current XBMC
+     */
+    private HostManager hostManager;
+
+    /**
+     * Handler on which to post RPC callbacks
+     */
+    private Handler callbackHandler = new Handler();
+
+    /**
+     * Activity to communicate potential actions that change what's playing
+     */
+    private HostConnectionObserver hostConnectionObserver;
+
+    /**
+     * The current active player id
+     */
+    private int currentActivePlayerId = -1;
+
+    /**
+     * Default callback for methods that don't return anything
+     */
+    private ApiCallback<String> defaultStringActionCallback = ApiMethod.getDefaultActionCallback();
+
+
+
+    private FragmentActivity dialogActivity;
+
+    public SubtitlesOptionsDialog(Activity activity){
+        hostManager =HostManager.getInstance(activity);
+        hostConnectionObserver=hostManager.getHostConnectionObserver();
+        hostConnectionObserver.registerPlayerObserver(this, true);
+        dialogActivity =(FragmentActivity)activity;
+    }
+
+
+    /**
+     * Display subtitles dialog
+     *
+     */
+    public void Show(){
+        int selectedItem = -1;
+
+        String[] subtitles = new String[(availableSubtitles != null) ?
+                availableSubtitles.size() + ADDED_SUBTITLE_OPTIONS : ADDED_SUBTITLE_OPTIONS];
+
+        subtitles[0] = dialogActivity.getString(R.string.download_subtitle);
+        subtitles[1] = dialogActivity.getString(R.string.subtitle_sync);
+        subtitles[2] = dialogActivity.getString(R.string.none);
+
+        if (availableSubtitles != null) {
+            for (int i = 0; i < availableSubtitles.size(); i++) {
+                PlayerType.Subtitle current = availableSubtitles.get(i);
+                subtitles[i + ADDED_SUBTITLE_OPTIONS] = TextUtils.isEmpty(current.language) ?
+                        current.name : current.language + " | " + current.name;
+                if (current.index == currentSubtitleIndex) {
+                    selectedItem = i + ADDED_SUBTITLE_OPTIONS;
+                }
+            }
+        }
+
+        GenericSelectDialog dialog = GenericSelectDialog.newInstance(this,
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
```diff
@@ -0,0 +1,313 @@
+package org.xbmc.kore.ui.generic;
```
poisdeux: Missing copyright statement

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
```diff
@@ -0,0 +1,313 @@
...
+        hostConnectionObserver.registerPlayerObserver(this, true);
+        dialogActivity =(FragmentActivity)activity;
+    }
+
+
+    /**
+     * Display subtitles dialog
+     *
+     */
+    public void Show(){
```
poisdeux: methods should start with a lower case

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
```diff
@@ -0,0 +1,313 @@
...
+    private int currentActivePlayerId = -1;
+
+    /**
+     * Default callback for methods that don't return anything
+     */
+    private ApiCallback<String> defaultStringActionCallback = ApiMethod.getDefaultActionCallback();
+
+
+
+    private FragmentActivity dialogActivity;
```
poisdeux: I find `dialogActivity` confusing. Why not name it `fragmentActivity` or just `activity`?

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
```diff
@@ -0,0 +1,313 @@
...
+            }
+        }
+
+        GenericSelectDialog dialog = GenericSelectDialog.newInstance(this,
+                SELECT_SUBTITLES, dialogActivity.getString(R.string.subtitles), subtitles, selectedItem);
+        dialog.show(dialogActivity.getSupportFragmentManager(), null);
+
+    }
+
+    public void addListener(SubtitlesOptionsDialogListener toAdd) {
```
poisdeux: Not clear why you need a listener. You add the `RemoteFragment` and `NowPlayingPanel` as listeners but they don't do anything when `onRemoteRequired()` is called.

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
```diff
@@ -0,0 +1,313 @@
...
+    public void playerOnPause(PlayerType.GetActivePlayersReturnType getActivePlayerResult, PlayerType.PropertyValue getPropertiesResult, ListType.ItemsAll getItemResult) {
+
+        setNowPlayingInfo(getActivePlayerResult, getPropertiesResult, getItemResult);
+    }
+
+    /**
+     * Sets whats playing information
+     * @param getItemResult Return from method {@link org.xbmc.kore.jsonrpc.method.Player.GetItem}
+     */
+    private void setNowPlayingInfo(PlayerType.GetActivePlayersReturnType getActivePlayerResult,
```
poisdeux: This doesn't set now playing info....

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
```diff
@@ -0,0 +1,313 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+/**
+ * Dialog that presents a list of options related to subtitles.
+ */
+public class SubtitlesOptionsDialog
+        implements HostConnectionObserver.PlayerEventsObserver,
```
poisdeux: In my opinion a dialog shouldn't listen for player events unless Android manages its lifecycle. Although you unregister in `finalize` the GC is not run immediately. I believe this might result in objects still listening for player events even when the fragment is gone.

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
```diff
@@ -0,0 +1,313 @@
...
+                    GenericSelectDialog.GenericSelectDialogListener{
+
+
+    // An interface to be implemented by everyone interested in RemoteRequired events
+    public interface SubtitlesOptionsDialogListener {
+        void onRemoteRequired();
```
poisdeux: Interface is never really used.

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
```diff
@@ -0,0 +1,313 @@
...
+     * @param token
+     * @param which
+     */
+    public void onDialogSelect(int token, int which) {
+        switch (token) {
+            case SELECT_SUBTITLES:
+                Player.SetSubtitle setSubtitle;
+                // 0 is to download subtitles, 1 is for sync, 2 is for none, other is for a specific subtitle index
+                switch (which) {
+                    case 0:
```
poisdeux: Please use class variables (private final static int) for cases to make the code more readable.

poisdeux:
Thanks for contributing. I have a few questions/issues regarding your code. Please see my inline comments.

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
n123al: Will fix

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
n123al: The reason the dialog is listening the player’s events is to be able to have an up to date list of the available subtitles. I think a better solution to this is to create a function updateSubtitleList that will be called by the activity in order to provide the dialog object with the latest list of subtitles when needed.

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
n123al: The use of onRemoteRequired is to inform the activity which created an instant of this dialog that a screen with the ability to control kodi is needed e.g. action like sync, download subtitles they need a screen that will allow to perform some movement/select actions on Kodi, currently this is only done by NowPlayingPanel. Instead of just displaying always the NowPlayingPanel I decided to make it more generic using events and allow the developer chose what to display.

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
n123al: Will fix

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
n123al: will fix

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
poisdeux: Ah, I missed the implementation in NowPlayingPanel. Sorry about that.

https://github.com/xbmc/Kore/pull/427/files#diff-e3313a18ffdcf69828879c6a5097266eR50

app/src/main/java/org/xbmc/kore/ui/generic/SubtitlesOptionsDialog.java
poisdeux: Do you assume the user will keep the dialog open while searching and downloading subtitles in Kodi? I would expect them to close the dialog and use the remote (control pad) to manage Kodi.

SyncedSynapse:
Sorry for not replying earlier but i only now got to check this.

And i'm sorry to say, but i don't think that a subtitles button belongs on the medial panel, either in the remote or in the now playing panel. Those controls only have media controls, a subtitles button is out of place there. (Furthermore, the button is available in every situation, even when it isn't applicable - music playback, DVR, addons, etc - though that could be fixed.)

And i'm saying this as a user that uses subtitles a lot, so having a subtitles button so easily accessible would be a plus for me, but it doesn't make sense.

I haven't looked at the implementation, but from a UI point of view, i can't merge this. I'd like to reiterate @poisdeux 's thanks for the contribution though.

madmax2:
@SyncedSynapse 
Why don't you look at the real world of real life remote controls
rather than saying why the UI doesn't make sense.

In a real world remote control, there is usually a subtitle button
because that is one of the most used buttons besides the media control keys.
Even for DVR, PVR, the subtitle function is used here as well.

As for music, who cares if it is not used..
The real remote has the subtitle button and we don't use that for music playback either..

This is ridiculous that you don't add this merge because that is, one of the most needed button 
that we will actually use in the real world, rather than the current cumbersome implementation of how to turn on and off subtitles...

Just because you don't think this UI implementation makes sense to you, doesn't mean majority of regular users would think the same as you.

**Why not just add a feature, so people can added their own button that maps to a function that they want?
Then if you don't think subtitle button is useful for you, then don't add it.**

Anyway your decision will only drive people to use other kodi remote control apps..
So I hope you are happy with your decision.

As for myself, I have abandon using Kore as a remote control app because of the these poor UI design.

Issue #383, opened 2017-05-01
Remote buttons too small


shiba89:
Hi, since version 2.2.0 the remote buttons turned from this:
![screenshot_2017-05-01-18-06-28](https://cloud.githubusercontent.com/assets/3816409/25585078/33f1246a-2e99-11e7-97a2-28a43f73fff9.png)
to this:
![screenshot_2017-05-01-18-02-41](https://cloud.githubusercontent.com/assets/3816409/25585086/379d3e6e-2e99-11e7-8c05-78bf6c230bd1.png)

I think it started with https://github.com/xbmc/Kore/issues/299. That guy and I share a similar screen resolution, but with my 54 more vertical pixel I never had overlapping problems. Can this be somehow fine-tuned? Right now I'm on 2.1.0 because I never get the correct button with the small version.

poisdeux:
I believe the way it is now is as it was originally designed. I wouldn't mind increasing the remote size. However, I'm afraid everyone else who was always using the correct size will start complaining...

SyncedSynapse:
It's also my opinion that the PR that changed this changed it for the better, and it should be kept. It is also worth pointing out that, even at the smaller resolutions, the remote buttons are bigger (52dp) than what Google recomends as the minimum (48dp).

Nevertheless, maybe they can be tweaked somewhat in smaller resolutions. Currently they are 52dp, we could try adding 4-8 dp.

shiba89:
I made a side by side comparison:
![comparison](https://cloud.githubusercontent.com/assets/3816409/25662264/31b28492-3014-11e7-8060-23a052e59d74.png)
To me it seems that 2.1.0 is way more similar to the Google Play screenshot.

I also made a comparison of the remote with my thumb on the screen (I only have my old netbook webcam...):
![comparison2](https://cloud.githubusercontent.com/assets/3816409/25662822/343d942a-3016-11e7-8def-d15347459eac.png)
As you can see, with 2.2.0 my thumb almost covers all the remote buttons.

I'm not saying this should be reverted, because button overlapping is in fact a problem. Maybe vertically center the buttons could be enough? Oh well, I made another comparison :)
![comparison3](https://cloud.githubusercontent.com/assets/3816409/25663453/06f27376-3018-11e7-8730-325b6358a70a.png)

SyncedSynapse:
I think that vertically centering the buttons would be worse, as it would make the buttons further away from the natural position of the thumb.

What device are you using and what dp does it have?

[...]

PR #428, opened 2017-08-02
Implemented a scalable control pad (remote)

poisdeux:
* Refactored RemoteFragment and created a compound view for the
     actual remote. I called it ControlPad to make it more clear what
     its main function is.
   * Implemented a custom grid layout (SquareGridLayout) that will
     always be square.

The control pad uses the square grid layout instead of a relative
layout to position the buttons. By using the square grid layout the
control pad is sized based on available space instead of using the
static smallest width dimensions.
To prevent the control pad from getting too big on large devices
(smallest width 410 and larger) a blank space has been added between the
media panel and the control pad.

I included the sw400dp just to be sure people with a 6 inch device still be able to use the remote with one hand. I included the sw600dp value as the remote got really big on my Nexus 7. We might need to add even higher values for 10 inch tablets. But I assume people using 10 inch devices use both hands to control the device.

Below some screenshots:

* AVD resembling a Nexus 6 device.
nexus_6

* AVD resembling a Nexus 7 device.
nexus_7

* AVD resembling a Nexus 10 device.
nexus_10

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -115,6 +115,7 @@ dependencies {
     compile "com.android.support:cardview-v7:${supportLibVersion}"
     compile "com.android.support:preference-v14:${supportLibVersion}"
     compile "com.android.support:support-v13:${supportLibVersion}"
+    compile "com.android.support:gridlayout-v7:${supportLibVersion}"
 
     compile 'com.fasterxml.jackson.core:jackson-databind:2.5.2'
     compile 'com.jakewharton:butterknife:6.1.0'
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteFragment.java
@@ -15,24 +15,14 @@
  */
 package org.xbmc.kore.ui.sections.remote;
 
-import android.annotation.TargetApi;
-import android.content.res.ColorStateList;
-import android.content.res.Resources;
 import android.content.res.TypedArray;
-import android.graphics.BitmapFactory;
-import android.graphics.PorterDuff;
-import android.graphics.PorterDuffColorFilter;
-import android.graphics.drawable.BitmapDrawable;
 import android.os.Bundle;
 import android.os.Handler;
 import android.preference.PreferenceManager;
 import android.support.v4.app.Fragment;
 import android.view.LayoutInflater;
-import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
-import android.view.animation.Animation;
-import android.view.animation.AnimationUtils;
 import android.widget.ImageButton;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
@@ -56,8 +46,8 @@ import org.xbmc.kore.jsonrpc.method.Player;
 import org.xbmc.kore.jsonrpc.type.GlobalType;
 import org.xbmc.kore.jsonrpc.type.ListType;
 import org.xbmc.kore.jsonrpc.type.PlayerType;
+import org.xbmc.kore.ui.widgets.ControlPad;
 import org.xbmc.kore.utils.LogUtils;
-import org.xbmc.kore.utils.RepeatListener;
 import org.xbmc.kore.utils.UIUtils;
 import org.xbmc.kore.utils.Utils;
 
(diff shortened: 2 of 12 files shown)


SyncedSynapse:
@poisdeux Sorry for not looking at this earlier, but i've been on vacation (and i will still be).

I've tested this on my Nexus 5x and a MotoX and didn't noticed a difference, which is a good thing, but looking at the AVD image you posted of a Nexus 6, it looks like the remote got a lot bigger on it (which wouldn't be good imo). 

Have you tested physically on a Nexus 6?

SyncedSynapse:
@poisdeux if you read my previous comment, ignore it...

I'm not sure about this. On my Nexus 5x the d-pad is enormous, i think it looks bad and is less user-friendly than how it was. There's also a slight bug in the media panel on top, where the second line is getting cut (at least on my device).

I'm not sure that letting the d-pad scale is the right way to do this. The d-pad size is nowadays more or less set so that each button is a little bigger than what android recommends (48dp vs 84dp), and it usually aligns with the 3 middle icons on the bottom bar. With this PR the d-pad size can vary a lot, specially with new devices coming out that are taller than wider.

Haven't really thought this through, but as is i think it's a step back for the majority of users.

poisdeux:
I used the following criteria when designing this solution:

* phone: d-pad MUST be usable single handed with a thumb
* tablet: d-pad should be usable using one hand to hold the tablet and one hand to press the buttons.

So if you still can use the pad with your thumb from the hand holding the phone, I think the increase in size is only for the better. Less issues with pressing the wrong buttons accidentally and more usable for people with thick fingers.

I find it very hard to determine if people will like this or not. We have several reports from people  (GitHub issues and Googleplay) complaining about the buttons being too small and/or the pad being too small. 
Maybe we could extend the beta stage to get as much feedback as possible and maybe even post a news item about the beta trying to get as many people to test it out before we release a production version?

Just to illustrate the change here a screenshot on a Nexus 5x of the current control pad:
image


And here with the new scalable control pad:
image

SyncedSynapse:
Look at them side by side (and picture them way bigger, as they appear on a 5" device):

device-2017-08-14-131259 | device-2017-08-14-131848

Making the d-pad buttons bigger is probably going to reduce errors, but that's not necessarily an upgrade in usability, given that the thumb has to navigate farther between presses. I believe there's a trade-off here, and making the d-pad occupy the whole width of the screen is on the wrong side of that trade-off.

Then there's my biggest complaint: it looks wrong. It's not balanced with anything, the icons on the d-pad are much bigger than the others (and making then smaller won't solve the issue, as that would create a lot of empty space), the control isn't properly sized vs the rest of the elements, etc. Even if usability would be somewhat better with bigger buttons, i would be willing to sacrifice that little (potential) improvement for the improvement in visuals. This point is subjective i agree.

Additionally, from a design perspective, this screen somewhat tries to mimic a physical device (it's not skeumorphic, but the d-pad tries to be similar to a physical one). I'm not a designer, but i believe that in these situations it's relevant to make sure that the size of those elements is similar to the physical ones (maybe a little bigger, to compensate the fact that there's no physical touch feedback), so we need control over those sizes. That can only be done through the use of fixed DP sizes. Letting the d-pad size vary according to the screen size is too much variation imo. 

Finally, i haven't seen many people complaining about the d-pad size. There's the github issue regarding too small buttons, but that could and should be dealt with taking into consideration the specific device the user is using. If that was a widespread issue, there would be a lot of reports about it, and i haven't seen that. Obviously there will always be people complaining that the d-pad is too small, and there will also always be people complaining that the d-pad is too big...

I'm also not a fan of putting this design on beta test. I'm not opposed to it but i don't think it's the right way to go about this. I believe Fritsch has the following on its sig: "Your most vocal users are not reflective of your userbase" and i totally subscribe that. We would get feeback (and i bet that some of it would be very vocal) alright, but nothing guarantees that that would be representative feedback.

Side note: The bug on the media panel (the second line being cut) is probably due to the fact that the d-pad is being vertically sized based on the width of the display, then the spacer is added above that (fixed space), and then there's not enough space for the media panel as it was designed, so something gets chopped.

poisdeux:
As I see it we have the following options:

1. Don't fix issue #383 
2. Accept deviations in size and use swNdp sizes for the spacer to minimize the deviation
3. Only apply the scalable control pad for devices < sw360dp and use the current static control pad sizes for devices >= sw360dp

As I understand from your last comment I think you'll prefer option 3. So I already anticipated on that and made it so in the last commit. If you approve I'd like to squash the two commits before merging this to master.

SyncedSynapse:
@poisdeux Sorry for the late reply. I would go with this version for now, and maybe revisit this again later on (after the release).

Travis is complaining, and i don't understand why. I noticed that you've added a debug flag, so i suppose you'll wait until tomorrow to look at it, right?

poisdeux:
I added the debug to see if I could find the cause. But travis is unable to handle that amount of logging so that wasn't very useful. It builds fine on Jenkins so I don't think we should worry about travis. (I'm actually disliking him more and more).

I'll squash this so it can be merged... hang on.

poisdeux:
Done.

Issue #235, opened 2016-04-26
Use hardware volume keys from anywhere inside Kore


KindKitty:
Currently the hardware volume change works only when the remote is open.
It would be useful to be able to change volume using hardware keys also from other parts of the Kore app (e.g.: browsing music/shows).

When I'm browsing media in the app and whatever's playing on Kodi gets to a louder part I go for the hardware keys (to change it quickly) and it doesn't work from there. I then need to wait for the phone volume rocker to get out of the way, select 'remote', and press hardware key again.

Is it possible to change this behaviour?
My neighbors would also benefit from this feature too.

Peque:
Duplicate of #157?

Peque:
I created a new issue (#298), requesting a more complete media control while the screen is locked (i.e.: volume and playback controls).

glenntanner3:
I dont think this is a duplicate, those are different behaviours. I do not want the volume to change when my phone is in my pocket and locked; so that should be a togglable feature. While I have tried to adjust the volume while browsing shows when attempting to add more to the playlist.

PR #453, opened 2017-09-20
Use hardware volume keys from anywhere inside Kore

Calebzor:
* Move dispatchKeyEvent callback for volume control to the BaseActivity to allow all activities to receive hardware button events

resolves xbmc/Kore#235

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/BaseActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/BaseActivity.java
@@ -28,20 +28,15 @@ import org.xbmc.kore.utils.UIUtils;
  */
 public abstract class BaseActivity extends AppCompatActivity {
 
-	@Override
-	protected void onCreate(Bundle savedInstanceState) {
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(this);
         setTheme(UIUtils.getThemeResourceId(
                 prefs.getString(Settings.KEY_PREF_THEME, Settings.DEFAULT_PREF_THEME)));
         super.onCreate(savedInstanceState);
-	}
-
-    @Override
-    public void onPause() {
-        super.onPause();
     }
 
-//    @Override
+    //    @Override
 //    public boolean onCreateOptionsMenu(Menu menu) {
 //        getMenuInflater().inflate(R.menu.global, menu);
 //        return super.onCreateOptionsMenu(menu);
@@ -50,8 +45,8 @@ public abstract class BaseActivity extends AppCompatActivity {
 //    @Override
 //    public boolean onOptionsItemSelected(MenuItem item) {
 //        switch (item.getItemId()) {
-//			case R.id.action_settings:
-//				return true;
+//            case R.id.action_settings:
+//                return true;
 //            default:
 //                break;
 //        }
--- a/app/src/main/java/org/xbmc/kore/ui/BaseMediaActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/BaseMediaActivity.java
@@ -28,6 +28,7 @@ import android.support.v7.app.ActionBar;
 import android.support.v7.widget.Toolbar;
 import android.text.TextUtils;
 import android.transition.TransitionInflater;
+import android.view.KeyEvent;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
@@ -48,6 +49,9 @@ import org.xbmc.kore.jsonrpc.type.ListType;
 import org.xbmc.kore.jsonrpc.type.PlayerType;
 import org.xbmc.kore.ui.generic.NavigationDrawerFragment;
 import org.xbmc.kore.ui.sections.remote.RemoteActivity;
+import org.xbmc.kore.ui.volumecontrollers.OnHardwareVolumeKeyPressedCallback;
+import org.xbmc.kore.ui.volumecontrollers.VolumeControllerDialogFragmentListener;
+import org.xbmc.kore.ui.volumecontrollers.VolumeKeyActionHandler;
 import org.xbmc.kore.ui.widgets.MediaProgressIndicator;
 import org.xbmc.kore.ui.widgets.NowPlayingPanel;
 import org.xbmc.kore.ui.widgets.VolumeLevelIndicator;
@@ -63,7 +67,8 @@ public abstract class BaseMediaActivity extends BaseActivity
         implements HostConnectionObserver.ApplicationEventsObserver,
                    HostConnectionObserver.PlayerEventsObserver,
                    NowPlayingPanel.OnPanelButtonsClickListener,
-                   MediaProgressIndicator.OnProgressChangeListener {
+                   MediaProgressIndicator.OnProgressChangeListener,
+                   OnHardwareVolumeKeyPressedCallback {
     private static final String TAG = LogUtils.makeLogTag(BaseMediaActivity.class);
 
     private static final String NAVICON_ISARROW = "navstate";
@@ -79,6 +84,7 @@ public abstract class BaseMediaActivity extends BaseActivity
 
     private HostManager hostManager;
     private HostConnectionObserver hostConnectionObserver;
+    private VolumeKeyActionHandler volumeKeyActionHandler;
 
     private boolean showNowPlayingPanel;
 
@@ -180,7 +186,6 @@ public abstract class BaseMediaActivity extends BaseActivity
                                                .getBoolean(Settings.KEY_PREF_SHOW_NOW_PLAYING_PANEL,
                                                            Settings.DEFAULT_PREF_SHOW_NOW_PLAYING_PANEL);
 
-
         if(showNowPlayingPanel) {
             setupNowPlayingPanel();
         } else {
@@ -203,6 +208,30 @@ public abstract class BaseMediaActivity extends BaseActivity
 
         hostConnectionObserver.unregisterApplicationObserver(this);
         hostConnectionObserver.unregisterPlayerObserver(this);
+    }
+
+    /**
+     * Override hardware volume keys and send to Kodi
+     */
+    @Override
+    public boolean dispatchKeyEvent(KeyEvent event) {
+        if (volumeKeyActionHandler == null) {
+            volumeKeyActionHandler = new VolumeKeyActionHandler(hostManager, this, this);
+        }
+        return volumeKeyActionHandler.handleDispatchKeyEvent(event) || super.dispatchKeyEvent(
+                event);
+    }
+
+    @Override
+    public void onHardwareVolumeKeyPressed() {
+        showVolumeChangeDialog();
+    }
+
+    private void showVolumeChangeDialog() {
+        VolumeControllerDialogFragmentListener volumeControllerDialogFragment =
+                new VolumeControllerDialogFragmentListener();
+        volumeControllerDialogFragment.show(getSupportFragmentManager(),
+                VolumeControllerDialogFragmentListener.class.getName());
     }
 
     public boolean getDrawerIndicatorIsArrow() {
@@ -296,6 +325,7 @@ public abstract class BaseMediaActivity extends BaseActivity
 
     @Override
     public void playerOnStop() {
+        currentActivePlayerId = -1;
         //We delay hiding the panel to prevent hiding the panel when playing
         // the next item in a playlist
         callbackHandler.removeCallbacks(hidePanelRunnable);
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
@@ -49,7 +49,6 @@ import org.xbmc.kore.jsonrpc.method.Player;
 import org.xbmc.kore.jsonrpc.method.Playlist;
 import org.xbmc.kore.jsonrpc.method.System;
 import org.xbmc.kore.jsonrpc.method.VideoLibrary;
-import org.xbmc.kore.jsonrpc.type.GlobalType;
 import org.xbmc.kore.jsonrpc.type.ListType;
 import org.xbmc.kore.jsonrpc.type.PlayerType;
 import org.xbmc.kore.jsonrpc.type.PlaylistType;
@@ -59,6 +58,9 @@ import org.xbmc.kore.ui.generic.NavigationDrawerFragment;
 import org.xbmc.kore.ui.generic.SendTextDialogFragment;
 import org.xbmc.kore.ui.sections.hosts.AddHostActivity;
 import org.xbmc.kore.ui.views.CirclePageIndicator;
+import org.xbmc.kore.ui.volumecontrollers.OnHardwareVolumeKeyPressedCallback;
+import org.xbmc.kore.ui.volumecontrollers.VolumeControllerDialogFragmentListener;
+import org.xbmc.kore.ui.volumecontrollers.VolumeKeyActionHandler;
 import org.xbmc.kore.utils.LogUtils;
 import org.xbmc.kore.utils.TabsAdapter;
 import org.xbmc.kore.utils.UIUtils;
@@ -77,8 +79,8 @@ import butterknife.InjectView;
 public class RemoteActivity extends BaseActivity
         implements HostConnectionObserver.PlayerEventsObserver,
         NowPlayingFragment.NowPlayingListener,
-        SendTextDialogFragment.SendTextDialogListener {
-	private static final String TAG = LogUtils.makeLogTag(RemoteActivity.class);
+        SendTextDialogFragment.SendTextDialogListener, OnHardwareVolumeKeyPressedCallback {
+    private static final String TAG = LogUtils.makeLogTag(RemoteActivity.class);
 
 
     private static final int NOWPLAYING_FRAGMENT_ID = 1;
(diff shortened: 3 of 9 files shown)


app/src/main/java/org/xbmc/kore/ui/BaseActivity.java
```diff
@@ -15,32 +15,88 @@
...
+		hostManager = HostManager.getInstance(this);
+
+		// Check if we have any hosts setup
+		if (hostManager.getHostInfo() == null && !(this instanceof AddHostActivity)) {
+			final Intent intent = new Intent(this, AddHostActivity.class);
+			intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+			intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+			startActivity(intent);
+			finish();
+		}
```
SyncedSynapse: This seems to be useful, but how did you come up with it? Did you ran into a situation where this happened to you? If so can you describe it.

SyncedSynapse: Forget the comment. I see that you migrated this from RemoteActivity. Anyway, the question still stands: did you ran into a situation where that happened (and the remote activity didn't handle it)? Guess that removing all hosts and then going into another activity might do that, but need to check.

SyncedSynapse:
Personally i'm not sure about this: i think that we should intercept the hardware buttons only on limited circunstances, but i'm not entirely opposed to this.

@poisdeux what do you think about extending the use of hardware volume keys everywhere?

app/src/main/java/org/xbmc/kore/ui/BaseActivity.java
Calebzor: Hi!

I'm not quite sure I follow you. What do you mean by: 

> This happened to you?

What is "this"?

The code change here are more of a refactor, all of this was existing code that I've just moved around in order to allow volume keys to control Kodi's volume from anywhere within Kore. The only thing I've added is:

` && !(this instanceof AddHostActivity)` since clearly we'd not want an infinite loop in case of the `AddHostActivity`

#453 requested this change and I'd personally would like to use volume buttons everywhere too.

SyncedSynapse: Ok. This particular code isn't related to the volume control functions, so i wondered why it was moved.

It starts the add host activity if there's no hosts configured, so it's a safeguard. I was asking if you moved it because a specific situation happened to you and moving it solved the issue.

Calebzor: `hostManager` is used inside `dispatchKeyEvent`, this was a way to make it visible for that method.

poisdeux:
I like the idea. However, the current implementation is not enough from a UX perspective. We will need to provide feedback to the user that the volume has actually changed. So IMO it should popup a widget (similar to a Toast) showing a slider showing the current volume level. This should disappear after 2 seconds unless the user interacts with the slider, which should also allow the user to change the volume.

Btw, although previously we decided that the mute status shouldn't be included in BaseActivity:  https://github.com/xbmc/Kore/pull/320#issuecomment-265457147 which eventually resulted in the NowPlayingPanel.
I actually favour this to be available within all of Kore when a host is available. But from a UX point of view I would expect the NowPlayingPanel to be available as well whenever I'm able to change the volume... That would also be nice when switching between hosts that are currently playing media. You would instantly see what is playing on the other host.

SyncedSynapse:
Haven't looked into it, but as @bbastin suggest in https://github.com/xbmc/Kore/issues/456 , it might be possible to plug into  Android's media infrastructure and use the standard slider that appears when the volume buttons are used.

Calebzor:
This sounds like a good idea to me:

>  I would expect the NowPlayingPanel to be available as well whenever I'm able to change the volume

How I imagine is that when you use the volume keys it'd pop up the NowPlayingPanel and move the slider on that according to the hardware button press. From UX poing of view to me this approach makes more sense then adding another slider.

Calebzor:
I've added a bit of logic to not collapse the now playing panel after 2 seconds if a button was clicked in the panel. I think this makes sense, since if the users interacts more with the panel we should not hide it anymore.

poisdeux:
I like that you used the NowPlayingPanel (not because I'm a fan of the panel) as it doesn't require another view that pops in, which doesn't really feel to be a part of Kore.

Just had a quick test and I've found a few issues with the current setup.

1. It's not consistent. There is no feedback on the remote screen. Although this is the current behaviour, as we are now changing it for other activities this is (IMO) bad UX-design. 
2. If nothing is playing on Kodi the NowPlayingPanel appears and also shows the play buttons, progress bar,  etc. which only have meaning if something is playing. We should disable these if nothing is playing.
3. The panel doesn't disappear when volume was changed. It just stays collapsed.
4. From a UI design perspective it doesn't look nice when nothing is playing. A possible solution for might be making the height of the collapsed view 0dp when nothing is playing. 

screenshot_1506405925

I think that for now we should go with the Toast/Dialog at least on the remote screen and if above issues can be fixed use the NowPlayingPanel for the other screens. @SyncedSynapse what do you think?

axelsimon:
Just a quick comment on the situation when nothing is playing: having some form of sound control with visual feedback would be useful. Several times I've wanted to turn the sound down in Kodi before playing something, as the sound output on the TV or the amplifier was up high (and I didn't have the remote nearby, etc.).

In this case, showing nothing on the "now playing" screen means the only way to know how high up the volume is in Kodi before playing is either to just start playing and hope for the best or use the hardware keys (which I do, as it's my favourite way of controlling sound level) but this requires knowing it's possible (yes, it's in the options, but I'm sure a lot of people 1/ never go into the options and 2/ forget even if they used it once).

I'm not really in favour of just having a volume control slider on the "now playing" screen even if nothing is playing, but I think this case is worth mentioning.

More generally, better, more visual sound volume from Kore would be great. Thanks 😃

Calebzor:
@poisdeux 

I've fixed 3) 4) Those were bugs I've introduced, sorry about that.

For 2) and 5) I've made that if the panel was hidden, it goes back to hidden. But I think this is kind of only issue in case of 2) and for 2) This makes me wonder if the now playing panel is at all the correct place to show volume change when nothing is playing. Maybe as you've suggested a general dialog that shows volume control would be a better approach, then that can be re-used on the remote too. I'll look into that.

Calebzor:
@poisdeux 

I've added a dialog to display and control the volume on screens without an existing `VolumeLevelIndicator` widget.

kore_media_activity_volume_dialog
kore_remote_volume_dialog

Do you have any suggestion for improvement? 

I was thinking if these needs a title or not, but since the mute button is there, I don't think it needs a title.

poisdeux:
Looks really good. But I did find some issues:

1. When dragging the volume slider the dialog still disappears after two seconds.
2. When the dialog is shown it does not disappear after a configuration change (device rotation)
3. When the NowPlayingPanel is in expanded state changing the volume using the hardware keys causes it to collapse.
 
BTW: I noticed you are also using the testing facilities introduced in Kore. It's not mandatory, but It would be really nice if you could add instrumentation tests (in a separate PR) to test all expected scenarios. It is really hard to do these manually in a consistent way.

SyncedSynapse:
Thanks for this, i think this is the right way to do it. I have some comments on it, which we should discuss (besides @poisdeux 's findings):

1. Are we sure we don't want to show the dialog if the nowplaying panel is visible or if the user is in the now playing activity? I understand the rationale - if there's a volume control accessible, that's the one that should be used - but i'm not sure i agree with it. I would prefer that the volume control dialog always appeared - it would be a more coherent experience to the user.

2. The dialog could use some visual tweaks, like increased margins and better alignment (the icon should be aligned with the slider), but i can do that afterwards.

Regarding the implementation, i took a look at the code, and the feel i got is that it grew organically, and could use a step back to think about what is essential. I'll do some line comments in the code, but generally i have the following concerns:
1. In this version you don't need to change the BaseActivity, so it should stay as was. This would imply moving the creation of hostManager to the specific activities, instead of relying on it being available on the parent, which i prefer as it is explicit.

2. I don't really like using inheritance for this behaviour. We are slowly building a deep hierarchy (BaseActivity->VolumeControllerActivity->BaseMediaActivity->FinalActivity?), which is complicating things and will become worst in the future, if we keep this up. Can't this be done via aggregation. Just place the behaviour in an object that gets created in the RemoteActivity and BaseMediaActivity (haven't worked out the details). There will be some duplicate code on those 2 activities for handling the volume key presses, but that would be minor imo.

3. The interaction between VolumeController Activity, Handler and Fragment looks more complicated than it needs to be. AFAICT the handler is created in the activity, which gets called back on a key press, then creating the Fragment, which created another Handler, which calls back the Fragment. Haven't thought this through, but seems too convoluted.

4. Some other points i'll point out in line comments.

app/src/main/java/org/xbmc/kore/ui/BaseActivity.java
```diff
@@ -15,33 +15,50 @@
  */
```
SyncedSynapse: Like i said, i think this activity shouldn't be touched, as it doesn't really need to be changed.

app/src/main/java/org/xbmc/kore/ui/BaseActivity.java
```diff
@@ -15,33 +15,50 @@
...
  * Base activity, where common behaviour is implemented
  */
 public abstract class BaseActivity extends AppCompatActivity {
 
-	@Override
-	protected void onCreate(Bundle savedInstanceState) {
+    /**
+     * Host manager singleton
+     */
+    protected HostManager hostManager = null;
```
SyncedSynapse: I think that each specific activity creates (gets) its hostManager, instead of relying on a protected member inherited from the superclass. It's not obvious when reading the code where it comes from.

app/src/main/java/org/xbmc/kore/ui/BaseMediaActivity.java
```diff
@@ -48,6 +48,7 @@
 import org.xbmc.kore.jsonrpc.type.PlayerType;
```
SyncedSynapse: If we ditch the mechanism of opening the now playing panel, and always show the volume control dialog, this activity would also remain (almost) unchanged, which is a plus imo.

SyncedSynapse:
Still regarding point 3. of the implementation, i was just looking at the code and i would say that if you just ditched the VolumeControllerActivity, and moved the creation of the Handler/Fragment to the RemoteActivity and the BaseMediaActivity (on the dispatchKeyEvent), it would be an improvement.
Still not sure about the whole Activity/Handler/Fragment interaction though.

app/src/main/java/org/xbmc/kore/ui/volumecontrollers/VolumeControllerDialogFragment.java
```diff
@@ -0,0 +1,172 @@
...
+import org.xbmc.kore.jsonrpc.method.Application;
+import org.xbmc.kore.ui.widgets.HighlightButton;
+import org.xbmc.kore.ui.widgets.VolumeLevelIndicator;
+import org.xbmc.kore.utils.LogUtils;
+
+import butterknife.ButterKnife;
+import butterknife.InjectView;
+
+public class VolumeControllerDialogFragment extends AppCompatDialogFragment
+        implements HostConnectionObserver.ApplicationEventsObserver,
```
SyncedSynapse: I would appreciate if we could minimize the use of the HostConnectionObserver. 
In this case i'm not sure it is strictly needed. It seems that you use it to get a callback about the volume changed, but that might be done differently. I'm not sure which way is better, so this can be in for now, i need to think a bit more on it.

app/src/main/java/org/xbmc/kore/ui/volumecontrollers/VolumeControllerDialogFragment.java
```diff
@@ -0,0 +1,172 @@
...
+                    //We depend on the listener to correct the mute button state
+                }
+
+                @Override
+                public void onError(int errorCode, String description) {
+                    LogUtils.LOGE(TAG,
+                            "Got an error calling Application.SetMute. Error code: " + errorCode +
+                                    ", description: " + description);
+                }
+            }, new Handler());
```
SyncedSynapse: A new Handler shouldn't be created here. `callBackHandler` should be used.

app/src/main/java/org/xbmc/kore/ui/volumecontrollers/VolumeControllerDialogFragment.java
```diff
@@ -0,0 +1,172 @@
...
+        volumeMuteButton.setOnClickListener(onMuteToggleOnClickListener);
+        volumeMutedIndicatorButton.setOnClickListener(onMuteToggleOnClickListener);
+
+        volumeLevelIndicator.setOnVolumeChangeListener(
+                new VolumeLevelIndicator.OnVolumeChangeListener() {
+                    @Override
+                    public void onVolumeChanged(int volume) {
+                        cancelDismissDialog();
+                        new Application.SetVolume(volume).execute(hostManager.getConnection(),
+                                defaultIntActionCallback, new Handler());
```
SyncedSynapse: A new Handler shouldn't be created here. `callBackHandler` should be used.
PS. I get that you got this code from somewhere else (`NowPlayingFragment`), but it is also wrong there. I'll fix it.

Calebzor:
@poisdeux 1) 2) Good find, I've fixed them.

@SyncedSynapse 
1a) I've removed the volume keys interaction with the now playing panel and instead the dialog shows up. I think this is OK. The dialog disappears quite quick and can be also dismissed, so in my opinion it's not that distracting from what the user was doing with the app. This reduced code complexity too, so that's probably a +1 for the pros column for this approach.

2a) I've only added a top margin for the `VolumeLevelIndicator` so the text is not at the edge of the dialog. To me aligning the base of the slider with the base of the mute icon does not look so good, but for sure you know the app better, maybe that layout fits better with the app, so I'll leave the small visual improvements to you. 

3b) To me it looks like as soon as the dialog is up, the activity stops receiving the key events and only the dialog gets it. So the logic here is that the key events in the activities make the volume controller dialog show up and once it's shown, the dialog handles the key events from then on (till it gets dismissed).

app/src/main/java/org/xbmc/kore/ui/BaseActivity.java
Calebzor: I've made the changes, because you were making a valid point with the inheritance depth growth, but if I get it right the point you are trying to make here is not something that applies here in my opinion.

To me as soon as I knew that `HostManager` is a singleton, I've stopped caring about where it comes from, in fact I'd actually expect it to be held somewhere "deep", like the `BaseActivity`. I guess maybe if you are looking to go away from the singleton pattern in the future for the `HostManager` then I can get behind what you were saying.

poisdeux:
@Calebzor what about my point 3:

   3. When the NowPlayingPanel is in expanded state changing the volume using the hardware keys causes it to collapse.

Maybe it wasn't clear on what I meant. I'll try to explain it differently. 
If I expand the panel state manually and afterwards use the volume keys to change the volume, the panel will change state to collapsed after two seconds. IMO it should stay expanded.

Calebzor:
@poisdeux yes, yes I've got it the first time. ;)

In the last iteration, I've removed the interaction with the `NowPlayingPanel` so the issue you are describing is no longer valid.

SyncedSynapse:
@Calebzor 
Thanks for the changes. I think that the code is simplified now, so that's definitely a +1. Regarding your points:

1. I agree with not showing the dialog in the `NowPlayingFragment`. Makes sense what you say.
2. Keep it as is. I might take a look afterwards
3. Yeah, i figured there should be a reason. Let's just discuss an alternative that might makes things simpler. Don't know if this will work, but what about:
- Removing the `OnHardwareVolumeKeyPressedCallback` interface and `VolumeKeyActionHandler` completly.
- In the `BaseMediaActivity` and `RemoteActivity`, `dispatchKeyEvent` would look something like this:
```
if (shouldInterceptKey(event))
    showVolumeChangeDialog()
```
`shouldInterceptKey` might be a static function somewhere on the `Utils` package (and should also be used in the `VolumeChangedDialog`).
- The code that deals with the volume change in `VolumeKeyActionHandler` should just migrate to `VolumeControllerDialogFragmentListener`.

So, only 1 class needed (the dialog), the activities just show it up, and it handles all the key events. 

This way, we could also ditch the `volumecontrollers/` directory, and keep everything in `ui/`.

Would this work?

Calebzor:
Hi @SyncedSynapse 

I don't think that'd work, here is why:

In the `NowPlayingFragment` then either we'd have to have the volume key control code there too, because `Application.SetVolume(volume).execute(hostManager.getConnection(), null, null);` in `VolumeControllerDialogFragmentListener` would only get called if it was shown. Which is not the case here. Or we'd have to have the volume control dialog show up there too, which we kinda agreed is ugly. Basically this is why `OnHardwareVolumeKeyPressedCallback` interface is needed.

In my opinion while the code might be a bit more complex like this, but I'm generally in favor a bit more complexity over duplicate code. This way in the future if there'd be a new activity / fragment that needs special volume key handling, it'd just be a matter of implementing the `OnHardwareVolumeKeyPressedCallback#onHardwareVolumeKeyPressed()` instead of re-implementing/copy-pasting a key event listener.

SyncedSynapse:
Ok, i'll merge this as is, but i reserve the right to simplify it later :)

I agree that we should avoid duplicating code, but to do that, you don't need to create a new class and a new interface. It's over-engineering for this specific case. Just put the common code in a static function, maybe in the Utils package. If you look there, you'll see a lot of common code. If each function were transformed into a new class and interface it would be a lot more complicated. OO isn't a solution for everything.

In the future, if a new activity/fragment needs volume key handling, it needs to implement the event listener (this is true in all solutions), call the common function to handle it and show the dialog if it wants to.

PR #458, opened 2017-09-26
Unit test ide execution improvements

Calebzor:
I was having trouble with running the unit tests locally. With this solution the unit tests can be run one by one from the IDE on the already checked in test json files.

What I did is: load the test data json files from the file system with standard java instead of through android context in order to allow more convenient execution of unit tests.

I've only tested this on Windows.


Calebzor:
Sorry wrong branch.

Issue #191, opened 2016-01-31
[PVR] problem with channel color option


criniera:
I make use of the PVR IPTV simple client addon (playlist url: http://kodilive.eu/tv) in my XBMC. this addon make use of the color option in the channel name, so that a channel name is like this "[COLOR orange]Rai 1[/COLOR]". In XBMC i see "Rai 1" written in orange, while in the PVR section of kore i see "[COLOR orange]Rai 1[/COLOR]", this makes impossible to read the channel name because is too long.

![screenshot_2016-01-31-00-55-32](https://cloud.githubusercontent.com/assets/16980409/12701964/b8fa9aac-c819-11e5-9069-e8f8d13923e0.png)
![20160131_005438](https://cloud.githubusercontent.com/assets/16980409/12701965/b90a517c-c819-11e5-8459-08d7e5f3e5c0.jpg)

SPDurkee:
I second this. IPTV is much more popular now. Please add support for tags in titles. Thanks!

PR #471, opened 2017-10-17
Implemented simple Kodi markup code applier for the Now Playing fragment

monzee:


SyncedSynapse:
I can merge this as-is, but would you consider making the `format` function a `static` function in `UIUtils`? 

We would avoid creating an object. Not that it makes any difference in here, but this can/should be used in lists (particularly in the PVR channels list), and there it can save something (not much, i agree, but something nevertheless).

PS. If instead of starting at the beginning of the string it started at the first [ (basically what is done in the `else` clause) it would save some cycles on the most common case (not having any markup code).

monzee:
I implemented the suggested changes. Should I add other tags? I don't really know which ones are often seen by the client. These ones are sent by the YouTube plugin (except for `[I]`). From http://kodi.wiki/view/Label_Formatting:
- `[LOWERCASE]` and `[CAPITALIZE]` are trivial to add
- As for `[LIGHT]` I guess this means thinner font? I'm not sure what it means.
- `[COLOR ...]` would be tricky, I don't know what colors could appear there.
- the others I'm not sure if those would even be sent to the client in that form.

SyncedSynapse:
Your call. I can merge this as-is. Other tags can be implemented later.

Though, if you wanted to close #191 you could add the COLOR tag :) My suggestion is to ignore it, just remove it from the string.

monzee:
I think it's ready. I'm going to post a fix for #191 when this is merged.

PR #474, opened 2017-10-20
Apply markup to PVR text views; addresses #191

monzee:
Not actually tested. I thought PVR was something I could just enable. I just looked for `TextView#setText(CharSequence)` calls in the PVR fragments and inserted `UIUtils#applyMarkup(Context, String)`


kib:
An alternative to the skin-breaking overriding of colours everywhere is to strip the colour tags from the titles. This would have my personal preference.

monzee:
It does indeed just strip out the color tags because I'm too lazy to do it properly.

SyncedSynapse:
I can't test the handling of markup code, as my PVR listings are all sane (no markup). They work the same as they did.

Issue #444, opened 2017-09-04
Control Pad Buttons stopped working BETA 2.3.0


ukjohnd:
Kore 2.3.0 on Nexus 5x Android 8.0 
Kodi 17.3 (LibreELEC)

Remote Up-Down-Left-Right- Centre Select buttons in App stopped working 
All other buttons work ok.

SyncedSynapse:
Thanks for the report. Could you please check if enabling/disabling the option "Use EventServer" on the media center configuration (in Kore) makes any difference regarding making the buttons work.

Also, are you able to sideload a previous version of Kore, so we can have a clearer idea on what might be causing that. Specifically, i'm interested in knowing if the nightly version from 14-08-2017 has the same issue. You can get it here:
http://mirrors.kodi.tv/tools/kore/nightlies/

(the second one).

SyncedSynapse:
Thanks for the report. Could you please check if enabling/disabling the option "Use EventServer" on the media center configuration (in Kore) makes any difference regarding making the buttons work.

Also, are you able to sideload a previous version of Kore, so we can have a clearer idea on what might be causing that. Specifically, i'm interested in knowing if the nightly version from 14-08-2017 has the same issue. You can get it here:
http://mirrors.kodi.tv/tools/kore/nightlies/

(the second one).

Kjartansson:
The one from 14.08.2017 is working

poisdeux:
I'm unable to reproduce this issue. @SyncedSynapse are you able to reproduce this?

@Kjartansson what kind of device and which version of Android and Kodi are you using?

Kjartansson:
I am using Android asop version 7.1.2 on Huawei p9 Lite

poisdeux:
@Kjartansson just to be sure, you have the same issue with 2.3.0 as @ukjohnd does

Kjartansson:
Jep same problem

poisdeux:
@SyncedSynapse I think we should revert 3106a5fd6196022710cd983605154727df44e16e to see if that resolves the issue.

MartijnKaijser:
@poisdeux you can always make a test branch and get jenkins to build it for users to test if it solves it

SyncedSynapse:
@poisdeux I'm also unable to reproduce this (Nexus 5x on Android 7), but there are several reports on the Google Play beta feedback, so it's a somewhat widespread issue.

I'll try to take a closer look at that PR today but i agree that a revert will probably need to be done for now.

[...]

PR #477, opened 2017-10-23
Implemented a scalable control pad

poisdeux:
This should hopefully fix issue #383 and fix the issue with the original version from PR #428. The original version did not include a check to see if the event server was disabled by the user. This was probably the reason some users reported [issues](#444) with the control pad not working during the beta release of 2.3.X.

I falsely assumed that checking if the event server connection failed, the variable [eventServerConnection](https://github.com/xbmc/Kore/pull/428/files#diff-ac8a1a6886da87c8fca766e573e68538R563) would be set to null. This is only the case if hostname resolving fails. Thus, if the event server is not available and the user disabled its usage we would still be using it and do not fall back to using the HTTP input events. This has been [fixed](https://github.com/xbmc/Kore/compare/master...poisdeux:feature/scalablecontrolpad?expand=1#diff-ac8a1a6886da87c8fca766e573e68538R238) now.

In general the following changes have been made in this PR:

   * Refactored RemoteFragment and created a compound view for the
     actual remote. I called it ControlPad to make it more clear what
     its main function is.
   * Implemented a custom grid layout (SquareGridLayout) that will
     always be square. When its width and height are both set to
     match_parent, it will take the smallest of the two as the
     actual size.
   * For devices with a smallest width smaller then 360dp the ControlPad
     is sized to the maximum available space. For larger devices we still
     use the old fixed sizes.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteFragment.java
@@ -15,24 +15,14 @@
  */
 package org.xbmc.kore.ui.sections.remote;
 
-import android.annotation.TargetApi;
-import android.content.res.ColorStateList;
-import android.content.res.Resources;
 import android.content.res.TypedArray;
-import android.graphics.BitmapFactory;
-import android.graphics.PorterDuff;
-import android.graphics.PorterDuffColorFilter;
-import android.graphics.drawable.BitmapDrawable;
 import android.os.Bundle;
 import android.os.Handler;
 import android.preference.PreferenceManager;
 import android.support.v4.app.Fragment;
 import android.view.LayoutInflater;
-import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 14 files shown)


SyncedSynapse:
Looks good at first sight. Let me just test a little bit more.

SyncedSynapse:
@poisdeux Just one observation: With this PR we get 2 fragment_remote that look very similar (both in portrait and in landscape). The only differences that i could see (in the portrait version) are the `FrameLayout` that surrounds the top panel on the non-360dp version (which i didn't really understood why it was needed, can't it be removed?) and the `FrameLayout` that surrounds the control pad on the 360dp version.

Can't this be done using the same `fragment_remote`, applying different styles? I mean, keep the necessary `FrameLayout`s and give them a style that gets set to `wrap_content` or a fixed dimension depending on the version.

poisdeux:
Good suggestion. I changed the code to use two new styles for devices with a smallest width equal or bigger then 360dp. One for portrait orientation and one for landscape orientation. That should reduce the duplication considerably.

PR #504, opened 2018-01-24
Refactor/fabbutton

poisdeux:
This should finish #301 to support local playback of movies and tv shows.
I chose not to use any of the existing FAB speed dial implementations as I couldn't find any that was well supported and met my design requirements. The one from @yavski, that we used in #301, seems to be abandoned.

@SyncedSynapse when you merge this please don't squash this as the first commit contains @yytsui's contribution.


SyncedSynapse:
Due to the previous merges, this PR now has some conflicts. @poisdeux can you please take a look. Thks

poisdeux:
Ok, I'll refactor the code as well to include the new pattern using Futures

SyncedSynapse:
Looks good and works fine (on my tests). It's good to go for me, maybe with some visual tweaks.

Some remarks/questions regarding the code:
- In which scenario did you feel the need to add `updateClientCallback`?  I know that, for instance, on a orientation change the callback result is lost, but where specifically did you need it. What about connections through HTTP? AFAICT, `updateClientCallback` only works when on TCP, so is there something that might go wrong when going through HTTP? And - just for some brain storming - if this is important should we consider using the Evenbus for announcing the result.
- I noticed that the FAB speed dial is specific for this use case. No arguments there, no need to have something generic for 1 use case and i don't anticipate needing more than this FAB speed dial, but maybe we should consider making the class name less generic - `FABSpeedDial` is very generic. Maybe something the includes the work `Play` on the name would be more appropriate.

Regarding the UI/UX, and considering the guidelines Google has for the FAB (and particularly its specific usage on some apps - i looked at Google Calendar for a comparison):
- I think the rounded corners on the label aren't consistent with the rest of the UI, and they seem to be non-standard
- The ALL CAPS label is also somewhat strange. I noticed that you applied the standard button style, but all the FAB examples that i've seen are in Mixed Caps.
- Related to the previous, i think that "Play Locally" would be better the "Play on Remote"
- The speed dial opening animation could be a little bit simplified: the bouncing is nice the first time, but it's not standard and i think it gets tiring after a while.
- If possible, when the fab is open a touch outside of it should close it.

Now, a personal request: i think that the local play option should be an option in the settings. Something like "Allow local play" or similar, which could default to true, but, if false, the FAB shouldn't have a speed dial, and should just behave as currently.

And another question: have you tried local playing music? Does it work?

SyncedSynapse:
btw, nicw work introducing `CoordinatorLayout` and the behaviors

SyncedSynapse:
Just one more thing that i forgot: If you take a look at Google's apps (calendar for instance), when FAB "opens", the background becomes dimmer. The FAB is probably changing the transparency of a background `FrameLayout` or similar that covers the entire screen. Is is possible to do the same without much trouble? I can take a look at it later, but you might have an obvious/simple idea for implementing it that i'm not aware.

poisdeux:
Thanks for the feedback.

Regarding your remarks/questions:
1. It's needed to provide the user with feedback when the action is still pending. The current situation is that the callback is never returned if the device configuration changes. The FAB icon will just return to its original leaving the user in the dark regarding the state of the intended action. This is wrong in my opinion and Kore should try its best to provide correct feedback to the user.
And yes, it currently only works when using TCP, I neglected HTTP at first as it takes more time to implement. So for HTTP connections the busy animation will just run indefinitely if Kodi isn't responding.
I totally forgot about the `EventBus` I'll have a look if it can be implemented using that.
2. Making the name more specific would assume it also contains logic to perform the action conforming to its name. The widget is still only meant to enable user interaction and the logic is still contained in the activity or fragment. The widget's design is a **F**loating **A**ction **B**utton with optional **Speed Dial**s. Hence, its name.

Regarding the UX/UI,
1. You'll have to be more specific. Is there a standard radius we should use? I designed the roundness to be consistent with its FAB icon. Making it more clear they belong together. Color alone is not enough (some people might not see the different colors).
2. Indeed it just follows the Android design guidelines regarding buttons. I'm hesitant to follow any other FAB examples. Using all caps is in general used to "raise importance" and increase "distinction with other text". 
3. Good suggestion. That also makes the text more distinctive from "Play on Kodi".
4. I followed android's recommendations regarding motion. See https://material.io/guidelines/motion/material-motion.html#material-motion-why-does-motion-matter
5. As there were issues with that feature with the previous FABSpeedDial version we used I deliberately put that aside to be implemented later. Using the FAB to close as well isn't bad UX in my opinion. It also makes using the dials less error prone. I.e. pressing just outside a dial's boundaries, resulting in the FAB closing and NOT performing the dial's action. That could be confusing/frustrating to the user. Although we do provide user with feedback the action was received by changing the FAB icon.

I think the settings screen is growing a bit out of control. But I think some of our users would prefer to be able to disable it, so I'll add a setting to disable it.

Haven't tried it for music yet. I think it will work on individual songlevel, but it will probably be a bit harder to implement on artist/album level.

Regarding your last comment. I didn't implement this because if the background is faded, in my opinion this indicates it will not respond to user input. In the current setup this is not the case, so I don't think we should do it.

SyncedSynapse:
Some comments on your comments (i agree with most everything, these are just some notes):

Regarding the callbacks:
Note that, even with the current code, the callback is always called. The callback exists, as it is being held by the `HostConnection`, so it gets executed (and the activity also exists. It might have been destroyed - hence the `isAdded` check, but it hasn't been GC'ed, as it is referenced by the callback). We can't access UI elements if it has been destroyed but it exists. So, my thinking was that we just need to send a message to the new activity that was created, and that could be done via the `EventBus`. Something along these lines (docs on https://github.com/greenrobot/EventBus/blob/v2/HOWTO.md , we're still using v2.4):
1. The activity registers/unregisters with the EventBus.
2. The callback (currently `createPlayItemOnKodiCallback`) should be something like:
```Java
    private ApiCallback<String> createPlayItemOnKodiCallback() {
        return new ApiCallback<String>() {
            @Override
            public void onSuccess(String result) {
                EventBust.getDefault().post(new SomeSuccessEvent(result);
            }

            @Override
            public void onError(int errorCode, String description) {
                  EventBust.getDefault().post(new SomeErrorEvent(description);
        };
    }
```
- Then we need the handlers, which should be run on the UI thread:
```Java
 public void onEventMainThread(SomeSuccessEvent event) {
        // Whatever was in the onsuccess callback
 }
 public void onEventMainThread(SomeErrorEvent event) {
        // Whatever was in the onError callback
 }

Note that these will be posted in the UI thread, so everything will work. I think that we don't even need to use `onEventMainThread` because the callback should already be running on the UI thread, even if the activity is destroyed, but it might be possible that Android also changes the UI thread, so using `onEventMainThread` guarantees that it is posted there, and on the new Activity.

This is kind of convoluted, but i think it will work, and if it is used sparingly throughout the code i don't see a problem. If we reach a situation where it is used a lot, we might need to rethink ditching the callbacks and using the EventBus from the HostConnection directly (need to think more about this).

Regarding the UI:
1. I was referring to the labels, not the icon. I don't like the round rectangles, and i don't think they are standard in Android. The material docs don't really dictate a style for labels on FAB speed dials, so we have some latitude here. My suggestion is to copy the style Google uses in Google Calendar: a rounded rectangle but with a small radius (1 or 2 dp), black background, and MixedCaps. Note that these aren't buttons, so the applying the Button style is debatable (hence my dislike for the ALL CAPS).

4. I can take a look at the animation later. I just think that the FAb opening should use the standard easing animation curve (https://material.io/guidelines/motion/duration-easing.html#duration-easing-natural-easing-curves) `FastOutSlowInInterpolator` i believe, instead of bouncing off. No big deal anyway.

5. Ok (regarding this and the dimming of the background). 

And finally, regarding music, yeah, i also thought of that after writing my comment. It would probably work with individual songs but artists/albums would pose some issues, and those are the valuable addition.

SyncedSynapse:
Good to go for me (i might tweak the visual layout of the FAB a little bit later, if you don't mind).

Do you want to add anything more or can i merge it?

poisdeux:
I want to fix the commit history and remove the unit tests from this PR first. I'll notify you when I'm done.

poisdeux:
I'm done. I've also added a setting to allow users to disable the `FABSpeedDial` and revert back to its old behaviour.
I think we should use a separate PR to do the refactoring of the callback code as it touches much more than just the `FABSpeedDial`.

PR #509, opened 2018-02-16
Tweak FAB speed dial design

SyncedSynapse:
Tweak FAB labels colors and align mini buttons with the main FAB button

### DIFF
--- a/app/src/main/res/color/fabspeeddial_label.xml
+++ b/app/src/main/res/color/fabspeeddial_label.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:state_focused="true"
+          android:color="?attr/fabLabelColorFocused" />
+    <item android:state_pressed="true"
+          android:color="?attr/fabLabelColorPressed" />
+    <item android:color="?attr/fabLabelColorNormal" />
+</selector>
--- a/app/src/main/res/drawable/rounded_corners_shape.xml
+++ b/app/src/main/res/drawable/rounded_corners_shape.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <shape xmlns:android="http://schemas.android.com/apk/res/android">
     <solid android:color="#FFFFFF"/>
-    <corners android:radius="6dp"/>
-    <padding android:left="5dp" android:top="5dp" android:right="5dp" android:bottom="5dp" />
+    <corners android:radius="3dp"/>
+    <padding android:left="8dp" android:top="8dp" android:right="8dp" android:bottom="8dp" />
 </shape>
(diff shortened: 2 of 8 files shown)


poisdeux:
To compare your PR with the current setup I created a side-by-side comparison below:

Current master version:
screenshot_1519070975 screenshot_1519071014

Your proposal:
screenshot_1519070223 screenshot_1519070387

I have a few remarks regarding your choices:

1. You centralised the mini FABs with the main FAB. IMO this introduces two issues:
   1. It leaves less space on the left which might cause issues on low-res devices or with longer label text.
   2. IMO it's less balanced. Remember the material design tries to abide to the [rules of physics](https://material.io/guidelines/#introduction-principles). I know that the current way is not physically correct but it does more adhere to the rules of physics where the button and label are balancing on the center of the main FAB icon.
2. I don't understand why you increased the label background size. The label text is much smaller so no need to increase the background size for that. I would think you wanted it to be more aligned with the mini FAB size but it's smaller so that argument doesn't hold. IMO the increase in background size is not an improvement.
3. The colouring of the label's background for the day theme is not consistent with the FAB colouring. Without the animation I would not think they actually belong together.

SyncedSynapse:
The changes i made were to make the FAB more similar to what i've seen in the "wild", particularly from Google.

The FAB design document isn't overly detailed, but if you look at it, the section that mostly applies to this is the "Transitions"->"Speed dial" section:

https://material.io/guidelines/components/buttons-floating-action-button.html#buttons-floating-action-button-transitions

In the example animation we can see that:
1. The mini icons are horizontally aligned with the main FAB;
2. The mini icons don't use the Accent color, they use a neutral color.

I couldn't find any specs for the labels, so we'll need to take a look at some apps. Surprisingly i find that very few apps use a FAB speed dial. Google Calendar, Hangouts, Sheets/Slides/Docs and the official Reddit app were the ones i found. Google's design is consistent in all those apps. Some screenshots:

device-2018-02-20-181912 device-2018-02-20-183130 device-2018-02-20-182429

You'll notice the same as above: centered mini buttons, Google's apps don't use the accent color, Reddit does. None of the labels use the accent color, they use neutral colors, and to my eye, they have more spacing than what was before this PR.

So i've tried to make the FAB more like these examples. In the night theme, i've kept the accent color for the label background because I think that it didn't look very good in white (though this can be changed, but IMO, to use a neutral color, not the accent color). Regarding your objection about belonging together, as you said, the animation takes care of that, and i haven't given up on including a background dim later, which will help with readability.

Summing up: 
1. i think that centering the mini FABs is what material design dictates (both in the docs and pratical examples);
2. The design also points to not using the accent color in the mini FABs. I didn't do this, which can be discussed, but if anything, we should avoid using the accent color here, not adding more of it;
3. I also think that the buttons should be closer, but they're using the `appcompat` spacing so i refrained from changing them as i didn't understood how it would impact pre-lollipop devices.

poisdeux:
1. Agreed.

2. Ah I understand. IMO the dark background for the labels is ugly and only works when fading out the background. But then I still think it doesn't look right, no matter what Google thinks. A good UI also takes the visually impaired into account. This usually results in a better UI even for the users that don't have any vision problems.
So my suggesting is to keep using the accent color until we change the UI to better handle dark and white backgrounds for the labels.

3. OK, hopefully translators will keep the translations condensed.

Further, I think increasing the padding in the label left and right is an improvement. The vertical padding is not and looks odd. The examples you provided don't use that much vertical spacing. So I would prefer if we revert the vertical padding to its original.

PR #512, opened 2018-02-21
Change the source of AlbumGenres table to support Kodi v18

SyncedSynapse:
As detailed in [#13051](https://github.com/xbmc/xbmc/pull/13051), Kodi v18 changes the way album genres are handled, as the genreId info ceases to be available at the album level being only available at the songs level.

This impacts Kore because `GetAlbums` and `GetAlbumDetails` won't return the `genreId` tag, which is used to populate the local AlbumGenres table.

This PR takes that into account, and changes the source of the local AlbumGenres table to be the `genreId` returned at the song level (by `GetSongs`), making it more consistent with the way Kodi will handle things from now on.

@DaveTBlake you might want to take a look at this, to make sure that it is working correctly. Note that Kore has a local database, which has the genre info at the album level (not the song level). I haven't changed this, as it isn't strictly necessary to make it work with the jsonrpc API changes. 

PS. One other suggestion: you might want to remove the `genreid` params from the [json rpc wiki](https://kodi.wiki/view/JSON-RPC_API/v8#Audio.Fields.Album)


app/src/main/java/org/xbmc/kore/host/HostInfo.java
```diff
@@ -55,11 +54,20 @@
 	 */
 	public static final int DEFAULT_EVENT_SERVER_PORT = 9777;
 
-    public static final int DEFAULT_KODI_VERSION_MAJOR = 16;
+	public static final int KODI_V12_FRODO = 12;
```
poisdeux: Why did you include FRODO, HELIX, ISENGARD, and LEIA? They're not used. Is it just for completeness?

DaveTBlake: I'm confused why you use Kodi version at all, isn't the JSON API version what matters?

DaveTBlake:
if Kore wants to collect song genre IDs ands store for the album that is fine. Hard for me to comment on much else since I don't really know the overall design.

Out of date wiki... Well that is for API v8 and we are on v9 for Leia. I hope that something can extract new wiki from the JSON spec, don't fancy manually editing. Meanwhile  I am trying to keep a list of changes for v9 on the forum for JSON users to watch

app/src/main/java/org/xbmc/kore/host/HostInfo.java
MartijnKaijser: i'd agree.
Every Kodi release there's a JSON  version bump done. Either a minor or major version.

SyncedSynapse: @poisdeux Yes, it's for completeness (though i didn't bother to implement the rest of `isXXXorHigher` functions...)

@DaveTBlake @MartijnKaijser 
Kind of. First, even though this PR includes changes to these constants, this particular change doesn't discriminate based on Kodi versions or JSON RPC API versions: It's a change of behaviour (getting the genres from songs instead of albums) to all versions of Kodi.

We currently discriminate behaviour based on Kodi versions on 2 situations:
1. To open the subtitles search, which was an Addon pre-Gotham and a internal window post-Gotham. I'm not sure, but this change in Kodi might not have been reflected in a JSON RPC API bump, as it has nothing to do with it;
2. To display codec info, whose window changed from Kripton onwards *in event server*.

AFAICT, these situations have anything to do with JSON RPC API changes, so we use Kodi's version to change the behaviour.

SyncedSynapse: But i must say that JSON RPC versioning is something that isn't well handled by Kore right now (it's handled on a case by case basis, without any support infrastructure for versioning). We haven't had any problems so far, but we might have to look at it more thoroughly .

SyncedSynapse:
@DaveTBlake Thanks for maintaining that post. It is really helpful.

PR #515, opened 2018-02-23
Prevent the refresh animation from appearing in a silent sync

SyncedSynapse:
When a silent sync is running the refresh animation shouldn't appear and this wasn't being respected in some situations. For instance, when not connected to Kodi, starting from a TV Show details, selecting one episode or season and hitting back, the animation would appear. This PR makes sure that it is only shown if it isn't silent.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/service/library/SyncUtils.java
+++ b/app/src/main/java/org/xbmc/kore/service/library/SyncUtils.java
@@ -457,29 +457,28 @@ public class SyncUtils {
     };
 
     /**
-     * Checks if a running LibrarySyncService is syncing any given SyncType from a specific host
+     * Returns the first {@link SyncItem} of the given type that is currently
+     * syncing in a running {@link LibrarySyncService} for a specific host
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/xbmc/kore/service/library/SyncUtils.java
```diff
@@ -460,26 +460,24 @@ public static void disconnectFromLibrarySyncService(Context context, ServiceConn
...
      * @param hostInfo host to check for sync items currently running or queued
-     * @param syncTypes sync types to check
-     * @return true if any of the given syncTypes is running or queued, false otherwise
+     * @param syncType sync type to check
+     * @return The first {@link SyncItem} of the given syncType if any is running or queued, null otherwise
      */
-    public static boolean isLibrarySyncing(LibrarySyncService service, HostInfo hostInfo, String... syncTypes) {
-        if (service == null || hostInfo == null || syncTypes == null)
-            return false;
+    public static SyncItem isLibrarySyncing(LibrarySyncService service, HostInfo hostInfo, String syncType) {
```
poisdeux: We should rename this method to `getSyncItem` as it is convention that `is*` methods return true or false.

poisdeux:
Looks good, just one minor naming issue I would like to have changed.

Issue #514, opened 2018-02-23
Memory Leaks


Robigus92:
Memory Leak 1

After a double rotation of the MusicActivity there are 2 instances more of:

class org.xbmc.kore.ui.sections.audio.ArtistListFragment @ 0x12cfec00
class org.xbmc.kore.ui.generic.NavigationDrawerFragment @ 0x12d00400
class org.xbmc.kore.ui.sections.audio.AlbumListFragment @ 0x12d0b400

It should not be like this because the rotations return to the same GUI state and the memory should stay in the same state as well. 

----
Memory Leak 2

After a double rotation of the MoviesActivity there are 2 instances more of:

class org.xbmc.kore.ui.generic.NavigationDrawerFragment @ 0x12cfc400
class org.xbmc.kore.ui.sections.video.MovieListFragment @ 0x12d05800

It should not be like this because the rotations return to the same GUI state and the memory should stay in the same state as well. 

---
Memory Leak 3

After a double rotation of the TVShowsActivity there are 2 instances more of:

class org.xbmc.kore.ui.generic.NavigationDrawerFragment @ 0x12cff400
class org.xbmc.kore.ui.sections.video.TVShowListFragment @ 0x12d06800

It should not be like this because the rotations return to the same GUI state and the memory should stay in the same state as well.

poisdeux:
Thx for reporting this. Do note that it is only a memory leak if we keep a reference to the objects which won't allow the garbage collector to clean up. Is that the case? And if so will you fix it or do you expect us to fix it?

Robigus92:
I have developed an automated Memory Leak detection tool which test Activities executing different lifecycle events (rotations, background-foreground, dialog pop-up events) then forcing an ending Garbage Collections. The idea is that the activity lifecycles are well managed there should not be an increase of the used heap memory.

In your activities, rotation events seem to cause an increasing trend of many Fragments usage. 
Is that correct? Old Fragments should not be removed?

poisdeux:
I just ran a test together with a memory profiler and indeed there seems to be a problem with increased heap memory usage. Unfortunately it didn't make clear what is causing the increased usage. Under normal usage patterns this will probably not cause any problems. Depending on the device it requires a couple of thousand rotations before the app actually crashes. 
But it would be good if we fix this nonetheless. Can you provide a more detailed analysis to better pinpoint the cause of the increased heap usage?

SyncedSynapse:
[...]

PR #516, opened 2018-02-25
Fix memory leaks in lists that use CursorAdapter

SyncedSynapse:
Following issue #514, activities that have lists were being leaked, the reason being the internal `Cursors` (which seem to be reused/pooled in Android, so they have a lifecycle separate from activities) were keeping references to the activity's `Adapter`, therefore keeping it from being GC'ed.

The constructor for `CursorAdapter` being used up until now is [this one](https://developer.android.com/reference/android/widget/CursorAdapter.html#CursorAdapter(android.content.Context,%20android.database.Cursor,%20boolean)), which sets the flag `FLAG_REGISTER_CONTENT_OBSERVER`, unnecessary in `CursorAdapters` that will be used with `CursorLoaders` (according to [the docs](https://developer.android.com/reference/android/widget/CursorAdapter.html#FLAG_REGISTER_CONTENT_OBSERVER)).

Using the other constructor doesn't set that flag, so the `Cursor` doesn't hold up the `Adapter` and the enclosing activities.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/audio/AlbumListFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/audio/AlbumListFragment.java
@@ -269,7 +269,7 @@ public class AlbumListFragment extends AbstractCursorListFragment {
         private int artWidth, artHeight;
 
         public AlbumsAdapter(Context context) {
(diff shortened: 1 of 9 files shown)


poisdeux:
Nice one!

Issue #508, opened 2018-02-11
back arrow button not working


carminemassarelli:
if I touch the back arrow button in the upper left corner that appears in sub menù, I can see only the new button with 3 lines but the content does not change. My phone is a samsung s2 with lineageOS 14.1. no google.

MartijnKaijser:
Hi, I'm [GitMate.io](https://gitmate.io)!

It seems you've just enabled the issue triaging. I'm just scraping all issues from your repository and will give you some more information about this in a few minutes or so.

Because of the rate limit we can't scrape all information (including all comments and authors) right now - our system is already set up to scrape this in the next days over which the predictions will become more precise every day.

If you want me to use a different account for triaging your issues, simply create one and log in with it.

**Sit tight!**

SyncedSynapse:
Where is this happening to you (in which section screen)? I'm not able to reproduce this.

carminemassarelli:
It happens in PVR. Please see screenshot before and after.
![before](https://user-images.githubusercontent.com/16013878/36221285-8a6accce-11b5-11e8-925d-db6bd3708632.png)
![after](https://user-images.githubusercontent.com/16013878/36221286-8acd4412-11b5-11e8-8282-1a4ad004c806.png)

SyncedSynapse:
Fixed in #517

csaavedra:
When is this going to hit the Android play store?

SyncedSynapse:
It's still not on the play store. When I get back I'll do a release

On 19 Aug 2018 19:20, "Martijn Brekhof" <notifications@github.com> wrote:

This fix is included in the latest play store release.
Are you saying it still isn't working?

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/xbmc/Kore/issues/508#issuecomment-414145537>, or mute
the thread

poisdeux:
I'm confused. This issue has been fixed by #517 which is of march 12. So AFAICT this should be in the latest release.
Or are we talking about the back button issue that refuses users to exit Kore?

SyncedSynapse:
You're right. Sorry for the misunderstanding I'm on mobile and didn't check
the issue.
This should be fixed on the play store. Back button not exiting isn't yet.

csaavedra:
Uh, strange, it doesn't work for me. I need to go to another section and then back to PVR. I'm using version 2.4.4, which is the latest installed in my device.

carminemassarelli:
[...]

PR #517, opened 2018-03-09
Fix PVR section back navigation

SyncedSynapse:


poisdeux:
This should fix #508?

PR #524, opened 2018-03-29
Upgraded butterknife to 8.8.1

poisdeux:
Refactored MusicListFragment and PVRListFragment to use AbstractTabsFragment
Fixed scrolling in a nested scroll view using espresso
Fixed issue with setting and checking Kodi major version


app/src/main/java/org/xbmc/kore/ui/widgets/VolumeLevelIndicator.java
```diff
@@ -59,10 +61,18 @@ public VolumeLevelIndicator(Context context, AttributeSet attributeSet, int defS
         initializeView(context);
     }
 
+    @Override
+    protected void onDetachedFromWindow() {
+        super.onDetachedFromWindow();
+        unbinder.unbind();
+        onVolumeChangeListener = null;
+        volumeBarTouchTrackerListener = null;
```
SyncedSynapse: Just curious, you're setting the listeners to null now to prevent some leak or just to help the GC? 
Did you ran into some issue?

poisdeux: Indeed to help the GC. A possible leak might be the BaseMediaActivity holding a reference to the widget and the widget holding a reference to the Activity through the listener. Although I'm not sure the problem exists here, I just added it as it doesn't hurt.

PR #527, opened 2018-04-02
Add metadata for fdroid

mueller-ma:
I downloaded the description and screenshots from Google Play and put them into the format that F-Droid can use it.

After merging this, I will prepare a PR to remove the description from the fdroiddata repo.


SyncedSynapse:
Merged this before asking something:

Does the `fastlane` folder needs to be at root? Would it work if it was inside the `tools` folder?
PS. We don't use fastlane so this is just for f-droid.

mueller-ma:
I do not think it would work when it is in the tools folder.
You do not have to use fastlane.

PR #529, opened 2018-04-05
Fixed issue with control pad button tests

poisdeux:
* Fixed tests to determine which Kodi version Kore is connected too.
     This caused the long press tests to fail as Kore didn't see v17 as
     the version that should use the new JSON RPC call.
   * Refactored setting the Kodi version during testing to also set the
     correct version in the database, instead of only setting it in the HostInfo
     instance.


SyncedSynapse:
@poisdeux Same here?

poisdeux:
Fixed

PR #530, opened 2018-04-06
Refactored ApiFuture

poisdeux:
Refactored the ApiFuture setup, introduced by @monzee in #473, to make the code more [SOLID](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design)).

* Moved logic on how to execute an ApiMethod from ApiFuture to HostManager to make
  usage more clear and remove the cyclic dependency between ApiFuture and HostConnection.
  Cyclic dependency was found in ApiFuture.from() using the HostConnection to execute
  the ApiMethod, and the HostConnection containing an execute method that wraps
  the ApiMethod in an ApiFuture.
* Implemented a version of withCurrentHost to support using callbacks to
  simplify handling the result of an ApiFuture in an Activity or Fragment.
* Refactored ApiFuture and OpenSharedUrl to provide more meaningful error messages.
* Refactored Session interface into an HostAction abstract class to clarify
  its usage and be able to handle device configuration changes in Fragments
  as well as Activities.
* Added copyright statements to OpenSharedUrl and ApiFuture.
  Original author forgot to include them. I used the overall XBMC copyright
  as the project in general falls under the XBMC copyright license.
* Refactored RemoteActivity and moved waiting on an ApiMethod result to
  HostManager. RemoteActivity is notified of any result through callbacks
  now. This reduces code duplication.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/MainApp.java
+++ b/app/src/main/java/org/xbmc/kore/MainApp.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2018 Martijn Brekhof. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.xbmc.kore;
+
+import android.app.Application;
+import android.content.Context;
+
+/**
+ * Application class to make the context accessible from non UI threads without
+ * having to pass the context as argument to each method call.
+ * <br/>
+ * Usage: MainApp.getContext()
+ */
+public class MainApp extends Application {
+    private static Context context;
+
+    @Override
+    public void onCreate() {
+        super.onCreate();
+        context = this;
+    }
+    
+    public static Context getContext(){
+        return context;
+    }
+}
--- a/app/src/main/java/org/xbmc/kore/host/HostManager.java
+++ b/app/src/main/java/org/xbmc/kore/host/HostManager.java
@@ -21,15 +21,20 @@ import android.content.SharedPreferences;
 import android.database.Cursor;
 import android.net.Uri;
 import android.os.Handler;
+import android.os.Looper;
 import android.preference.PreferenceManager;
+import android.support.annotation.NonNull;
 import android.text.format.DateUtils;
+import android.util.SparseArray;
 
 import com.squareup.okhttp.OkHttpClient;
 import com.squareup.picasso.OkHttpDownloader;
 import com.squareup.picasso.Picasso;
 
 import org.xbmc.kore.Settings;
+import org.xbmc.kore.host.actions.HostAction;
 import org.xbmc.kore.jsonrpc.ApiCallback;
+import org.xbmc.kore.jsonrpc.ApiException;
 import org.xbmc.kore.jsonrpc.HostConnection;
 import org.xbmc.kore.jsonrpc.method.Application;
 import org.xbmc.kore.jsonrpc.type.ApplicationType;
@@ -39,10 +44,8 @@ import org.xbmc.kore.utils.NetUtils;
 
 import java.io.File;
 import java.util.ArrayList;
-import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
 
 /**
  * Manages XBMC Hosts
@@ -53,20 +56,18 @@ import java.util.concurrent.Future;
 public class HostManager {
 	private static final String TAG = LogUtils.makeLogTag(HostManager.class);
 
-    /**
-     * A block of code that is run in the background thread and receives a
-     * reference to the current host.
-     *
-     * @see #withCurrentHost(Session)
-     */
-    public interface Session<T> {
-        T using(HostConnection host) throws Exception;
+    public interface OnActionListener<T> {
+        void onSuccess(T result);
+        void onError(ApiException e);
     }
+
+    private final SparseArray<OnActionListener> actionListeners = new SparseArray<>();
+    private final Handler handler = new Handler(Looper.getMainLooper());
 
     /**
      * Provides the thread where all sessions are run.
      */
-    private static final ExecutorService SESSIONS = Executors.newSingleThreadExecutor();
+    private static final ExecutorService SINGLE_THREAD_EXECUTOR = Executors.newSingleThreadExecutor();
 
 	// Singleton instance
 	private static volatile HostManager instance = null;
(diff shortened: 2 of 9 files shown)


SyncedSynapse:
I'll need some time to review this, but i'm not sure about `MainApp`.

If we don't want to pass `Contexts`, why not use `Context.getApplicationContext` instead: https://developer.android.com/reference/android/content/Context.html#getApplicationContext()

No need for `MainApp` to access that context.

poisdeux:
I might be missing something, but AFAIK getApplicationContext() isn't static so we need the Context object to get to the application context.

SyncedSynapse:
I was missing something. Don't know what led me to believe it was static, but it's not.

poisdeux:
I never saw the Future design as a way to get rid of the callback design. In fact I think that the current Future design will actually result in a much worse situation, where each activity will require more complex logic which even has a strict requirement, running on a separate thread.
The complexity of the Future design became apparent to me when I implemented the playlist JSON RPC and I continuously had to look at the Future/Session implementation to understand how to do it. The reason might be my limited IQ, but as I have a lot less problems understanding the observer/callback design, I figured the code is just too complex.

Another thing I really think we should avoid is adding logic inside classes that should not be there. IMO the HostManager and HostConnection should be the ones waiting on the result and not the activity.

I'm not familiar with the `ContinuationStage` and `CompletableFuture` so I can't comment on that (yet). I'll look into it.

I agree that it might actually be better to remove it completely. What I liked is that it provided a way to group multiple JSON RPC requests (Session) and move data required for a JSON RPC request out of the Activity into a dedicated object (Future). But this can also be done without using the Future setup and all the logic required to get the result.

poisdeux:
Can't find anything about `ContinuationStage`, did you perhaps mean `CompletionStage`?

SyncedSynapse:
Yes, that's it, sorry. And yes, if we want to avoid callback hell, using Java's native patterns seems the way to go for me. Not sure that we need them though.

So this PR refactors the current design to make it more organized, whereas i believe that a more radical redesign might be better, which probably makes this redundant. I can take a try at it later, is that ok by you?

poisdeux:
Yes, that's OK. I'll close this then.

SyncedSynapse:
Reopening this to continue the discussion.

So, i've taken a stab at it, and it turns out that there aren't that many things that i'd like to change in the current code, nothing radical at all...
The changes that i've made are here: https://github.com/SyncedSynapse/Kore/commit/b8e0bb0c7e163d4f07c5fd01e2d67c91b35658c8

But first, we need to agree on what should be changed. I'll state my views so we can discuss point by point:
* `Future`/`ApiFuture`
IMO the current `ApiFuture` doesn't need to change. You argue that there's a circular dependency between `ApiFuture` and `HostConnection` and indeed there is, but i don't see that as a problem (more on this later). There's an intrinsic connection between them, in the sense that the `Future` correct behaviour depends on it being changed in the callback to `execute`, otherwise the future is incomplete. That should be explicit *and* self-contained, and the current implementation makes it clear that you're only supposed to instantiate an `ApiFuture` (and receive a `Future`) through `HostConnection.execute`.
As you've done, breaking the dependency implies that `ApiFuture` can be instantiated anywhere (public constructor), and that doesn't guarantee it's correct behaviour. It isn't self-contained (also, `Future` semantics imply using exceptions when there's an error, so that should be kept).
In fact, i would even argue that `ApiFuture` shouldn't be an explicit class at all. It should be an anonymous inner class inside `HostConnection`, returned by `execute` as a simple `Future`. Clients only need to know that it is a `Future` and they should call the `get` method to get the result. I haven't done this because it's more code on `HostConnection`, but i'm tempted, as it's what makes sense conceptually.

* `Session`/`withCurrentHost`
As i've said in the original PR, these are just a thin layer on `Callable`, and IMO they just obfuscate the code. I've removed them, and explicitly submit `Callables` on an `Executor`, to make it clear that we're just executing something on a background thread (so we can liberally call the future's `get` method).
Note that these are not a scheme to group future calls together. That's the impression the name `Session` gives, but there was no behaviour to facilitate that.
To me, removing this cleared the main source of confusion.

Now for the other stuff that this PR introduces:
* Callbacks
I don't think we should introduce callbacks here. Using futures is a way to avoid callbacks, otherwise, why should we use them? The thing is, this particular case (intent handling) maybe isn't the best scenario to use futures, because we need to interact with the UI when it's done. For instance in the `LibrarySyncService`, using futures would simplify the code (though i don't intend to change it right now). Nevertheless, i thing that, even though the `pendingShare`/`awaitingShare` scheme is somewhat contrived, it is not overly so.

I didn't understand this part of the comment:

> Refactored Session interface into an HostAction abstract class to clarify
its usage and be able to handle device configuration changes in Fragments
as well as Activities.

Namely the configuration changes bit. Can you point to where's the issue?

poisdeux:
* `Future / ApiFuture`
I agree that the `ApiFuture` is actually tightly coupled with `HostConnection` and should be moved to `HostConnection`. That would be a much better solution to fix the circular dependency, if we disallow usage of `ApiFuture` outside of `HostConnection`.
* `Session / withCurrentHost`
This makes it more clear indeed.
I thought `Session` was used to group multiple `ApiMethod`'s, not `Future`'s. As I see it, a single `ApiMethod` is executed using `HostConnection`, if multiple `ApiMethod`'s need to be chained, wrap these in a `Session` and have this executed using `HostManager`.
* Callbacks
IMO the `pendingShare/awaitingShare` scheme is much more error prone and will result in more code duplication inside the activities and fragments.
What I don't like is that we need to handle the result in the activity on a background thread ([awaitShare](https://github.com/xbmc/Kore/blob/master/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java#L390)). This is easily forgotten.
IMO activities should be kept as simple as possible and only manage the UI thread. I think we should have `HostManager` manage the execution on a background thread. 
How about using the `EventBus` to replace the callback infrastructure?

Regarding the configuration change handling. In the `Activity` you can pass an `Object` to the new instance. In a `Fragment` you are only allowed to pass a `Bundle` which makes it impossible to use the same scheme to handle configuration changes on fragment level. I used `HostManager` to keep a list of pending `ApiMethod` calls and have a `Fragment` reconnect itself as listener after a configuration change. But I think we should reconsider this completely and see if we can use the `EventBus` for this which I think should simplify the code considerably.

SyncedSynapse:
I've created a PR with this and the `ApiFuture` redesign. I actually didn't put it into `HostConnection`. On a second though i went with something similar to your redesign, as it is generic so can be used in other situations (though dangerous if you call `get` without completing the future). I just changed the complete methods to be similar to the ones present in Java's `CompletableFuture` - looks more standard.

Regarding the share i haven't changed it because it's not bothering me much now. 

Just one note regarding `HostManager`: notice that it doesn't manage any threads nor connections. It's a singleton to get the hostconnection/picasso, and a stub to interact with the database. Nothing more, so let's be careful adding more responsibilities to it.

SyncedSynapse:
I'm not sure i understand what you mean, but if you want to refactor that part, my suggestion is to use the `EventBus` ( as you've already suggested this somewhere):
- In `OpenSharedUrl`, at the end of `call`, post the result (a boolean or exception) to the bus;
- On the activity, implement `OnEventMainThread` and catch the result, doing what is currently in `awaitShare`.

(this needs to be fleshed out, in this scenario we won't use `Futures` at all, so maybe `OpenSharedUrl` can just be a `Runnable` instead of a `Callable`).

As for the Executor, what's the issue with it being on the activity? It just launches a background thread, which IMO is not conceptually wrong to do on the UI. And might i point out that, conceptually, it's really not at the UI, as it is the result of handling an intent, so, that part could be refactored to a new activity that doesn't have an UI, just handles the intent (and redirects to the remote afterwards).

poisdeux:
I didn't mean refactor the new Future/Executor setup but refactor my code that still uses the design from this PR.

IMO, the problem with the executor on the activity, is that it leads to more code duplication and boilerplate code when applied in other activities as well. The `EventBus` pattern would be better but still requires the developer to be aware that the result is published using the `EventBus`. 
That's why I prefer the callback pattern in situations where a method, that returns a result asynchronously, is used in multiple places. As the callback implements an interface the compiler is able to detect incorrect usage and the IDE can generate a template implementation which only needs to be filled in.

What I like about the `Future` pattern is that it replaces the hard to follow chaining of callbacks with something that can more easily be read. So I think we should still use the `Future`s in `OpenSharedUrl` but should consider another way to return the end result.

PR #536, opened 2018-04-28
Issue/fixinstrumentationtests

poisdeux:
During tests, activities were not closed properly causing increasing memory usage, which resulted in failing tests because of out of memory errors. This has been fixed by using the test orchestrator which runs each test in its own instrumentation instance.

To use the test orchestrator we needed to upgrade the build tools which caused 
the PagerSlidingTabStrip from astuetz to fail. As the project seems dead, I replaced it with
PagerTabStrip from the support library. This does use a different layout for the pager so
we probably need to discuss how to incorporate the PagerTabStrip or use an alternative.


SyncedSynapse:
Regarding the PagerTabStrip, it's fine by me. I think we should use All caps for the tab headings, as that seems to be Android's guideline, buy i can change that later.

This PR changes the build tools and gradle version, so i'm not sure it will build cleanly on jenkins/travis. Paging @Memphiz to check if it's ok to go ahead and merge this, and warn him that we might need his help if the build breaks.

Memphiz:
@SyncedSynapse well it still builds on Jenkins - but is there a chance that the changes influence the output directory of the apk file? I tried a test build (without signing and upload) and it failed because the apk was not in the directory were it is supposed to be. Maybe this is a bug on Jenkins because I never tried to do a build without signing and upload - or it’s a change of behaviour due to the changes in this PR.

poisdeux:
@Memphiz the link points to a non-existing page.
And sorry about the jenkins build failure. I only fixed the travis-ci build but forgot about the jenkins setup. I'll fix the scripts and update this PR in a moment.

BTW: does the jenkins server support running docker images? That would allow us to update the android build environment ourselves whenever needed.

Memphiz:
Link doesn’t work for you because you need to be logged in which is only possible for team members.

Also there is no dicker support. Would be a bit difficult aswell because this nose holds the signing infrastructure that needs to be kept private.

Beside that - Jenkins is green now.

PR #537, opened 2018-05-03
Add support for handling `Player.OnResume` event through TCP

SyncedSynapse:
As detailed [here](https://github.com/xbmc/xbmc/pull/13726#issuecomment-378188082), Kodi will stop sending the event `Player.OnPlay` when resuming a video, sending a `Player.OnResume` event instead. This PR handles that change.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
+++ b/app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
@@ -384,14 +384,17 @@ public class HostConnectionObserver
         chainCallGetActivePlayers();
     }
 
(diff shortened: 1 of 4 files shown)


poisdeux:
Shouldn't the activities that listen for Player events be updated as well, to handle the `Player.OnResume` event?

SyncedSynapse:
The activities don't listen to `HostConnection.PlayerNotificationsObserver` events (which is where the OnResume was added), they listen to `HostConnectionObserver.PlayerEventsObserver`.

The "mapping" is done in `HostConnectionObserver`

poisdeux:
Ah I missed the `chainCallGetActivePlayers` call.

PR #538, opened 2018-05-04
Fixed issue using drawableRight with vector drawables

poisdeux:
Stumbled upon this when running Kore on a pre-lollipop device. Kore crashed due to the drawableRight attribute that wasn't able to find the vector drawable.
AFAICT the drawableRight wasn't used and could simply be removed. Strange thing that the left button already had the drawableLeft commented out.

### DIFF
--- a/app/src/main/res/layout-land/fragment_remote.xml
+++ b/app/src/main/res/layout-land/fragment_remote.xml
@@ -36,7 +36,7 @@
             android:layout_weight="1"
             android:layout_width="match_parent"
             style="@style/Widget.Button.Borderless"
-            android:src="?attr/iconHome"
+            app:srcCompat="?attr/iconHome"
             android:tint="?attr/defaultButtonColorFilter"
             android:contentDescription="@string/home"/>
         <ImageButton
@@ -45,7 +45,7 @@
             android:layout_weight="1"
             android:layout_width="match_parent"
             style="@style/Widget.Button.Borderless"
-            android:src="?attr/iconMovies"
+            app:srcCompat="?attr/iconMovies"
             android:tint="?attr/defaultButtonColorFilter"
             android:contentDescription="@string/movies"/>
         <ImageButton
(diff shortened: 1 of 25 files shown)


SyncedSynapse:
The drawableRight is being used, AFAICT. The drawable left isn't used because that button isn't really a "previous" button, it changes between "cancel/search/previous" so it doesn't have a drawable, but the right button is always "next" so it gets a drawable.

The issue might be something else. drawableRight is supported from api 3, so it should work. If it isn't finding the vector drawable, it might have other issues later on other parts of the UI. I'm going to revice my Nexus to test this out and report back.

poisdeux:
I updated this PR, replacing all `android:src` attributes with `android:srcCompat`

SyncedSynapse:
So let's recap:

1. We were using the default behaviour, which was to use vector drawables on v21+ and auto generated pngs on < v21;
2. Adding `vectorDrawables.useSupportLibrary = true` changes this to use vector drawables on all versions, which needs the use of appcompat and hence your changes to use `srcCompact`;
3. This change doesn't support vector drawables on `TextViews` so the `drawableRight` had to go;
4. Using vector drawables on all versions saves about 150 Kb on the release apk

I'm not a fan of having to use `srcCompat` if we don't have to. Nothing wrong with it, but it's easy to forget/miss and lead to an error that won't be spotted easily. Furthermore, i would like to keep the `drawableRight` on the wizard if possible. 
Having to accept this to save 150 Kb is not a tradeoff that i would make, so IMO we should keep the previous way of handling drawables. This implies:
1. Getting rid of the `vectorDrawables.useSupportLibrary = true`
2. After that we get a new error, complaining about the launcher icon. This wasn't being reported before, and it's not really an error, but to make it go away we need to move `ic_launcher_foreground.xml` to a new `drawable-v21` directory.

I'll make a PR with this changes so we can have a discussion base.

poisdeux:
I agree. I'll test your PR and merge it if it works.

poisdeux:
Closing as superseded by #539

PR #566, opened 2018-07-20
Fix back button not exiting the app when on the remote activity

SyncedSynapse:
Hitting the back button in the remote activity was being ignored, thus preventing exiting the app. This issue was inadvertidely introduced in d11dbe68531f8e9f17cbbead11fa4a879eb89ee4


poisdeux:
Oops...

Issue #430, opened 2017-08-04
Share to alternate Kodi instances


Cj-Malone:
Currently "Play on Kodi" sends it to the currently selected host, but sometimes I don't want to send it to the main one.

With Android 6.0 (23) you can provide ChooserTargets, so when you share a link each Kodi instance could be in the share menu, by selecting one that's not the current Kore could switch to it and then play the link.

[Docs](https://developer.android.com/reference/android/service/chooser/ChooserTargetService.html) + [Sample Code](https://developer.android.com/samples/DirectShare/index.html)

SyncedSynapse:
Implemented in #572

PR #572, opened 2018-08-27
Use a ChooserTargetService to direct a shared url to a specific host

invisiblek:
Addresses #430 

NOTES:
- Screenshot of what it looks like: https://imgur.com/a/DFWENEn
- I have not checked if this breaks pre-M devices as I do not have any up and running at the moment. (this functionality was added in M)
- The official YouTube app sucks because it doesn't appear to use the standard android share dialog.
- I put my name on the copyright for the new java file, I am unsure if this is proper, but a quick look at other files appeared to be the convention in this project. Let me know if that is not the case.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/service/HostChooserTargetService.java
+++ b/app/src/main/java/org/xbmc/kore/service/HostChooserTargetService.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2018 Dan Pasanen. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.xbmc.kore.service;
+
+import android.annotation.TargetApi;
+import android.content.ComponentName;
+import android.content.IntentFilter;
+import android.graphics.drawable.Icon;
+import android.os.Build;
+import android.os.Bundle;
+import android.service.chooser.ChooserTarget;
+import android.service.chooser.ChooserTargetService;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.xbmc.kore.host.HostInfo;
+import org.xbmc.kore.host.HostManager;
+import org.xbmc.kore.R;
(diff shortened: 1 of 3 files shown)


invisiblek:
I have no idea what is up with that gradle error :/

poisdeux:
You can use emulators to test it on pre-Marshmallow versions of Android.

Regarding the travis-ci errors, at first it was travis' unable to download all the dependencies. I started a rebuild and now the lint check is failing. See the [details](https://travis-ci.org/xbmc/Kore/builds/421261176?utm_source=github_status&utm_medium=notification) for more info. You need to add [`@TargetApi`](https://developer.android.com/reference/android/annotation/TargetApi) annotations and add a check to see if the code is executed on a supported Android version.

app/src/main/java/org/xbmc/kore/service/HostChooserTargetService.java
```diff
@@ -0,0 +1,56 @@
...
+        HostManager hostManager = HostManager.getInstance(this);
+
+        final Icon icon = Icon.createWithResource(this, R.mipmap.ic_launcher);
+        final float score = 0;
+        final ComponentName componentName = new ComponentName(getPackageName(), "org.xbmc.kore.ui.sections.remote.RemoteActivity");
+
+        for (HostInfo host : hostManager.getHosts()) {
+            String title = host.getName();
+            Bundle intentExtras = new Bundle();
+            intentExtras.putString("host.name", title);
```
SyncedSynapse: Please use host.id instead of name which is not guaranteed to be unique

app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
```diff
@@ -386,6 +387,18 @@ private void handleStartIntent(Intent intent) {
             videoUrl = videoUri.toString();
         }
 
+        // If a host was passed from the intent use it
+        String host_name = intent.getStringExtra("host.name");
+        if (host_name != null) {
+            HostManager hostManager = HostManager.getInstance(this);
+            for (HostInfo host : hostManager.getHosts()) {
+                if (host.getName().equals(host_name)) {
```
SyncedSynapse: Please use host.id instead of name

app/src/main/java/org/xbmc/kore/service/HostChooserTargetService.java
invisiblek: Ah, good point. Will amend shortly (been a busy few days).

invisiblek:
Tested sharing from a lollipop emulator and although sharing from a youtube page didn't work (I didn't dig in to why but this appears to have been broken before my commit), vimeo videos shared just fine.

SyncedSynapse:
Good for me. @poisdeux do you want to add anything?

poisdeux:
Nope, fine by me.

PR #573, opened 2018-09-01
Instrumentedtests/recyclerview

poisdeux:
* Refactored the instrumentation tests to support RecyclerViews.
* Fixed an issue with restoring the search query state.
* Refactored AddonsHandler so we can now directly start the AddonsActivity in tests, instead of starting MoviesActivity and from there start the AddonsActivity.
* Upgraded test libraries, which also required us to upgrade to support lib version 27.1.1


SyncedSynapse:
@poisdeux I think the upgrade to 27 should be in a separate PR. In fact i think every upgrade (android, gradle, whatever) should have its own PR so that we can identify if something goes wrong due to it.

poisdeux:
@SyncedSynapse I removed the upgrade. Tests run fine on the minSDK (API 17) and targetSDK (API 26). Didn't run the tests on SDK levels inbetween.

PR #575, opened 2018-09-10
Addon list improvements

tomerf:
I added an icon to indicate which addons are disabled in the list view and also added options to choose whether to show disabled addons or the icon.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/Settings.java
+++ b/app/src/main/java/org/xbmc/kore/Settings.java
@@ -136,10 +136,6 @@ public class Settings {
     public static final String KEY_PREF_ADDONS_FILTER_HIDE_DISABLED = "addons_filter_hide_disabled";
     public static final boolean DEFAULT_PREF_ADDONS_FILTER_HIDE_DISABLED = false;
 
-    // Show disabled status on addon list
-    public static final String KEY_PREF_ADDONS_SHOW_DISABLED_STATUS = "addons_show_disabled_status";
-    public static final boolean DEFAULT_PREF_ADDONS_SHOW_DISABLED_STATUS = true;
-
     // Use hardware volume keys to control volume
     public static final String KEY_PREF_USE_HARDWARE_VOLUME_KEYS = "pref_use_hardware_volume_keys";
     public static final boolean DEFAULT_PREF_USE_HARDWARE_VOLUME_KEYS = true;
--- a/app/src/main/java/org/xbmc/kore/Settings.java
+++ b/app/src/main/java/org/xbmc/kore/Settings.java
@@ -132,6 +132,14 @@ public class Settings {
     public static final String KEY_PREF_TVSHOWS_SHOW_WATCHED_STATUS = "tvshows_show_watched_status";
     public static final boolean DEFAULT_PREF_TVSHOWS_SHOW_WATCHED_STATUS = true;
 
+    // Filter disabled addons on addons list
+    public static final String KEY_PREF_ADDONS_FILTER_HIDE_DISABLED = "addons_filter_hide_disabled";
+    public static final boolean DEFAULT_PREF_ADDONS_FILTER_HIDE_DISABLED = false;
+
+    // Show disabled status on addon list
+    public static final String KEY_PREF_ADDONS_SHOW_DISABLED_STATUS = "addons_show_disabled_status";
+    public static final boolean DEFAULT_PREF_ADDONS_SHOW_DISABLED_STATUS = true;
+
     // Use hardware volume keys to control volume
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/xbmc/kore/Settings.java
```diff
@@ -132,6 +132,14 @@
     public static final String KEY_PREF_TVSHOWS_SHOW_WATCHED_STATUS = "tvshows_show_watched_status";
     public static final boolean DEFAULT_PREF_TVSHOWS_SHOW_WATCHED_STATUS = true;
 
+    // Filter disabled addons on addons list
+    public static final String KEY_PREF_ADDONS_FILTER_HIDE_DISABLED = "addons_filter_hide_disabled";
+    public static final boolean DEFAULT_PREF_ADDONS_FILTER_HIDE_DISABLED = false;
+
+    // Show disabled status on addon list
+    public static final String KEY_PREF_ADDONS_SHOW_DISABLED_STATUS = "addons_show_disabled_status";
```
poisdeux: I do not think we need this as an option. Showing the disabled state should always be on IMO.

app/src/main/java/org/xbmc/kore/ui/sections/addon/AddonListFragment.java
```diff
@@ -156,13 +211,14 @@ public void onSuccess(List<AddonType.Details> result) {
...
                     adapter.clear();
                     for (AddonType.Details addon : result) {
-                        if (addon.type.equals(AddonType.Types.UNKNOWN) ||
-                            addon.type.equals(AddonType.Types.XBMC_PYTHON_PLUGINSOURCE) ||
-                            addon.type.equals(AddonType.Types.XBMC_PYTHON_SCRIPT) ||
-                            addon.type.equals(AddonType.Types.XBMC_ADDON_AUDIO) ||
-                            addon.type.equals(AddonType.Types.XBMC_ADDON_EXECUTABLE) ||
-                            addon.type.equals(AddonType.Types.XBMC_ADDON_VIDEO) ||
-                            addon.type.equals(AddonType.Types.XBMC_ADDON_IMAGE)) {
+                        if ((addon.type.equals(AddonType.Types.UNKNOWN) ||
```
poisdeux: I know this is not your code. But I really think this code smells. As you touched it :). Can you make it smell less?. For instance refactoring it, by replacing the conditions with a method call `isAddonSupported(addon)`?

app/src/main/res/layout/grid_item_addon.xml
```diff
@@ -21,34 +21,49 @@
...
                 android:layout_height="wrap_content"
+                android:layout_alignLeft="@id/title"
+                android:layout_alignStart="@id/title"
+                android:layout_below="@id/title"
                 style="@style/TextAppearance.Medialist.Details"
                 android:paddingBottom="0dp"/>
-        </LinearLayout>
-    </LinearLayout>
+
+            <ImageView
```
poisdeux: Instead of showing a cancel icon, why not grey-out the whole item? That would make it more consistent with how Android in general displays disabled UI elements.

app/src/main/java/org/xbmc/kore/Settings.java
tomerf: Alright, I agree but tried to follow the movies convention (show watch status).

app/src/main/res/layout/grid_item_addon.xml
tomerf: I'll see what I can do :)

tomerf: I'm not sure this is the best idea, the subtext is already grey (night theme) and you can still click and go to the addon details screen.
Making it look disabled would cause the users to think it cannot be clicked.

tomerf:
@poisdeux What do you think about the disabled icon?
Everything else is ready.

poisdeux:
@tomerf Hopefully I have some time tonight to look at your changes.

app/src/main/res/layout/grid_item_addon.xml
poisdeux: Ah yes, I see. Why not show the text "disabled" instead?
I think the icon is confusing. I feel the same about the checkmark icon shown on the Addon detail page. @SyncedSynapse do you have any ideas?

poisdeux:
Code looks good. Apart from the icon it's good to go.

app/src/main/res/layout/grid_item_addon.xml
SyncedSynapse: Not really. I can't think of a proper icon for enable/disable but i actually think that the cross icon on this PR is better than the checkmark i used on the details page.

PR #576, opened 2018-09-10
Support local play when Kodi requires authentication

tomerf:
Added credentials to URL when using local play.
Without it trying to play videos on the local device would fail because the external player couldn't connect to Kodi if it is configured to require a password.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/utils/FileDownloadHelper.java
+++ b/app/src/main/java/org/xbmc/kore/utils/FileDownloadHelper.java
@@ -108,7 +108,9 @@ public class FileDownloadHelper {
 
         public String getMediaUrl(HostInfo hostInfo) {
             String pathforUrl = fileName.replaceAll(" ", "%20").replaceAll("'","%27");
-            String videoUrl = String.format("http://%s:%d/vfs/%s", hostInfo.getAddress(),
+            String credentials = (hostInfo.getPassword() == null || hostInfo.getPassword().isEmpty()) ? "" :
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/xbmc/kore/utils/FileDownloadHelper.java
```diff
@@ -108,7 +108,9 @@ public String getDownloadDescrition(Context context) {
 
         public String getMediaUrl(HostInfo hostInfo) {
             String pathforUrl = fileName.replaceAll(" ", "%20").replaceAll("'","%27");
-            String videoUrl = String.format("http://%s:%d/vfs/%s", hostInfo.getAddress(),
+            String credentials = hostInfo.getPassword().isEmpty() ? "" :
```
poisdeux: Generates a NPE when no password is set.

app/src/main/java/org/xbmc/kore/utils/FileDownloadHelper.java
```diff
@@ -108,7 +108,9 @@ public String getDownloadDescrition(Context context) {
...
+                String.format("%s:%s@", hostInfo.getUsername(), hostInfo.getPassword());
```
poisdeux: Same here.

tomerf:
Fixed.

tomerf:
@poisdeux Merge?

poisdeux:
Didn't have time to test this until now. Looks good and works great. Thanks.

PR #581, opened 2018-09-24
Updated Dutch translations

ThrowsThrowable:
Updated Dutch translations and modified some to be consistent with Kodi translations


poisdeux:
Thanks. You added some great translations. There's only one translation I don't agree with and that's changing *film* to *speelfilm*. IMO it doesn't add anything but characters. And that might become an issue when there isn't enough room in the UI to show "speelfilm". It might get abbreviated to "Speel..." which makes it confusing as it has a total different meaning.

ThrowsThrowable:
Agree, I changed it back to "Films".
What do you think of changing "TV-programma's" to "Series". It's shorter, more correct (many of the series are not aired on TV anymore) and it can't be confused with the PVR section.

Issue #578, opened 2018-09-19
Watched check mark synchronisation


ThrowsThrowable:
When the "watched" check mark of a movie or episode is toggled in kore, the checkmark in the listview is mostly not synchronized. In this case, manually synchronising the list view by swiping down is required. Once in a while it seems to get updated automatically though.

Best way to reproduce this is to navigatie back immediately after toggling the check mark.

SyncedSynapse:
Yes this happens. It's the result of trying to minimize the calls to Kodi to sync the database, which has this side effect. It's a compromise.
Note that the info will be updated. If you navigate to other section and back, it will show the correct status, it's just the navigate back that doesn't immediately show the updated status.
It can be a nuisance when you are marking movies/shows as watched, but this is an operation that imo isn't very common, so i'm not sure it's worth the added overhead of making sure it's always synced, given that the vast majority of time that will be unnecessary.

ThrowsThrowable:
Ok I see, so it's an expected behaviour. No problem than, and marking just one episode / movie isn't common indeed.

I found out when I was actually trying to mark full seasons as unwatched. This is very useful for shows for toddlers with episodes of only five minutes. Therefore I was thinking about three dots menu in each season item of a show (Mark as watched / unwatched).

I would like to contribute to this project and help to develop new features but I'm not sure where to start. Do you prefer tickets here on github or on the kodi.tv forum? Is there something as a "where to start for developers" or overview of the code or should we just dig into the code? Finally, thanks a lot for this app!

SyncedSynapse:
It's a known behaviour, not necessarily optimal i agree. Marking the whole season as watched/unwatched seems useful to me, either via an entry on the three dots menu, or when entering the episodes list, via the top right overflow menu (not as obvious, but not all that bad).

Issues/tickets should be posted here. To contribute, we don't have any guide or overview of the code... Feel free do dive right in, and post here or on the forum any questions/doubts.

PR #582, opened 2018-09-27
Progressbar synchronisation

ThrowsThrowable:
When seeking in Kodi, the progressbars in Kore are not synchronised. Although the Player.OnSeek notifications are captured in kore, they are not passed to the HostConnectionObserver.PlayerEventsObserver.

1. I extended the PlayerEventObserver interface with the method:
https://github.com/xbmc/Kore/blob/65ae63966e8e9b0ecb2d242feaf966ef942f1eb2/app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java#L106-L111
2. I passed the OnSeek notification in the HostConnectionObserver class:
https://github.com/xbmc/Kore/blob/65ae63966e8e9b0ecb2d242feaf966ef942f1eb2/app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java#L412-L420

This enables the classes BaseMediaActivity and NowPlayingFragment to update the progressbars.

Not sure why this was not implemented though, probably a good reason I'm not aware of.

PR #583, opened 2018-09-30
Mark series watched

ThrowsThrowable:
Marking a complete season or even complete tv show as watched / not watched is something I missed in kore. Because @SyncedSynapse liked this as well https://github.com/xbmc/Kore/issues/578#issuecomment-423588288, I started to implement this.

I added three dots menu in the items of both the TVShowListFragment and TVShowProgressFragment.
The code in both classes are very similar so it will probably become one sigle method in a helper class.

Here in words what I'm doing when "Mark watched" is selected on a season item:
- I start a loader to get all episode id's
- when finished, VideoLibrary.SetEpisodeDetails is called for every episode
- each time kodi successfully responds, the progressbar and text of the item is updated. This results is a nice live updating animation.
- when the complete season is marked, the database in synchronised for the complete season.

Some issues, questions:
- Should I use someting like the RefreshItem? This is only available in AbstractInfoFragment...
- when updating the progressbars and text, I should be able to get the current Views for the item because now the wrong views may be updated when scrolling
- after setting the watched states for all episodes, the database is updated for the complete show
LibrarySyncService.SYNC_SINGLE_TVSHOW. Should I implement the option to sync a single season: LibrarySyncService.SYNC_SINGLE_TVSHOW_SEASON?


SyncedSynapse:
Same comments as #582.

PR #588, opened 2018-10-14
[docs] Add Github issue templates

MartijnKaijser:
This adds some issue templates that github can use so users have some guidance on how to fill in the correct information needed.
If user clicks on "new issue" he is presented with some template options to choose from. After that he can fill in the template.
Sadly Github mobile pages are lacking so we need to add a fallback template which is the same as the big report one.
image
This needs to be added to the default branch which currently is master.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,80 @@
+---
+name: Problem report
+about: Create an extensive report to help us document a problem
+
+---
+<!--- Please fill out this template to the best of your ability. You can always edit this issue once you have created it. -->
+<!--- Read the following link before you create a new problem report: https://kodi.wiki/view/HOW-TO:Submit_a_bug_report  -->
+## Bug report
+### Describe the bug
+Here is a clear and concise description of what the problem is:
+<!--- Provide a more detailed introduction to the issue itself, and why you consider it to be a bug -->
+<!--- A bug report that is not clear will be closed -->
+<!--- Put your text below this line -->
+
+
+
+## Expected Behavior
+Here is a clear and concise description of what was expected to happen:
+<!--- Tell us what should happen -->
+<!--- Put your text below this line -->
+
+
+
+## Actual Behavior
+<!--- Tell us what happens instead -->
+<!--- Put your text below this line -->
+
+
+
+## Possible Fix
+<!--- Not obligatory, but suggest a fix or reason for the bug -->
+<!--- Put your text below this line -->
+
+
+
+### To Reproduce
+Steps to reproduce the behavior:
+<!--- Provide a link to a live example, or an unambiguous set of steps to -->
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,80 @@
...
+<!--- Put your text below this line -->
+1. 
+2.
+3.
+
+
+### Debuglog
+<!--- Put your text below this line -->
+<!--- A debuglog is always mandatory when creating an issue. Provide one! -->
+The debuglog can be found here:
```
MartijnKaijser: @poisdeux @SyncedSynapse Not sure if this is wanted or not. You have any ideas/suggestions?

poisdeux: I think this should be optional. Most users reporting issues will not be able to provide a debug log. 
I suggest to make only *Expected Behavior*, *Actual Behavior*, *Steps to reproduce*, and *Your Environment* mandatory. All other sections can be optional.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,80 @@
+---
```
poisdeux: What's the difference between this bug report template and the one called [`ISSUE_TEMPLATE.md`](https://github.com/xbmc/Kore/pull/588/files#diff-1a1c3dd142f76a5fad803a0c52839881)

SyncedSynapse:
What should be the difference between a _Feature suggestion_ and a _Roadmap suggestion_. We don't really have a formal roadmap, so i'm not sure if it will be used.

.github/ISSUE_TEMPLATE.md
MartijnKaijser: We just added them two weeks ago and have not really gotten many tickets yet as we have not advertised them as such. This is mostly a copy of another project that had it like this. I'm fine reducing the template a bit more as long as we get the info we need. So open for suggestions and we can fix/use the same template in all our projects

SyncedSynapse: I don't want to necessarily reduce the template - the structure is fine IMO - i just don't want to introduce too much friction in the bug reporting process by including too many mandatory sections. But i'm fine with having them on the template. 
Anyway, on a second look, the sections i indicated aren't described as mandatory, so i'm fine with it. Including the log shouldn't also be mandatory on Kore (this is probably a difference with Kodi), as we don't have much use for it in the majority of bugs.
So to summarize, i'm fine with this, as long as the log isn't.
My only suggestion would be regarding the order of the sections. Maybe ordering them by their "mandatory" status would be helpful, so i would put the _Your system_ upwards (maybe even the 2nd section) and switch _Steps to reproduce_ with _Possible fix_. If this has been discussed elsewhere and agreed upon, disregard this, it's not very relevant.
- 
-

Issue #597, opened 2018-11-21
Connection problem when search


SecT0uch:
When I start a New search from Kore within any plugin via "Content" tab (Youtube for example), the keyboard appear on screen but not the Send text" pop-up on Kore. Then if I'm going to the remote I have a connection problem.

In contrast, it works well when I'm using Kodi. The pop up appear on Kore. 

Kore 2.4.4
Libreelec 8.2.5 (Kodi 17.6)

erudisch:
I've addressed the problem in this pr https://github.com/xbmc/Kore/pull/598

PR #598, opened 2018-11-23
add addon search

erudisch:
This pr adds a generic search function that should work for every addon that has a search function.

The current behavior of the search buttons in the addons view is nothing but a socket timeout and a not responding app during everlasting reconnections when using tcp connection.

The problem with the addon search is that it is actualy a `GetDirectory` request. But to obtain search results a text input have to be done. To do so you have to send the search string to the backend. But because of `Socket.getOutputStream()` blocks a request for sending until the response of the previous request is received the search string is sent via http as workaround (to overcome this nio channels could be used in the future).

There is also a fallback implementation to prevent socket timeouts and possible reconnection difficulties. This fallback sends a dummy string to the backend to force the response of the search request. The ugly thing is that this dummy string shows up in the addon history.

resolves https://github.com/xbmc/Kore/issues/597

### DIFF
--- a/app/src/main/java/org/xbmc/kore/jsonrpc/HostConnection.java
+++ b/app/src/main/java/org/xbmc/kore/jsonrpc/HostConnection.java
@@ -158,13 +158,22 @@ public class HostConnection {
 
     private static final int DEFAULT_CONNECT_TIMEOUT = 5000; // ms
 
-    private static final int TCP_READ_TIMEOUT = 30000; // ms
+    public static final int TCP_READ_TIMEOUT = 30000; // ms
 
     /**
      * OkHttpClient. Make sure it is initialized, by calling {@link #getOkHttpClient()}
      */
     private OkHttpClient httpClient = null;
     private static final MediaType MEDIA_TYPE_JSON = MediaType.parse("application/json; charset=utf-8");
+
+    /**
+     * Flag to indicate that the tcp response have to be ignored.
+     */
+    private boolean ignoreTcpResponse = false;
+
+    public void setIgnoreTcpResponse(boolean ignoreTcpResponse) {
+        this.ignoreTcpResponse = ignoreTcpResponse;
+    }
 
     /**
      * Creates a new host connection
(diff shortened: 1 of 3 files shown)


SyncedSynapse:
Thanks for this. I will only be able to look into it the next week, but it's not forgotten.

SyncedSynapse:
Once again, thanks and sorry for the delay. Some comments on this PR:

1. Can you please separate this PR in 2, where one only fixes the repositories in gradle (as you previously suggested).

2. I'm not a fan of the `shouldForceHttp` in `HostConnection`, which results in always sending `Input.SendText` through HTTP. You can specify which protocol to use on each method execution, so you don't really need this. Take a look at 
https://github.com/xbmc/Kore/blob/7ae217652c9c211d5abe1e3aa6de4a997754371e/app/src/main/java/org/xbmc/kore/ui/sections/remote/NowPlayingFragment.java#L527
for an example, but basically, in your PR, you can change the `sendTextInput` function to
```
    private void sendTextInput(String text, boolean done, boolean isDummy) {
        HostManager hostManager = HostManager.getInstance(this);
        hostManager.getConnection().setIgnoreTcpResponse(isDummy);

        HostConnection httpHostConnection = new HostConnection(hostManager.getHostInfo());
        httpHostConnection.setProtocol(HostConnection.PROTOCOL_HTTP);

        Input.SendText action = new Input.SendText(text, done);
        action.execute(httpHostConnection, null, null);
    }
```

3. I also don't like that this is implemented in `BaseMediaActivity`. That's a base class for a lot of others, and given that this is kind of an hack it should be as localized as possible, not applied on such a broad class. I suggest that you instead implement `inputOnInputRequested` on `AddonsActivity`, exactly with the code you used (some imports need fixing, but nothing much).

(i'm also not a big fan of including `ignoreTCPResponse` on `HostConnection`, but i understand its usefulness and i don't have an alternative, so it stays)

erudisch:
Thanks for your suggestions

I will do so in the next days.

erudisch:
Hey there

All changes are done. After the google repo pr https://github.com/xbmc/Kore/pull/603 is merged and rebased this should also finish successfully.

Regards
Erik

PR #604, opened 2018-12-09
upgrade of build tools, gradle wrapper, sdk

realrudi:


poisdeux:
You will need to update the [travis configuration](https://github.com/xbmc/Kore/blob/master/.travis.yml) as well.

BTW, any particular reason to upgrade?

erudisch:
No, it's just to be up to date

app/src/main/res/values/dimens.xml
```diff
@@ -141,4 +141,7 @@
     <dimen name="fab_normal_margin">8dp</dimen>
     <dimen name="fab_mini_margin">16dp</dimen>
 
+    <dimen name="remote_icon_size">84dp</dimen>
```
poisdeux: These are not needed as Android will pick the ones closest to the specified smallest width. I even think these will never be used as we use smallest width specific dimensions.

build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.2'
+        classpath 'com.android.tools.build:gradle:3.2.1'
```
poisdeux: Upgrading breaks the unit tests as. For some reason the shared resources required by the unit tests are not found when using the newer gradle version. So unless there is a really good reason to upgrade, I'd postpone the upgrade and wait for a newer version that hopefully also supports shared resources.

app/build.gradle
```diff
@@ -12,7 +12,7 @@ def getVersionName = { ->
 
 android {
     compileSdkVersion 26
```
poisdeux: Why not set this to 28 as well?

SyncedSynapse: What implications did you see? I didn't look very closely, but didn't noticed anything that might impact us...

poisdeux: Sorry, I usually update the `compileSdkVersion` together with the `targetSdkVersion`, so was actually referring to updating both. It compiles fine with 28 so updating the `compileSdkVersion` should be safe.

We might also want to set the `targetSdkVersion` to 28 as we just made a new release, there should be enough time to test if everything still works as intended.

SyncedSynapse:
@poisdeux I think this should be closed, but we should try to update the Gradle version. Does the issue still occurs with newer versions?

poisdeux:
Sorry, totally forgot about this one. Seems to be running fine now with 3.4 of the android gradle plugin.

See #640

PR #605, opened 2018-12-15
Fixed issue with registering observers multiple times

poisdeux:
Registering observers would cause the runnables to be
posted multiple times to the checkerHandler. This resulted
in callbacks being called more than once when a result was
received.


SyncedSynapse:
By the way, i would prefer to include this in the next release, to give us time to test it thoroughly

SyncedSynapse:
Dam... 
@poisdeux sorry, i looked at this PR when you submitted it, and made some comments but somehow forgot to post them...
In fact, my second comment was made after the review (which never got posted). Sorry for not spoting this earlier.

poisdeux:
I see you already refactored the code yourself in #616. Thx.

PR #607, opened 2018-12-24
Catalan translation

ferranroig:
Added catalan translation


SyncedSynapse:
Can you please check the Travis CI error, regarding the translation of "PVR". As far as i can see, it should be "pvr" in lowercase, but please confirm

ferranroig:
Catalan string.xml file fixed

PR #611, opened 2019-01-10
Add fastscroll to lists with RecyclerView

SyncedSynapse:


SyncedSynapse:
@poisdeux When you have time, can you please check this. I would like to include it in the next release, so that we don't loose functionality from the previous one.

Issue #592, opened 2018-11-05
Adding IMDB rating to movies tab


maxdido:
It would be nice to see, in the movies tab, the imdb rating of the movie next to the duration and year. See also this link to the Kodi Remote forum: [Link](https://forum.kodi.tv/showthread.php?tid=337193&pid=2788937#pid2788937) .Here you can see a picture what I have in mind. Probably best to use numbers like 7.4 or 4.2 etc.... and not use stars.

![kore](https://user-images.githubusercontent.com/30314086/48027073-099d5180-e148-11e8-822d-0c310a93ddbf.png)

PR #613, opened 2019-01-12
Added movie ratings to movie listview items

poisdeux:
Implements feature request #592

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/video/MovieListFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/video/MovieListFragment.java
@@ -15,7 +15,6 @@
  */
 package org.xbmc.kore.ui.sections.video;
 
-import android.app.Activity;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.content.res.Resources;
@@ -45,10 +44,12 @@ import org.xbmc.kore.service.library.LibrarySyncService;
 import org.xbmc.kore.ui.AbstractCursorListFragment;
 import org.xbmc.kore.ui.AbstractFragment;
 import org.xbmc.kore.ui.RecyclerViewCursorAdapter;
-import org.xbmc.kore.ui.sections.audio.SongsListFragment;
+import org.xbmc.kore.ui.views.RatingBar;
 import org.xbmc.kore.utils.LogUtils;
 import org.xbmc.kore.utils.UIUtils;
 import org.xbmc.kore.utils.Utils;
+
+import java.util.ArrayList;
 
 /**
  * Fragment that presents the movie list
@@ -64,6 +65,7 @@ public class MovieListFragment extends AbstractCursorListFragment {
     private OnMovieSelectedListener listenerActivity;
 
     private static boolean showWatchedStatus;
+    private static boolean showRating;
 
     @Override
     protected String getListSyncType() { return LibrarySyncService.SYNC_ALL_MOVIES; }
@@ -103,6 +105,7 @@ public class MovieListFragment extends AbstractCursorListFragment {
         }
 
         showWatchedStatus = preferences.getBoolean(Settings.KEY_PREF_MOVIES_SHOW_WATCHED_STATUS, Settings.DEFAULT_PREF_MOVIES_SHOW_WATCHED_STATUS);
+        showRating = preferences.getBoolean(Settings.KEY_PREF_MOVIES_SHOW_RATING, Settings.DEFAULT_PREF_MOVIES_SHOW_RATING);
 
         String sortOrderStr;
         int sortOrder = preferences.getInt(Settings.KEY_PREF_MOVIES_SORT_ORDER, Settings.DEFAULT_PREF_MOVIES_SORT_ORDER);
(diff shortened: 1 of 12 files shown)


SyncedSynapse:
I'm not sure about the presentation of this. It's kind of confusing having "120 min | 2016 | 7.3/10". It's too much info, doesn't look good and it's not obvious what the 7.3/10 refers to (it's also visually distracting having "|" and "\" so close together).
I'm not sure how to solve this, but as is doesn't seem to me as the best option. We could try:
- Using starts instead of of numbers. Probably not as informative, but it's obvious. If using this, we should also keep it separate from the runtime/year.
- Remove the year, which i don't think adds much, but there will inevitably be a backlash from some users who like it.
- Move the rating to the right, tough i'm not sure it will look good either (it will be near the checkmark).

I don't know, i just think that this needs some change.

tomerf:
What about just adding a ⭐ icon after the rating to make it clearer?

SyncedSynapse:
Maybe, but i'm not sure it will solve the issue of too much info in a short space. We might consider moving it somewhere else, maybe to the space between the undertitle and the runtime/year info (needs checking).
Also, mixing text and icons (like 5.1⭐) might look odd/ugly.

btw, not related to this, but i'm tempted to soon the cards interface and just use a plain list. They're not being used according to Google's guidelines (never were in fact) and Google is using them less and less. What do you think.

poisdeux:
I think we should go with the star only option. I'm thinking of using a maximum of 5 stars to indicate the rating instead of 10.

Regarding the card views, I agree. IMO cards in a listview do not really match.

poisdeux:
I've refactored the rating text into a rating bar:

image

image

I still need to tidy up the code but want your opinion on what to show when there is no rating available. Currently it shows the "empty" stars. However, IMO it indicates a 0 rating instead of no rating available. I'm thinking of removing the rating bar when there is no rating available. A disadvantage is that it makes the UI look less consistent. What do you think?

app/src/main/java/org/xbmc/kore/ui/sections/video/MovieListFragment.java
```diff
@@ -425,5 +426,18 @@ public void bindView(Cursor cursor) {
...
+            int movieYear = cursor.getInt(MovieListQuery.YEAR);
+
+            ArrayList<String> duration = new ArrayList<>();
+            if (runtime > 0)
+                duration.add(String.format(context.getString(R.string.minutes_abbrev),
+                                           String.valueOf(runtime)));
+            duration.add(String.valueOf(movieYear));
+
+            return TextUtils.join(" | ", duration);
+        }
```
SyncedSynapse: I have no problem with this, but i must point out that the previous version was slightly more efficient, in terms of not needing as many objects allocated (the arraylist). This code is run for every list item, and even though i don't think this constitutes a problem, this kind of micro-optimizations might add up over time.

Anyway, i'm probably splitting hairs here.

app/src/main/res/drawable/ic_star_white_24dp.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <vector android:alpha="1.00" android:height="24dp"
-    android:viewportHeight="24.0" android:viewportWidth="24.0"
+    android:viewportHeight="22.0" android:viewportWidth="22.0"
```
SyncedSynapse: Why the change here? If some tweak to the dimensions is needed shouldn't it be done at the point of usage instead of at the icon definition?

SyncedSynapse:
I think this is (much) better. I must confess that i'm not a fan, but that's a personal opinion (easily solved by a small request i'll ask later), and i see the value in it and why people might want it.

Regarding what to show when no rating is available, either way is fine by me, but maybe removing the rating bar is better, and i don't think it makes the UI less consistent, given that the checkmark already isn't displayed if the movie is unwatched. (BTW how are you getting the white/grey checkmarks? They should all be in colored in the accent color or be invisible...).

Now for the request: i think this should be optional, the same way the watched status is - by setting an option in the overflow menu (and i vote for it to be off by default, but like i said that's a personal opinion). And we should set the view to GONE if it isn't being shown (which also applies to the watched status, and is not being done right now).

Also check how this looks in landscape - i found that it interferes with the meta info, which is unavoidable but it's one more reason for it being optional.

app/src/main/res/layout/grid_item_playlist.xml
```diff
@@ -73,7 +73,7 @@
             android:layout_below="@id/title"
             android:layout_toRightOf="@id/art"
             android:layout_toEndOf="@id/art"
-            android:layout_toLeftOf="@id/duration"
+            android:layout_toLeftOf="@id/meta_info"
```
SyncedSynapse: Note that this makes #618 break

app/src/main/res/drawable/ic_star_white_24dp.xml
poisdeux: The actual vector drawable is only 22x22 which means that setting the viewport to 24x24 adds a hardcoded padding of 2. This became apparent when showing the stars side-by-side where the space between the stars was too much. I didn't want to add a negative margin and instead changed the vector drawable.

app/src/main/java/org/xbmc/kore/ui/sections/video/MovieListFragment.java
poisdeux: If scrolling becomes slow we will need to refactor this completely and generate the content, at least the heavy parts, on a separate thread.

poisdeux:
I removed the rating bar when the rating is zero.
As for the rating bar being optional and off by default. I agree that it should be possible to disable it. But I do not agree that it should be off by default. Especially as it is something new, we should show it to all users at first. If they find it annoying they can check the settings to see if it can be disabled just like the watched checkmark.

poisdeux:
Code has been refactored and cleaned up.
The rating bar can be disabled from the overflow menu. It is enabled by default.
When the rating bar is enabled and there is more meta info to show than fits the container the info is ellipsized. I think this should be a clear indication to the user to either turn their device in landscape mode or disable the rating bar to make the meta info fully readable.

SyncedSynapse:
Looks good, merging.

Issue #602, opened 2018-12-06
Kore does not show "Playing" UI after I start a video


jamesbroadhead:
### Describe the bug
From the "Remote Control" UI, after clicking the center button to play a video, Kore does not display the "Playing" UI. 

Sometimes, after 10-30 seconds, the Playing UI appears 
Other times, the Playing UI never appears. The app remains on the "connected to <server>" UI

If I force-quit the application and restart it, the Playing UI appears immediately 

## Expected Behavior
The 'Playing' UI should be shown as soon as the video begins to play. 

## Actual Behavior
Sometimes it never shows - as a result, I cannot pause or stop the video. 

## Possible Fix
Poll the kodi api more frequently? 

### To Reproduce
1. Navigate to a video using the 'remote control' UI
2. Use the middle button to play the video 
3. Wait. 

### Debuglog
Log: https://paste.kodi.tv/huniqijoyi.kodi
Started video at 23:52.
Kore did not show the 'playing' 
At 23:52:59, I restarted Kore, it showed the 'playing' UI & I stopped the video

Debug Log: http://dpaste.com/2QH0DR9

### Screenshots 
Here are some links or screenshots to help explain the problem:
https://photos.app.goo.gl/Sj3crzm7E5zdYUpa7 - the remote control ui
https://photos.app.goo.gl/6pvkCb9tjgBTieeA6 - the "playing" ui

## Additional context or screenshots (if appropriate)
Here is some additional context or explanation that might help:

### Your Environment
Used Operating system:
 - Mobile device: Huawei P20 Pro
 - Android version: 8.1.0
 - Kodi version and platform: 18.0_rc2  (this is a long running problem however)
 - Kore version: 2.4.4

SyncedSynapse:
Thanks for the detailed report.
I honestly have never seen this happening. On slower Kodi devices it might take a couple of secs to show up (for Kore to be notifies), but i shows.
Please check if the TCP option is enable in Kore for that Kodi instance, though it should work either way.
I'm more inclined to believe it has something to do with you device. Huawei has been pulling some naughty stuff with Android lately, namely killing (perfectly valid) background processes, in an effort to appear well in battery tests. Can you check if your battery saving settings isn't preventing Kore from functioning correctly (maybe whitelist Kore, or try another phone).

poisdeux:
Finally had some time to look into this. It seems there is a race condition causing problems.
Below is some logging when starting a video using the reported method.
[...]

PR #616, opened 2019-02-08
Add support for OnAVStart event and tweak HostConnectionOberver

SyncedSynapse:
This fixes https://github.com/xbmc/Kore/issues/602 by supporting the OnAVStart event, which supersedes OnPlay in Kodi Leia (introduced in this Kodi PR: https://github.com/xbmc/xbmc/pull/13726).

Some notes regarding this PR:
1. If Kodi is Leia or higher ignore the OnPlay event in favour of OnAVStart, so that we don't twice process the same thing;
2. Ignore the OnAVChange event in the `HostConnectionObserver`, as this event is fired very often by Kodi (about 3 times in quick succession when playing starts), and we are really interested is in the OnAVStart event;
3. This PR also introduces a mechanism to prevent simultaneous calls to `checkWhatsPlaying`. With Kodi Leia we're getting more events and processing them results in more calls to `checkWhatsPlaying`. This PR introduces a flag that tries to guarantee that there's only one call going on. It's not thread-safe, as i think it's not worth the trouble (if we get a race condition in this call there's no harm done, we're just calling Kodi one more time than needed). I'm not happy about the solution because every exit point of `checkWhatsPlaying` needs to unset the flag and that can introduce bugs in the future. I plan to revisit this to maybe use `Futures` so localizing the change to the flag in the same function.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
+++ b/app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
@@ -58,10 +58,10 @@ public class HostConnectionObserver
     public interface ApplicationEventsObserver {
         /**
          * Notifies the observer that volume has changed
-         * @param volume
-         * @param muted
+         * @param volume Volume level
+         * @param muted Is muted
          */
-        public void applicationOnVolumeChanged(int volume, boolean muted);
+        void applicationOnVolumeChanged(int volume, boolean muted);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


SyncedSynapse:
@poisdeux When you're able to, i would appreciate if you could review this, as i think we should do a quick release to fix that issue. There are still other stuff to be done, but at least this fixes the most visible one.

poisdeux:
Works great. Code looks OK.

actricom:
This update seems to have been released in F-Droid and as an apk. Any chance of uploading it to the Play Store? It's still on 2.4.6.

SyncedSynapse:
Sorry for the delay...

actricom:
Thanks for that. I've just picked up the latest version and I'll give it a try later.

PR #618, opened 2019-02-15
Implemented showing playlists even when not playing

poisdeux:
Implemented showing a music/video/picture playlist in PlaylistFragment even
when playback has stopped but the playlist is still available on Kodi.
The playlist is now only cleared if the Playlist.OnClear event is received
from Kodi.

As the pictures playlist does not support moving/removing items this has
been disabled for the pictures playlist only.

Added a new package org.xbmc.kore.host.actions to hold all Callable's that
we use to handle complex interactions with Kodi that use multiple JSON RPC
calls.

Fixed issue in HostConnectionObserver in method notifySomethingIsPlaying
where hostState.lastGetItemResult.label might not be set.

Refactored HostConnection(Callable) into something using
callbacks instead of the EventBus. With the EventBus we don't
really solve the problem with callbacks and boilerplate code
as it will simply be replaced by EventBus methods. However, a
major drawback is that we need to add comments to notify the developer
on how to receive the result. With ApiCallback this is evident as
the developer needs to provide an instance of ApiCallback as argument.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
+++ b/app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
@@ -18,6 +18,7 @@ package org.xbmc.kore.host;
 import android.os.Handler;
 import android.os.Looper;
 
+import org.xbmc.kore.host.actions.GetPlaylist;
 import org.xbmc.kore.jsonrpc.ApiCallback;
 import org.xbmc.kore.jsonrpc.HostConnection;
 import org.xbmc.kore.jsonrpc.method.JSONRPC;
@@ -25,6 +26,7 @@ import org.xbmc.kore.jsonrpc.method.Player;
 import org.xbmc.kore.jsonrpc.notification.Application;
 import org.xbmc.kore.jsonrpc.notification.Input;
 import org.xbmc.kore.jsonrpc.notification.Player.NotificationsData;
+import org.xbmc.kore.jsonrpc.notification.Playlist;
 import org.xbmc.kore.jsonrpc.notification.System;
 import org.xbmc.kore.jsonrpc.type.ApplicationType;
 import org.xbmc.kore.jsonrpc.type.ListType;
@@ -49,8 +51,25 @@ public class HostConnectionObserver
         implements HostConnection.PlayerNotificationsObserver,
                    HostConnection.SystemNotificationsObserver,
                    HostConnection.InputNotificationsObserver,
-                   HostConnection.ApplicationNotificationsObserver {
+                   HostConnection.ApplicationNotificationsObserver,
+                   HostConnection.PlaylistNotificationsObserver {
     public static final String TAG = LogUtils.makeLogTag(HostConnectionObserver.class);
+
+    public interface PlaylistEventsObserver {
+        /**
+         * @param playlistId of playlist that has been cleared
+         */
+        void playlistOnClear(int playlistId);
+
+        void playlistChanged(int playlistId);
+
+        /**
+         * @param playlists the available playlists on the server
+         */
+        void playlistsAvailable(ArrayList<GetPlaylist.GetPlaylistResult> playlists);
+
+        void playlistOnError(int errorCode, String description);
+    }
 
     /**
      * Interface that an observer has to implement to receive playlist events
@@ -87,8 +106,8 @@ public class HostConnectionObserver
          * @param getItemResult Currently playing item, obtained by a call to {@link org.xbmc.kore.jsonrpc.method.Player.GetItem}
          */
         void playerOnPlay(PlayerType.GetActivePlayersReturnType getActivePlayerResult,
-                                 PlayerType.PropertyValue getPropertiesResult,
-                                 ListType.ItemsAll getItemResult);
+                          PlayerType.PropertyValue getPropertiesResult,
+                          ListType.ItemsAll getItemResult);
 
         /**
          * Notifies that something is paused
@@ -97,8 +116,8 @@ public class HostConnectionObserver
          * @param getItemResult Currently paused item, obtained by a call to {@link org.xbmc.kore.jsonrpc.method.Player.GetItem}
          */
         void playerOnPause(PlayerType.GetActivePlayersReturnType getActivePlayerResult,
-                                  PlayerType.PropertyValue getPropertiesResult,
-                                  ListType.ItemsAll getItemResult);
+                           PlayerType.PropertyValue getPropertiesResult,
+                           ListType.ItemsAll getItemResult);
 
         /**
          * Notifies that media is stopped/nothing is playing
@@ -143,15 +162,19 @@ public class HostConnectionObserver
      */
     private List<PlayerEventsObserver> playerEventsObservers = new ArrayList<>();
     private List<ApplicationEventsObserver> applicationEventsObservers = new ArrayList<>();
+    private List<PlaylistEventsObserver> playlistEventsObservers = new ArrayList<>();
 
     // Associate the Handler with the UI thread
+    int checkPlaylistCounter = 0;
     private Handler checkerHandler = new Handler(Looper.getMainLooper());
     private Runnable httpCheckerRunnable = new Runnable() {
         @Override
         public void run() {
             final int HTTP_NOTIFICATION_CHECK_INTERVAL = 2000;
             // If no one is listening to this, just exit
-            if (playerEventsObservers.isEmpty() && applicationEventsObservers.isEmpty())
+            if (playerEventsObservers.isEmpty()
+                && applicationEventsObservers.isEmpty()
+                && playlistEventsObservers.isEmpty())
                 return;
 
             if (!playerEventsObservers.isEmpty())
@@ -159,6 +182,12 @@ public class HostConnectionObserver
 
             if (!applicationEventsObservers.isEmpty())
                 getApplicationProperties();
+
+            if (!playlistEventsObservers.isEmpty() && checkPlaylistCounter > 1) {
+                checkPlaylist();
+                checkPlaylistCounter = 0;
+            }
+            checkPlaylistCounter++;
 
             checkerHandler.postDelayed(this, HTTP_NOTIFICATION_CHECK_INTERVAL);
         }
@@ -168,7 +197,8 @@ public class HostConnectionObserver
         @Override
         public void run() {
             // If no one is listening to this, just exit
-            if (playerEventsObservers.isEmpty() && applicationEventsObservers.isEmpty())
+            if (playerEventsObservers.isEmpty() && applicationEventsObservers.isEmpty() &&
+                playlistEventsObservers.isEmpty())
                 return;
 
             final int PING_AFTER_ERROR_CHECK_INTERVAL = 2000,
@@ -181,9 +211,14 @@ public class HostConnectionObserver
                     // we were in a error or uninitialized state, update
                     if ((!playerEventsObservers.isEmpty()) &&
                         ((hostState.lastCallResult == PlayerEventsObserver.PLAYER_NO_RESULT) ||
-                        (hostState.lastCallResult == PlayerEventsObserver.PLAYER_CONNECTION_ERROR))) {
+                         (hostState.lastCallResult == PlayerEventsObserver.PLAYER_CONNECTION_ERROR))) {
                         LogUtils.LOGD(TAG, "Checking what's playing because we don't have info about it");
                         checkWhatsPlaying();
+                    }
+
+                    if ((!playlistEventsObservers.isEmpty()) &&
+                        (hostState.lastCallResult == PlayerEventsObserver.PLAYER_CONNECTION_ERROR)) {
+                        checkPlaylist();
                     }
 
                     checkerHandler.postDelayed(tcpCheckerRunnable, PING_AFTER_SUCCESS_CHECK_INTERVAL);
@@ -323,16 +358,44 @@ public class HostConnectionObserver
         }
     }
 
-    private void startCheckerHandler() {
-        // Check if checkerHandler is already running, to prevent multiple runnables to be posted
-        // when multiple observers are registered.
-        if (checkerHandler.hasMessages(0))
+    /**
+     * Registers a new observer that will be notified about playlist events
+     * @param observer Observer
+     * @param replyImmediately Whether to immediately issue a request if there are playlists available
+     */
+    public void registerPlaylistObserver(PlaylistEventsObserver observer, boolean replyImmediately) {
+        if (this.connection == null)
             return;
 
-        if (connection.getProtocol() == HostConnection.PROTOCOL_TCP) {
-            checkerHandler.post(tcpCheckerRunnable);
-        } else {
-            checkerHandler.post(httpCheckerRunnable);
+        if ( ! playlistEventsObservers.contains(observer) ) {
+            playlistEventsObservers.add(observer);
+        }
+
+        if (playlistEventsObservers.size() == 1) {
+            if (connection.getProtocol() == HostConnection.PROTOCOL_TCP) {
+                connection.registerPlaylistNotificationsObserver(this, checkerHandler);
+            }
+
+            startCheckerHandler();
+        }
+
+        if (replyImmediately)
+            checkPlaylist();
+    }
+
+    public void unregisterPlaylistObserver(PlayerEventsObserver observer) {
+        playlistEventsObservers.remove(observer);
+
+        LogUtils.LOGD(TAG, "Unregistering playlist observer " + observer.getClass().getSimpleName() +
+                           ". Still got " + playlistEventsObservers.size() +
+                           " observers.");
+
+        if (playlistEventsObservers.isEmpty()) {
+            // No more observers, so unregister us from the host connection, or stop
+            // the http checker thread
+            if (connection.getProtocol() == HostConnection.PROTOCOL_TCP) {
+                connection.unregisterPlaylistNotificationsObserver(this);
+            }
         }
     }
 
@@ -350,6 +413,7 @@ public class HostConnectionObserver
             connection.unregisterSystemNotificationsObserver(this);
             connection.unregisterInputNotificationsObserver(this);
             connection.unregisterApplicationNotificationsObserver(this);
+            connection.unregisterPlaylistNotificationsObserver(this);
             checkerHandler.removeCallbacks(tcpCheckerRunnable);
         }
         hostState.lastCallResult = PlayerEventsObserver.PLAYER_NO_RESULT;
@@ -370,7 +434,9 @@ public class HostConnectionObserver
     public void onPlay(org.xbmc.kore.jsonrpc.notification.Player.OnPlay notification) {
         // Ignore this if Kodi is Leia or higher, as we'll be properly notified via OnAVStart
         // See https://github.com/xbmc/Kore/issues/602 and https://github.com/xbmc/xbmc/pull/13726
-        if (connection.getHostInfo().isLeiaOrLater()) {
+        // Note: OnPlay is still required for picture items.
+        if (connection.getHostInfo().isLeiaOrLater() &&
+            ! notification.data.item.type.contentEquals("picture") ) {
             LogUtils.LOGD(TAG, "OnPlay notification ignored. Will wait for OnAVStart.");
             return;
         }
@@ -454,6 +520,40 @@ public class HostConnectionObserver
         }
     }
 
+    @Override
+    public void onPlaylistCleared(Playlist.OnClear notification) {
+        for (PlaylistEventsObserver observer : playlistEventsObservers) {
+            observer.playlistOnClear(notification.playlistId);
+        }
+    }
+
+    @Override
+    public void onPlaylistItemAdded(Playlist.OnAdd notification) {
+        for (PlaylistEventsObserver observer : playlistEventsObservers) {
+            observer.playlistChanged(notification.playlistId);
+        }
+    }
+
+    @Override
+    public void onPlaylistItemRemoved(Playlist.OnRemove notification) {
+        for (PlaylistEventsObserver observer : playlistEventsObservers) {
+            observer.playlistChanged(notification.playlistId);
+        }
+    }
+
+    private void startCheckerHandler() {
+        // Check if checkerHandler is already running, to prevent multiple runnables to be posted
+        // when multiple observers are registered.
+        if (checkerHandler.hasMessages(0))
+            return;
+
+        if (connection.getProtocol() == HostConnection.PROTOCOL_TCP) {
+            checkerHandler.post(tcpCheckerRunnable);
+        } else {
+            checkerHandler.post(httpCheckerRunnable);
+        }
+    }
+
     private void getApplicationProperties() {
         org.xbmc.kore.jsonrpc.method.Application.GetProperties getProperties =
                 new org.xbmc.kore.jsonrpc.method.Application.GetProperties(org.xbmc.kore.jsonrpc.method.Application.GetProperties.VOLUME,
@@ -475,6 +575,62 @@ public class HostConnectionObserver
                 notifyConnectionError(errorCode, description, playerEventsObservers);
             }
         }, checkerHandler);
+    }
+
+    private ArrayList<GetPlaylist.GetPlaylistResult> prevGetPlaylistResults = new ArrayList<>();
+    private boolean isCheckingPlaylist = false;
+    private void checkPlaylist() {
+        if (isCheckingPlaylist)
+            return;
+
+        isCheckingPlaylist = true;
+
+        connection.execute(new GetPlaylist(connection), new ApiCallback<ArrayList<GetPlaylist.GetPlaylistResult>>() {
+            @Override
+            public void onSuccess(ArrayList<GetPlaylist.GetPlaylistResult> result) {
+                isCheckingPlaylist = false;
+
+                if (result.isEmpty()) {
+                    callPlaylistsOnClear(prevGetPlaylistResults);
+                    return;
+                }
+
+                for (PlaylistEventsObserver observer : playlistEventsObservers) {
+                    observer.playlistsAvailable(result);
+                }
+
+                // Handle onClear for HTTP only connections
+                for (GetPlaylist.GetPlaylistResult getPlaylistResult : result) {
+                    for (int i = 0; i < prevGetPlaylistResults.size(); i++) {
+                        if (getPlaylistResult.id == prevGetPlaylistResults.get(i).id) {
+                            prevGetPlaylistResults.remove(i);
+                            break;
+                        }
+                    }
+                }
+
+                callPlaylistsOnClear(prevGetPlaylistResults);
+
+                prevGetPlaylistResults = result;
+            }
+
+            @Override
+            public void onError(int errorCode, String description) {
+                isCheckingPlaylist = false;
+
+                for (PlaylistEventsObserver observer : playlistEventsObservers) {
+                    observer.playlistOnError(errorCode, description);
+                }
+            }
+        }, new Handler());
+    }
+
+    private void callPlaylistsOnClear(ArrayList<GetPlaylist.GetPlaylistResult> clearedPlaylists) {
+        for (GetPlaylist.GetPlaylistResult getPlaylistResult : clearedPlaylists) {
+            for (PlaylistEventsObserver observer : playlistEventsObservers) {
+                observer.playlistOnClear(getPlaylistResult.id);
+            }
+        }
     }
 
     /**
@@ -544,7 +700,7 @@ public class HostConnectionObserver
                 PlayerType.PropertyName.AUDIOSTREAMS,
                 PlayerType.PropertyName.SUBTITLES,
                 PlayerType.PropertyName.PLAYLISTID,
-        };
+                };
 
         Player.GetProperties getProperties = new Player.GetProperties(getActivePlayersResult.playerid, propertiesToGet);
         getProperties.execute(connection, new ApiCallback<PlayerType.PropertyValue>() {
@@ -606,7 +762,7 @@ public class HostConnectionObserver
                 ListType.FieldsAll.WRITER,
                 ListType.FieldsAll.YEAR,
                 ListType.FieldsAll.DESCRIPTION,
-        };
+                };
 //        propertiesToGet = ListType.FieldsAll.allValues;
         Player.GetItem getItem = new Player.GetItem(getActivePlayersResult.playerid, propertiesToGet);
         getItem.execute(connection, new ApiCallback<ListType.ItemsAll>() {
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/xbmc/kore/ui/widgets/PlaylistsBar.java
```diff
@@ -0,0 +1,175 @@
...
+            }
+        });
+    }
+
+    private void enableTab(int tabPosition) {
+        getTabAt(tabPosition).getIcon().setAlpha(255);
+    }
+
+    private void disableTab(int tabPosition) {
+        getTabAt(tabPosition).getIcon().setAlpha(127);
```
SyncedSynapse: Somehow this is interfering with the same icon on other places. The movies/music/pictures icons on the main remote shortcut bar and on the navigation drawer also become dimmed. Android seems to be caching them and using the same copy on all places.

doc/diagrams/sequence/apimethodcallprocedure.puml
```diff
@@ -0,0 +1,79 @@
+@startuml
```
SyncedSynapse: This file seems to be an old leftover, right?

app/src/main/res/layout/fragment_playlist.xml
```diff
@@ -15,20 +15,35 @@
...
+        app:tabBackground="?attr/contentBackgroundColor"
+        style="@style/Widget.Button.Borderless"
+        android:tint="?attr/defaultButtonColorFilter"
+        android:elevation="2dp"/>
+
     <!-- Playlist -->
     <org.xbmc.kore.ui.viewgroups.DynamicListView
         android:id="@+id/playlist"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
```
SyncedSynapse: Changing this to match_parent will make the list start from the top, which i think is more intuitive.

app/src/main/java/org/xbmc/kore/ui/viewgroups/DynamicListView.java
```diff
@@ -167,6 +172,9 @@ public boolean isItemBeingDragged() {
     private AdapterView.OnItemLongClickListener mOnItemLongClickListener =
             new AdapterView.OnItemLongClickListener() {
                 public boolean onItemLongClick(AdapterView<?> arg0, View arg1, int position, long id) {
+                    if (!itemDraggingEnabled)
+                        return false;
+
                     mTotalOffset = 0;
 
                     mOriginalPosition = position;
```
SyncedSynapse: I haven't figured out why, but now dragging an item on the list will cause a npe in `onSwapFinished`

SyncedSynapse:
This looks nice, though i think there are still some bugs to iron out.

I've started reviewing it and added some comments, but i haven't got the change to review it all, so i'll add some more comments tomorrow (hopefully)

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
```diff
@@ -160,6 +180,10 @@ public void run() {
             if (!applicationEventsObservers.isEmpty())
                 getApplicationProperties();
 
+            if ( ! playlistEventsObservers.isEmpty() ) {
+                checkPlaylist();
+            }
+
```
SyncedSynapse: We should be careful about this. Through HTTP we are polling Kodi every 2 seconds. Up until now, we were calling Kodi 4 times (`getActivePlayers`, `getPlayerProperties`, `getItems` and `getProperties`). With this, we're adding 4 more calls, for a total if 8 calls to Kodi, every 2 seconds.
We should evaluate if this might start to impact Kodi. One possible solution is to poll the playlists less often, in another runnable that runs every 4 seconds or so (that was the initial though, when we had more than one runnable here).

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
```diff
@@ -186,6 +211,10 @@ public void onSuccess(String result) {
                         checkWhatsPlaying();
                     }
 
+                    if (! playlistEventsObservers.isEmpty()) {
+                        checkPlaylist();
+                    }
+
```
SyncedSynapse: This is contrary to the current design. Through TCP we are not polling Kodi, we rely on Kodi's notifications to be notified that something changed, so we don't need to Poll anything (this runnable just pings Kodi, so that it doesn't close the TCP connection, which it does after 30 seconds of inactivity).
So, to keep consistency, we should implement the `Playlist.onAdd` and `Playlist.onRemove` notifications in `HostConnection`, catch them here, and call `checkPlaylist` there.

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
```diff
@@ -477,6 +548,27 @@ public void onError(int errorCode, String description) {
...
+            @Override
+            public void onError(int errorCode, String description) {
+                for (PlaylistEventsObserver observer : playlistEventsObservers) {
+                    observer.playlistOnError(errorCode, description);
+                }
+            }
+        }, new Handler());
+    }
+
```
SyncedSynapse: Edit: i'm not saying that we should definitely store and check the previous results, just that we should evaluate if that's feasible, and if not, what are the consequences (observers being continuously notified, even without any changes, which causes them to update the UI unnecessarily).

app/src/main/java/org/xbmc/kore/jsonrpc/HostConnection.java
```diff
@@ -101,6 +105,10 @@
         void onVolumeChanged(Application.OnVolumeChanged notification);
     }
 
+    public interface PlaylistNotificationsObserver {
+        void onPlaylistCleared(Playlist.OnClear notification);
+    }
+
```
SyncedSynapse: Pointed out previously: The other notifications should be implemented here, so that `HostConnectionObserver` doesn't need to poll Kodi if connection is through TCP.

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
SyncedSynapse: I'll have some more comments on the `HostConnection.execute(Callable)`, but here i'd just like to point out that, unlike in other notifications, we are not storing the previous result, and only passing the notification if something changes.
So, we are always notifying observers of `playlistsAvailable`/`playlistOnError` - every 2 seconds if through HTTP. A notification should only be sent when something changes - because it makes sense conceptually, and it minimizes observers work, for instance if they are always getting the `somethingIsPlaying` notification and redrawing the UI it can cause flicker - , and that's why in the other cases we're saving and checking the previous results. I'm not sure if that can be done with Playlists, but we should be aware of this.

app/src/main/java/org/xbmc/kore/jsonrpc/HostConnection.java
```diff
@@ -119,10 +127,6 @@
 	 * Socket used to communicate through TCP
 	 */
 	private Socket socket = null;
-	/**
-	 * Listener {@link Thread} that will be listening on the TCP socket
-	 */
-	private Thread listenerThread = null;
```
SyncedSynapse: I'll just point out that i liked having this thread separate, as this is a "special" thread that is always running (blocked on the TCP connection) so it might merit a special treatment.

app/src/main/java/org/xbmc/kore/jsonrpc/HostConnection.java
```diff
@@ -380,6 +409,51 @@ public void onError(int errorCode, String description) {
...
+    
+    /**
+     * Executes the {@link Callable} and waits for it to finish on a background thread. The
+     * result is returned using the {@link ApiCallback} and handler
+     * @param callable executed using an {@link ExecutorService}
+     * @param apiCallback used to return the result of the callable
+     * @param handler used to execute the {@link ApiCallback} methods
+     * @param <T>
+     */
+    public <T> void execute(Callable<T> callable, final ApiCallback<T> apiCallback, final Handler handler) {
```
SyncedSynapse: I think that this method isn't consistent with the other `execute`, or even with `HostConnection`. `execute` is defined as executing a jsonrpc method on the server, and return using either a `Future` or a callback.
This method is very generic - it runs a `callable` on a background thread, and posts the result on a callback (which is `Apicallback` just because it's handy). That has nothing to do with `HostConnection`, it's generic (and i don't like using `ApiCallback` here, because of that same reason).
Maybe we should think this more structurally. What is needed here? A way to execute several jsonrpc methods in sequence? I'm not convinced that we need to have something specific to do this (versus just creating a runnable and executing it in place in a thread), but if we need it it might make sense to create a specific class for it (a la `HostAction`, though i'm not convinced that it should be needed) or just look at `CompletableFuture`.

Anyway, this can be merged as is for now, but i plan to revisit this.

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
```diff
@@ -454,6 +505,26 @@ public void onVolumeChanged(Application.OnVolumeChanged notification) {
...
+    public void onPlaylistCleared(Playlist.OnClear notification) {
+        for (PlaylistEventsObserver observer : playlistEventsObservers) {
+            observer.playlistOnClear(notification.playlistId);
+        }
+    }
+
+    private void startCheckerHandler() {
+        // Check if checkerHandler is already running, to prevent multiple runnables to be posted
+        // when multiple observers are registered.
+        if ( checkerHandler.hasMessages(0) )
```
SyncedSynapse: For consistency with the rest of the code, no spaces inside ()

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
```diff
@@ -160,6 +180,10 @@ public void run() {
             if (!applicationEventsObservers.isEmpty())
                 getApplicationProperties();
 
+            if ( ! playlistEventsObservers.isEmpty() ) {
```
SyncedSynapse: For consistency with the rest of the code, no spaces inside ()

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
```diff
@@ -83,21 +89,30 @@
...
+    private PlayerType.GetActivePlayersReturnType lastGetActivePlayerResult;
+    private HashMap<String, PlaylistHolder> playlists = new HashMap<>();
+
+    private enum PLAYER_STATE {
+        CONNECTION_ERROR,
+        NO_RESULTS_YET,
+        PLAYING,
+        PAUSED,
+        STOPPED
+    }
```
SyncedSynapse: Similar to what's on `HostConnectionObserver.PlayerEventsObserver`?

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
```diff
@@ -127,11 +144,34 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
...
             }
         });
 
+        playlistsBar.setOnPlaylistSelectedListener(new PlaylistsBar.OnPlaylistSelectedListener() {
+            @Override
+            public void onPlaylistSelected(String playlistType) {
+                userSelectedTab = true; // do not switch to active playlist when user selected a tab
+                displayPlaylist();
+                PlaylistHolder playlistHolder = playlists.get(playlistType);
+                if ( playlistHolder != null) {
```
SyncedSynapse: For consistency with the rest of the code, no spaces inside ()

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
```diff
@@ -170,10 +214,9 @@ public void onDestroyView() {
     public boolean onOptionsItemSelected(MenuItem item) {
         switch (item.getItemId()) {
             case R.id.action_clear_playlist:
-                Playlist.Clear action = new Playlist.Clear(currentPlaylistId);
+                int playlistId = playlists.get(playlistsBar.getSelectedPlaylistType()).getPlaylistId();
+                Playlist.Clear action = new Playlist.Clear(playlistId);
                 action.execute(hostManager.getConnection(), defaultStringActionCallback, callbackHandler);
-                // If we are playing something, refresh playlist
-                forceRefreshPlaylist();
```
SyncedSynapse: Sure we don't need to refresh the playlists after a clear?

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
```diff
@@ -182,88 +225,99 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+        playerState = PLAYER_STATE.PLAYING;
 
-        // Save results
-        lastCallResult = PLAYER_IS_PLAYING;
-        lastGetActivePlayerResult = getActivePlayerResult;
-        lastGetPropertiesResult = getPropertiesResult;
         lastGetItemResult = getItemResult;
+        lastGetActivePlayerResult = getActivePlayerResult;
+
+        refreshPlaylist(lastGetActivePlayerResult.type);
```
SyncedSynapse: Now that we have the playlistsNotifications do we need this? Why calling Kodi to get the playlists when we receive a onPlay notification, if we are also receiving the Playlists notification with the result already?

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
```diff
@@ -182,88 +225,99 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
-        }
+        playerState = PLAYER_STATE.PAUSED;
 
-        lastCallResult = PLAYER_IS_PAUSED;
-        lastGetActivePlayerResult = getActivePlayerResult;
-        lastGetPropertiesResult = getPropertiesResult;
         lastGetItemResult = getItemResult;
+        lastGetActivePlayerResult = getActivePlayerResult;
+
+        refreshPlaylist(getActivePlayerResult.type);
```
SyncedSynapse: Same comment as before

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
```diff
@@ -289,101 +344,121 @@ public void playerNoResultsYet() {
...
+
+        PlaylistHolder playlistHolder = playlists.get(playlistsBar.getSelectedPlaylistType());
+        if (playlistHolder == null) {
             displayEmptyPlaylistMessage();
             return;
         }
-        switchToPanel(R.id.playlist);
+
+        GetPlaylist.GetPlaylistResult getPlaylistResult = playlistHolder.getPlaylistResult;
+        if ( getPlaylistResult == null ) {
```
SyncedSynapse: For consistency with the rest of the code, no spaces inside ()

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
```diff
@@ -289,101 +344,121 @@ public void playerNoResultsYet() {
...
 
         //If a user is dragging a list item we must not modify the adapter to prevent
         //the dragged item's adapter position from diverging from its listview position
-        if (!playlistListView.isItemBeingDragged()) {
-            // Set items, which call notifyDataSetChanged
-            playListAdapter.setPlaylistItems(playlistItems);
-            highlightItem(getItemResult, playlistItems);
-        } else {
-            highlightItem(getItemResult, playListAdapter.playlistItems);
+        if ( ! playlistListView.isItemBeingDragged() ) {
```
SyncedSynapse: For consistency with the rest of the code, no spaces inside ()
And i would prefer no spaces after ! also

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
```diff
@@ -289,101 +344,121 @@ public void playerNoResultsYet() {
...
+            playListAdapter.setPlaylistItems(getPlaylistResult.items);
         }
+
+        if ( isPlaying(getPlaylistResult) )
```
SyncedSynapse: For consistency with the rest of the code, no spaces inside ()

app/src/main/java/org/xbmc/kore/jsonrpc/HostConnection.java
```diff
@@ -195,7 +207,7 @@ public HostConnection(final HostInfo hostInfo, int connectTimeout) {
         // Start with the default host protocol
         this.protocol = hostInfo.getProtocol();
         // Create a single threaded executor
-        this.executorService = Executors.newSingleThreadExecutor();
+        this.executorService = Executors.newFixedThreadPool(10);
```
SyncedSynapse: This might have unintended consequences. Having a single thread guaranteed that calls to Kodi were sequential, which is not guaranteed now.
I think that one example of this is happening in `PlaylistFragment`, when you choose the `Remove` action of an item in its overflow menu. That calls `Remove` on Kodi, doesn't care about the result, and immediately calls the code to refresh the playlist. With a single thread this would work because the `GetPlaylists` would only run after `Remove`, now that's not true, and it's causing the item not to be removed on the UI...
If this change stays, we need to review every single call to Kodi that doesn't care about the result to make sure that another call doesn't immediately follows, and expects the previous to have run.

app/src/main/java/org/xbmc/kore/ui/widgets/PlaylistsBar.java
poisdeux: I am unable to reproduce this on an emulator running API 15 and also tested on API 28. On what device and Android version does this occur?

app/src/main/res/layout/fragment_playlist.xml
poisdeux: Ah yes. When the list does not contain enough elements to fill the screen they show at the bottom now. I changed it to `match_parent` which looks much better.

doc/diagrams/sequence/apimethodcallprocedure.puml
poisdeux: removed

app/src/main/java/org/xbmc/kore/ui/widgets/PlaylistsBar.java
SyncedSynapse: Xperia XZ1, with Android Pie. I'll test on Oreo later, meanwhile, here are some screenshots showing the issue
screenshot_20190223-192708
screenshot_20190223-192716

poisdeux: Found it. By default drawable resources share their state. To fix this I made the icons for the playlistbar mutable.

app/src/main/java/org/xbmc/kore/jsonrpc/HostConnection.java
poisdeux: Indeed it's more structural. I really don't like the threading logic in the activities and fragments. I think the activities and fragments should only manage the view elements and all else should be placed in other classes. I'm a fan of Clean Architecture and MVVM and if I had the time I would refactor Kore as a whole. But for the time being I would appreciate it if we could move as much logic away from the activities and fragments as possible. If I get to it I will propose a setup and plan to gradually refactor Kore. It will probably be something based on Clean Architecture, and MVVM using Kotlin with Coroutines and RXjava.

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
poisdeux: Indeed. So if I understand correctly I can refactor this and make the `HostConnectionObserver` actually maintain the player state. If `PlaylistFragment` needs the player state it should ask the `HostConnectionObserver`. Correct?

app/src/main/java/org/xbmc/kore/ui/viewgroups/DynamicListView.java
poisdeux: I have not been able to reproduce the NPE. Can you be more specific in what circumstance this occurs?

SyncedSynapse: Forget about this, i can't reproduce it either, but i swear that i saw npes before...

app/src/main/java/org/xbmc/kore/jsonrpc/HostConnection.java
SyncedSynapse: I don't share your dislike of using threads in activities/fragments, as long as it's for a UI reason. Typically the reason to launch background threads is to free the UI thread so that it continues to be responsive, and that's intrinsically an UI issue, so i don't see why it shouldn't be where the UI is managed.
Anyway,  my issue with this method isn't related to that, it's just that i don't think its semantics are consistend with the rest of this class.

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
SyncedSynapse: I wasn't going that far, i was just pointing out that you could use the same enums.
I'm not sure where you're going, but my preference is to not overload `HostConnectionObserver` with much more to do, and if that means that some other classes need to maintain state and not rely on it to do everything, so be it.

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
poisdeux: People connected to Kodi on WiFi will probably not have any problems. On slower connections this may become an issue with huge playlists. We could just disable this feature for HTTP connections? (I'm not a fan of adding another setting to allow users to enable/disable this.)

SyncedSynapse: Why not simply decrease the frequency of this? Creating another runnable might be too much trouble, but you could just check the playlist every 2 runs of this runnable, for instance.

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
poisdeux: It's there to make sure we also check if there are playlists available when we start the playlist fragment. If we would only depend on `onAdd`/`onRemove` then we would only see the available playlists when there is a change in one of the playlists on Kodi.
Another problem with `onAdd`/`onRemove` is that these are called for each item in a playlist when you add or remove multiple items. This makes it a bit harder, as we need to ignore subsequent calls. We could use a delay to prevent calling `checkPlaylist()` many times.

I'm currently thinking of refactoring this as follows:
* On registering a playlist observer call `checkPlaylist()` when using TCP.
* Implement `Playlist.onAdd` and `Playlist.onRemove` to handle changes to the playlist. Adding a delay (2 seconds?), to make sure when we delete a playlist with a trillion items we only call `checkPlaylist()` once after 2 seconds when we received the trillionth `onRemove` notification.

SyncedSynapse: - If you look a couple of lines up from this, you'll see that `checkWhatsPlaying` had the same exact problem, and the solution was calling it only when we were unintialized, not every time. That and a `forceReply` with last results, as done on the remote fragment.

- Use the same solution that was used on `checkWhatsPlaying`: use a static var to control whether a call is already in progress. Ignore race conditions, as they are inconsequential here.

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
poisdeux: Yes, it's now performed when registering the playlist observer in onResume

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
poisdeux: Code has been refactored and now uses onAdd/onRemove for TCP connections. I also implemented getting the playlist in the `tcpCheckerRunnable` only when there is a connection error. If I understand correctly that's the only situation in which the PlaylistFragment may keep showing an incorrect result when the connection is restored.

SyncedSynapse: Just for safety maybe we should also call `checkPlaylist` when `(hostState.lastCallResult == PlayerEventsObserver.PLAYER_NO_RESULT)`

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
```diff
@@ -477,6 +577,37 @@ public void onError(int errorCode, String description) {
         }, checkerHandler);
     }
 
+    private boolean isCheckingWhatsPlaying = false;
```
SyncedSynapse: Isn't `isCheckingPlaylist` a better name?

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
SyncedSynapse: This is the code for the menu clear playlist function, i fail to see the connection to `onResume`.

SyncedSynapse:
I think this is ok to be merged, we just need to discuss my previous comment regarding using the `FixedThreadPool(10)` in host connection. Try this: set Kore to use HTTP; play some video; stop it. The video playlist should still have that item in it. Remove it with the item menu. Does it get remove from the list? Mine isn't, and i suspect it's because of this (look at `PlaylistFragment` Lines 542 for context on my previous comment).

poisdeux:
I understand, but I have not had time to look into it thoroughly. I'm currently working on the HTTP only situation where the playlist does not get cleared on Kore when it is cleared on Kodi. 
The comments left open are those that I still need to address or want to look at more closely.

app/src/main/java/org/xbmc/kore/host/actions/GetPlaylist.java
```diff
@@ -170,5 +170,26 @@ private GetPlaylistResult(int playlistId, String type, List<ListType.ItemsAll> i
...
+            if ((! (obj instanceof GetPlaylistResult)) )
+                return false;
+
+            GetPlaylistResult other = (GetPlaylistResult) obj;
+
+            if (id != other.id || ! type.equals(other.type))
+                return false;
+
+            for(int i = 0; i < items.size(); i++) {
+                if (items.get(i).equals(other.items.get(i)))
```
SyncedSynapse: Can you please check if this is right. Seems to be missing a !. 
It's also assuming that the items are in the same order, which should be validated.

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
```diff
@@ -592,9 +593,29 @@ public void onSuccess(ArrayList<GetPlaylist.GetPlaylistResult> result) {
                 if (result.isEmpty())
                     return;
 
-                for (PlaylistEventsObserver observer : playlistEventsObservers) {
-                    observer.playlistsAvailable(result);
+                boolean isPlaylistChanged = false;
```
SyncedSynapse: I think this code could use a comment explaining what it is doing, as it is not immediately obvious.

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
```diff
@@ -592,9 +593,29 @@ public void onSuccess(ArrayList<GetPlaylist.GetPlaylistResult> result) {
...
+                for (GetPlaylist.GetPlaylistResult getPlaylistResult : result) {
+                    if (! isPlaylistChanged &&
```
SyncedSynapse: The `isPlaylistChange` check here is not necessary (it just avoids another assignment, but if makes reading the code more difficult, as you need to keep another variable in mind, which is burdensome and unnecessary).

app/src/main/java/org/xbmc/kore/jsonrpc/type/ListType.java
```diff
@@ -284,6 +284,17 @@ public ItemBase(JsonNode node) {
...
+
+        @Override
+        public boolean equals(Object obj) {
+            if (obj == this)
+                return true;
+
+            if (! (obj instanceof ItemBase))
+                return false;
+
+            return ((ItemBase) obj).id == this.id;
```
SyncedSynapse: This doesn't feel totally right.
What guarantees do we have that this id is unique? If we have them (or if we don't have them but are assuming it), we should explicitly make a comment documenting that.

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
```diff
@@ -592,9 +593,29 @@ public void onSuccess(ArrayList<GetPlaylist.GetPlaylistResult> result) {
...
+                        ! prevGetPlaylistResults.contains(getPlaylistResult)) {
+                        isPlaylistChanged = true;
+                    }
+
+                    prevGetPlaylistResults.remove(getPlaylistResult);
+                }
+
+                if (isPlaylistChanged) {
```
SyncedSynapse: Can't we just call `playlistsAvailable ` on all playlists we got? Is this just an optimization?
Not needing this would obliviate the need for the `equals` method. 
Seems to me that we just need the set difference between the playlists ids on preResults and results, so that we can call `playlistOnClear` on them, and if so this should be more explicit.

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
poisdeux: Ah my bad. As we now handle the onClear notification it is not needed here anymore. Although I do still need to fix handling HTTP only connections. I'm thinking of just calling `playlistOnClear` here and not wait for the HTTP checker to finally call the `playlistOnClear`.

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
poisdeux: Indeed it is.

app/src/main/java/org/xbmc/kore/host/actions/GetPlaylist.java
poisdeux: Thx, negation added.
If I'm correct playlists with the same items but in a different order should be considered different to handle the situation when the user reordered the playlist items on Kodi.

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
poisdeux: done.

app/src/main/java/org/xbmc/kore/host/HostConnectionObserver.java
poisdeux: Yes, it's an optimisation. For HTTP connections there is a chance we need to check a lot of items each time the checker is called. But I've been thinking about this a bit more and as it only benefits the case where the user changes the playlist, it benefits only a very small portion of the amount of times the `checkPlaylist` method is called. So this logic can be removed clearing up the code considerably.

app/src/main/java/org/xbmc/kore/jsonrpc/type/ListType.java
poisdeux: Code has been removed as it is not needed anymore.

app/src/main/java/org/xbmc/kore/jsonrpc/HostConnection.java
poisdeux: I think I fixed any possible racing condition regarding the playlists by reducing the calls to `refreshPlaylist` to a single call when the playlist has changed. We now fully depend on observing playlist notifications to handle playlists.

SyncedSynapse:
Looks good. Sorry for not being able to review this earlier.

PR #619, opened 2019-02-16
Fix adaptive icon foreground dimensions

ssaqua:
Hello, I found the foreground layer drawable is incorrectly sized which causes display issues in certain parts of the system settings. The correct size should be 72dp as per the [adaptive icon documentation](https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive).

Here are some screenshots to show the effects of the change.

24dp vs 72dp on a Samsung Note 8
image

24dp vs 72dp inside Samsung Secure Folder
image

24dp on API 27 emulator
image

72dp on API 27 emulator
image

### DIFF
--- a/app/src/main/res/drawable-v21/ic_launcher_foreground.xml
+++ b/app/src/main/res/drawable-v21/ic_launcher_foreground.xml
@@ -1,6 +1,6 @@
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
-    android:width="24dp"
-    android:height="24dp"
+    android:width="72dp"
+    android:height="72dp"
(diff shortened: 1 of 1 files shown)


poisdeux:
Thanks. Never noticed this on my devices... I need a new phone ;).

SyncedSynapse:
Yeah, thanks for spotting this - i had wondered why the icon wasn't showing up right in the share panel, but didn't look into it.

I'm going to merge this as is, but i think that the correct dimensions should be 108dp. But for that to work, the vector drawing needed to be restricted to the 72dp area inside that, and i'm not sure it is.

PR #627, opened 2019-03-04
Update Polish translations

Valdnet:
New translation - shopping


SyncedSynapse:
Those strings aren't used in Kore. They are commented out in the main file, but they should be removed, as they are leftovers from a long time ago.
Given that, there's no sense in translating them. Sorry for your trouble.

Valdnet:
OK. Please delete so that they do not mislead other translators.
Thanks 😉

Issue #599, opened 2018-11-29
[bug/feat] Trunicated title information?


momoe:
To amend Fix https://github.com/xbmc/Kore/commit/da90f92ddcb6743249ca0387b88989c7b298524d instead of trunicating long titles, please have the main remote title scroll the title back/forth horizontally while a video is playing. As some local media does have very long title information, especially episode numbers in a series is impossible to read. 

Please provide some way to either view the entire title, or scroll the title information that will make key information be presentable such as the episode numbers. 

-Thank you.

Isayso:
I would also like to have a smaller font option or wordwrap for the playlist view for the same reason. I'm guessing the title most of the time now. YouTube titles are often very long and rotating to landscape is sometimes inkonvenient. ;) 
Thanks a lot, its great work.

Isayso:
I modified one grid_item  with maxlines and smaller font and deleted ellipsize. 
I think this is a better solution because the important information is often at the end.
Thanks!

![kore_mods](https://user-images.githubusercontent.com/35309607/67385959-60d44600-f594-11e9-98d0-18539f886697.jpg)

PR #635, opened 2019-04-15
Implement scrolling title on now playing and general media info screens

ssaqua:
Hello 😃 following on from #612 for issue #599

When the title length would exceed maxLines, convert it instead to a single line ellipsized title which scrolls horizontally when focused. Titles that would fit normally on <= 2 lines are unaffected.

Found the marquee doesn't take effect unless `setHorizontallyScrolling()` is directly called which is why it is there instead of in styles.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/NowPlayingFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/NowPlayingFragment.java
@@ -20,7 +20,6 @@ import android.content.res.Resources;
 import android.os.Bundle;
 import android.os.Handler;
 import android.support.v4.app.Fragment;
-import android.support.v4.widget.TextViewCompat;
 import android.text.TextUtils;
 import android.util.DisplayMetrics;
 import android.view.LayoutInflater;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/xbmc/kore/ui/sections/remote/NowPlayingFragment.java
```diff
@@ -752,6 +753,21 @@ private void setNowPlayingInfo(PlayerType.GetActivePlayersReturnType getActivePl
         }
 
         mediaTitle.setText(title);
+        mediaTitle.post(new Runnable() {
+            @Override
+            public void run() {
+                int lines = mediaTitle.getLineCount();
+                int maxLines = TextViewCompat.getMaxLines(mediaTitle);
+                if (lines > maxLines) {
+                    mediaTitle.setHorizontallyScrolling(true);
```
poisdeux: This causes the textview to be a single line for text that needs more than max lines even when not scrolling. How about setting it based on the selected state of the `TextView`?

```
                if (lines > maxLines) {
                    mediaTitle.setClickable(true);
                    mediaTitle.setOnClickListener(new View.OnClickListener() {
                        @Override
                        public void onClick(View v) {
                            if (v.isSelected()) {
                                ((TextView) v).setMaxLines(1);
                                mediaTitle.setHorizontallyScrolling(true);
                            } else {
                                ((TextView) v).setMaxLines(2);
                                mediaTitle.setHorizontallyScrolling(false);
                            }
                            v.setSelected(!v.isSelected()); }
                    });
                }
            }
        });
```

ssaqua: This causes the layout (posters, buttons, descriptions, etc) to reposition as the `TextView` switches between 1 and 2 lines height. Is that a concern?

Otherwise am happy to update. I found I needed a few more changes to get better control of the ellipsis (the ellipsis is presumably hidden when `lines > maxLines` and the mode is marquee by default as it would be generated on the hidden last line):

```java
// default ellipsize style <item name="android:ellipsize">end</item>
if (lines == maxLines && mediaTitle.getLayout().getEllipsisCount(lines - 1) > 0) {
    mediaTitle.setClickable(true);
    mediaTitle.setOnClickListener(new View.OnClickListener() {
        @Override
        public void onClick(View v) {
            v.setSelected(!v.isSelected());
            TextUtils.TruncateAt ellipsize;
            if (v.isSelected()) {
                ellipsize = TextUtils.TruncateAt.MARQUEE;
            } else {
                ellipsize = TextUtils.TruncateAt.END;
            }
            mediaTitle.setEllipsize(ellipsize);
            mediaTitle.setHorizontallyScrolling(v.isSelected());
        }
    });
}
```
```

poisdeux: I don't mind that it jumps now. It's only when the user clicks the title to make it scroll.
And nice touch changing the ellipsize to END to make the user aware there is more text. 
There is a small problem here that the END ellipsize is only shown when the user switches back from marquee to multiline. If I remember correctly you probably have to add it programmatically as setting the ellipsize to END in the style will brake the marquee functionality.

ssaqua: Have pushed an update, hopefully I haven't misunderstood the desired effect.

poisdeux: Works great! @SyncedSynapse any objections?

Regarding the refactoring. It would be nice to put the runnable in a separate method. Preferably in the [UIUtils](https://github.com/xbmc/Kore/blob/4dda3ac6e85b84dc2205950eea9b4269edf75c97/app/src/main/java/org/xbmc/kore/utils/UIUtils.java) class. That would also reduce the code duplication, as I believe both runnables are the same.

SyncedSynapse:
I'm having some issues with my main pc, so i'm not able to test this right now. @poisdeux if it's good by you, please merge it.

PR #638, opened 2019-05-11
Fix ANR in service launch

SyncedSynapse:
Fix ANRs reported on Play Store when launching the observers service.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/service/ConnectionObserversManagerService.java
+++ b/app/src/main/java/org/xbmc/kore/service/ConnectionObserversManagerService.java
@@ -49,13 +49,13 @@ public class ConnectionObserversManagerService extends Service
         implements HostConnectionObserver.PlayerEventsObserver {
     public static final String TAG = LogUtils.makeLogTag(ConnectionObserversManagerService.class);
 
-    private HostConnectionObserver mHostConnectionObserver = null;
+    private HostConnectionObserver hostConnectionObserver = null;
 
-    private List<HostConnectionObserver.PlayerEventsObserver> mConnectionObservers = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/xbmc/kore/service/ConnectionObserversManagerService.java
```diff
@@ -67,35 +67,31 @@ public void onCreate() {
...
-        HostConnectionObserver connectionObserver = HostManager.getInstance(this).getHostConnectionObserver();
+        HostConnectionObserver connectionObserver = HostManager.getInstance(this)
+                .getHostConnectionObserver();
 
         // If we are already initialized and the same host, exit
         if (mHostConnectionObserver == connectionObserver) {
             LogUtils.LOGD(TAG, "Already initialized");
-            return START_STICKY;
+            stopForeground(true);
+            stopSelf();
```
poisdeux: I fail to see the logic here. If the service is already running for the HostConnectionObserver we now stop it?

SyncedSynapse: You're right. My bad. Ignore this please.

SyncedSynapse:
@poisdeux Should be more robust now.

I'm not convinced that this will solve the ANRs reported on the Play Store, but at least we tried (the ANRs seem to be caused by something inside Android's core instead of this code). Curiously they are more prevalent on Samsung devices, so i suspect that it's one more misguided attempt of Samsung to improve battery life, with side-benefits ...

poisdeux:
Looks good. I'm not sure why you are not convinced this will fix it. In the previous version it was possible the service was started without calling `startForeground`, resulting in the reported error. Now `startForeground` is always called.

PR #639, opened 2019-05-13
Fix Jenkins build to upload the correctly signed apk

SyncedSynapse:


tools/jenkins/buildsteps/package
```diff
@@ -52,4 +52,4 @@ function getBuildRevDateStr ()
 #rename for upload
 #e.x. kore-20130314-8c2fb31.apk.
 UPLOAD_FILENAME="kore-$(getBuildRevDateStr).apk"
-cd $WORKSPACE;mv app/build/outputs/apk/release/*.apk $UPLOAD_FILENAME
+cd $WORKSPACE;mv kore-release.apk $UPLOAD_FILENAME
```
poisdeux: Are you sure this is correct? If I run `./gradlew assembleRelease` as executed by the `build-release` script the apk is located at `app/build/outputs/apk/release/`.
If I generate a signed apk through Android Studio it is placed under `app/release/`.

SyncedSynapse: Can't test it, but i'm fairly sure it is correct. Guess you didn't receive Memphiz reply to the email i sent (apksign is run before this, and you can check the logs to see the correct paths):

> Hi,
> 
> https://github.com/xbmc/Kore/blob/master/tools/jenkins/buildsteps/package
> 
> Last line does the move while using the wrong source apk (it uses the unsigned one).
> It needs to use the $workspace/kore-release.apk instead which is generated by the sign script.
> 
> For reference see the log output here: https://jenkins.kodi.tv/view/Android/job/Kore/3911/console
> 
> ++ /Users/Shared/jenkins/workspace/Kore/../../android-dev/signing/doreleasesign.sh Signing... /Users/Shared/jenkins/workspace/Kore/kore-release.apk created
> 
> ++ mv app/build/outputs/apk/release/app-release-unsigned.apk kore-20190510-fd70552-master.apk
> 
> That is also the reason why it always uploads 2 files - one with the correct filename but unsigned and one signed with the name kore-release.apk which gets overwritten every time.
> 
> This broke with those changes in the output Dir a couple of month ago I think.
>

PR #640, opened 2019-05-16
Updated gradle and compile and target SDK version

poisdeux:


poisdeux:
Upgrading somehow causes the unit tests to fail as the robolectric classes can not be found. We will probably need to upgrade to Robolectric version 4. I'll update this PR when I have fixed it.

app/src/test/java/org/xbmc/kore/provider/mediaprovider/AlbumsTest.java
```diff
@@ -57,11 +58,11 @@ public void queryAlbumTest() {
     }
 
     @Test
-    public void queryAlbumsForArtistTest() {
+    public void queryAlbumsForArtistTest() throws Exception {
         Uri uri = MediaContract.AlbumArtists.buildAlbumsForArtistListUri(hostInfo.getId(),
                                                                          TestValues.Artist.artistId);
 
-        Cursor cursor = shadowContentResolver.query(uri, TestValues.Album.PROJECTION, null, null, null);
+        Cursor cursor = client.query(uri, TestValues.Album.PROJECTION, null, null, null);
```
SyncedSynapse: Guess this needs to be done in `GenresTest`, `SongsTest` and `ArtistsTest` also.

poisdeux: Yeah, still working on it. I didn't have time left to do it yesterday.

poisdeux: All green!

app/src/main/res/values/dimens.xml
```diff
@@ -53,6 +53,9 @@
     <dimen name="tab_strip_height">48dp</dimen>
     <dimen name="view_pager_indicator_height">12dp</dimen>
 
+    <dimen name="remote_icon_size">67dp</dimen>
+    <dimen name="remote_size">202dp</dimen>
```
SyncedSynapse: We should align stuff at 8dp, 4dp if necessary. Anyway, i think this size is always overriden isn't it?

poisdeux: Why 8 or 4 and not 2?
I haven't checked in the source code but from what I understand from the theory it should indeed never match.

SyncedSynapse: It's what the guidelines recommend: https://material.io/design/layout/understanding-layout.html#usage

PR #642, opened 2019-05-28
Fix issues with connection threads

SyncedSynapse:
This PR fixes some issues with connections and threading. Specifically, the  changes in #618 which introduced threading in `HostConnection`, also introduced some issues. To fix them, the following changes were made:
- A specific TCP listener thread is used and manually started, instead of using one of the threads in the pool. The TCP listener thread is a long lived thread that should always be running (as long as the connection is through TCP) and be blocked listening on the TCP socket, so it shouldn't be managed in a pool, where, theoretically, it can be paused and reused;
- Changed the number of threads to 5. We need 2 threads to run correctly, so we shouldn't create much more than this, otherwise we can overwhelm some Kodi hardware;
- Had to sprinkle some `synchronized`'s to avoid race conditions. For instance, through a TCP connection, as soon as Kore is opened (from a clean state) it will call at least `GetActivePlayers`, `GetProperties`, `Ping`. Given that the TCP socket isn't set up yet, each of these calls has the potential to create a socket (and a TCP listener thread), so we would open 3 or more sockets, 2 of which would be leaked because we would use only 1 from then on. A `synchronized` in `executeThroughTcp` prevents this. The others prevent similar issues.

Additionally:
- Tweaked the playlist fetching code, so that most of it happens in `HostConnectionObserver` and when PlaylistFragment is notified of changes, it already gets the playlist data. This somewhat simplifies the code, and makes it more consistent with the Player Observer code;
- Change `EventServerConnection` to accept a Handler on which to post the result of the connection, so that the caller can control on which thread the result is called;
- Calls to the various `RegisterObserver` loose the reply immediately parameter, as it was always true.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/eventclient/EventServerConnection.java
+++ b/app/src/main/java/org/xbmc/kore/eventclient/EventServerConnection.java
@@ -81,13 +81,17 @@ public class EventServerConnection {
     /**
      * Constructor. Starts the thread that keeps the connection alive. Make sure to call quit() when done.
      * @param hostInfo Host to connect to
+     * @param callback Callback to call with the connection result
+     * @param callbackHandler Handler on which to call the callback
      */
-    public EventServerConnection(final HostInfo hostInfo, final EventServerConnectionCallback callback) {
+    public EventServerConnection(final HostInfo hostInfo,
+                                 final EventServerConnectionCallback callback,
+                                 final Handler callbackHandler) {
         this.hostInfo = hostInfo;
 
         LogUtils.LOGD(TAG, "Starting EventServer Thread");
         // Handler thread that will keep pinging and send the requests to Kodi
-        handlerThread = new HandlerThread("EventServerConnection", Process.THREAD_PRIORITY_DEFAULT);
+        handlerThread = new HandlerThread("EventServerConnection", Process.THREAD_PRIORITY_BACKGROUND);
         handlerThread.start();
 
         // Get the HandlerThread's Looper and use it for our Handler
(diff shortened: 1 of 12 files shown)


SyncedSynapse:
@poisdeux When you have the time, i would appreciate if you could make some tests on this, specifically to make sure that the playlist fetching code is working in all situations. I tried to cover them all, but you probably have more knowledge about edge cases.

poisdeux:
Only had a look at the code. I will do some testing later on. The code looks good.

There is only one edge case I can currently think of. I didn't bother implementing it as I found it to be a far stretched case that might not be an issue at all. So was thinking on only solving it when users start complaining :). 
The edge case has to do with adding a large amount of items at once. As we also listen to `Playlist.OnAdd` and `Playlist.OnRemove` events we might want to add some logic to wait until we have received the last event. We now start fetching the new playlist at the first call which might result in only a partial playlist as the other items are still being added. We could post delay the call to `checkPlaylist()` and if we get a subsequent `Playlist.OnAdd` call before the delay passes we remove the previous call and post a new delayed call. This would make sure that if someone adds a zillion items at once we wait until the last item has been added before calling `checkPlaylist()`. A drawback is that it might take a while before the playlist is shown in Kore. 
(Of course the ideal solution would be to implement pagination)

SyncedSynapse:
Yeah, but maybe that's not an issue at all, depending on how Kodi adds items to its playlist. Maybe It adds all of them at once,  and them sequentially notifies clients with `Playlist.OnAdd` for each one. So, maybe the first `checkPlaylist` is enough to get all the added items. Anyway, i also think that's something for later.
I was referring more to test cases that the previous code handled that might turn out wrong with the modifications of this PR.

SyncedSynapse:
@poisdeux Could you take a look at this PR when you have the time.

I would like to do a release soon, and the issues with threads should be fixed.

poisdeux:
@SyncedSynapse Sorry, but I've been rather busy lately. Hopefully I can find some time to look at this and #658 in about two weeks.

poisdeux:
@SyncedSynapse I tried to simply rebase this branch on top of #658 to see if the tests run and succeed. Unfortunately they do not and as there has been a lot of changes on `master` since this PR was created it is a bit difficult to see why things are failing. Can you rebase this branch on the the latest commit on `master` and check if your changes still work?

SyncedSynapse:
@poisdeux This PR applies cleanly on top of the current `master`, and works as intended, afaict. #658 doesn't apply cleanly because of the changes to `PlaylistFragment`, which honestly i'm not a fan of, as they seem a particularly complex way of solving a race issue that only occurs in testing. Serialing the access would be simpler and not have impacts on performance as it only applies to tests.

poisdeux:
Woohoo! All green!

PR #658, opened 2019-07-04
Implemented UI tests for PlaylistFragment

poisdeux:
Fixed racing condition with refresh playlist being called for each
Playlist.OnAdd notification when adding multiple items/playlists at
once.

### DIFF
--- a/app/src/androidTest/java/org/xbmc/kore/testhelpers/EspressoTestUtils.java
+++ b/app/src/androidTest/java/org/xbmc/kore/testhelpers/EspressoTestUtils.java
@@ -104,6 +104,18 @@ public class EspressoTestUtils {
     }
 
     /**
+     * Clicks the button with given resourceId and checks if the
+     * button is displayed. As we occasionally use the same identifiers
+     * in multiple fragments, we need to check if it is visible as well
+     * to prevent Espresso from finding multiple views that match the
+     * resource identifier.
+     * @param resourceId
+     */
+    public static void clickButton(int resourceId) {
+        onView(allOf(withId(resourceId), isDisplayed())).perform(click());
+    }
+
+    /**
      * Clicks on the search menu item and enters the given search query
      * @param activity
      * @param query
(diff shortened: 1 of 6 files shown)


SyncedSynapse:
@poisdeux Notice that #642 also changes `PlaylistFragment.java`. With those changes does this issue still occurs?

Where exactly is the race condition you're fixing here? In `refreshPlaylist` on `PlaylistFragment.java`? I see that there's the potential for one there, but wouldn't a simple `lock` on some object be a simpler fix (I'll also post a comment on the code)?

app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
```diff
@@ -223,32 +224,38 @@ public boolean onOptionsItemSelected(MenuItem item) {
         return super.onOptionsItemSelected(item);
     }
 
-    private boolean refreshingPlaylist;
-    private void refreshPlaylist(GetPlaylist getPlaylist) {
-        if (refreshingPlaylist)
```
SyncedSynapse: Protecting this check and assignment with a `lock` doesn't also fix the possible race condition?

poisdeux: No, the problem occurs when the playlist tests add multiple playlists in the setup fase before each test. When I implemented #618 I only implemented tests using a single playlist. Afterwards I added some additional tests to test more scenario's which caused the race condition to arise. Probably not something that will/can happen in real usage but is at least needed to make the playlist tests work.

poisdeux:
> @poisdeux Notice that #642 also changes `PlaylistFragment.java`. With those changes does this issue still occurs?

I just rebased #642 on this and unfortunately the tests fail. Have not had time to look in detail why it fails. I only tested with `PlaylistFragment.java` from #642 as the version in this PR uses `PlaylistEventsObserver.playlistChanged()` and I do not have time right now to fix this and see if the `PlaylistFragment.java` from this PR does work with your changes from #642.

poisdeux:
Also tested #642 merged with this PR and with this PR's version of `PlaylistFragment.java` and the tests still fail.

poisdeux:
I finally had some time to look into the race condition and I was able to come up with another solution that does not require the more complex locking mechanism in `PlaylistFragment`.
I have not had time yet to see if #642 no passes the tests as well.

PR #659, opened 2019-07-05
Add sort by year section thumb text

wbarnard:
When album, movie and tv show lists are sorted by year, use the 2 digit year as the section thumb text rather than the title first letter.


SyncedSynapse:
Nice find. I'll only be able to look into this next week, but i will.

PR #660, opened 2019-07-08
Fix href tag closing

ssaqua:
I found a problem with displaying these two HTML strings during setup of a new host. The missing closing quotes would cause any text from the \<a> tag onward to not be displayable.

Somehow this wasn't a problem on support lib version 26.1.0 and gradle 3.1.2 before #640. On the current versions, adding the closing quotes is enough to get it to parse correctly.

While we are looking at this.. `Html.fromHtml(String source)` is deprecated since N. Would it be worthwhile to add a compatibility method like in [androidx HtmlCompat](https://android.googlesource.com/platform/frameworks/support/+/f7528a970ad7f2f099b4dc1397084cc388c9193c/core/src/main/java/androidx/core/text/HtmlCompat.java#145)? I guess we would just pass the `FROM_HTML_MODE_LEGACY` flag anyway so doesn't make any difference for now.

image

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -79,14 +79,14 @@
     <string name="wizard_welcome">Welcome</string>
     <string name="wizard_welcome_message"><![CDATA[
     Let\'s start by adding a media center. Make sure your Kodi/XBMC is running, properly configured and on the same network as your device.<br/><br/>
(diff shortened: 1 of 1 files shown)


SyncedSynapse:
Yeah, and this is present in other languages also. I'll try to fix them.

Regarding `Html.fromHtml`, i guess that the we should just use the version with the `flags` parameter.

PR #664, opened 2019-07-11
Fixes crash due to missing title in PlaylistFragment

poisdeux:


poisdeux:
Oops sorry didn't see there is already a PR that fixes this.

PR #671, opened 2019-09-03
Support sharing SoundCloud tracks

jaylinski:
### Description

This adds support for sharing SoundCloud tracks. :notes: 

### Notes

I noticed that the other supported plugins have tests. I guess I should also add tests for this?


SyncedSynapse:
It's good by me. @poisdeux  any comments?

poisdeux:
@SyncedSynapse Nope. Works and looks good!

PR #678, opened 2019-10-13
Fix issue with showing multiple bookmarked addons

nickgrim:
When multiple addons are bookmarked, the bookmarked tabs all show the content of the last addon, due to reusing a `Bundle` when setting up the tabs.

This code fixes that by cloning the `Bundle` and passing a copy to each `addTab` call.


SyncedSynapse:
Looks good. Thanks!

Issue #585, opened 2018-10-09
[Feature request] Change UI language


NeroRome:
It seems to be not possible to change the language of the app Kore.
Since I'm using Kodi in English but the system language of my Android device is German, also Kore is in German (and I don't like the translation).

I'd like to have a language selection as feature request.

PR #680, opened 2019-10-19
Implemented changing language in-app

poisdeux:
This allows users to use a different supported locale than the one that
matches the device's locale.

Implements feature request #585

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -21,6 +21,12 @@ android {
         versionCode 25
         versionName = getVersionName()
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+
+        def supportedLocales = ["en", "ast", "ru", "it", "ca", "cs", "zh-CN", "ja", "pt", "pt-BR",
+                                "pl", "sl", "sk", "lt", "eu", "iw", "fr", "es", "hr", "hu", "nl",
+                                "bg", "de", "ko", "uk"]
+        buildConfigField "String[]", "SUPPORTED_LOCALES", "new String[]{\""+
+                supportedLocales.join("\",\"")+"\"}"
     }
 
     File keystoreFile = file('keystore.properties')
--- a/app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
@@ -15,8 +15,8 @@
  */
 package org.xbmc.kore.ui.sections.settings;
 
-import android.annotation.TargetApi;
 import android.Manifest;
+import android.annotation.TargetApi;
 import android.app.NotificationManager;
 import android.content.Context;
 import android.content.Intent;
@@ -33,6 +33,7 @@ import android.support.v7.preference.PreferenceFragmentCompat;
 import android.support.v7.preference.TwoStatePreference;
 import android.widget.Toast;
 
+import org.xbmc.kore.BuildConfig;
 import org.xbmc.kore.R;
 import org.xbmc.kore.Settings;
 import org.xbmc.kore.host.HostManager;
@@ -43,8 +44,12 @@ import org.xbmc.kore.utils.UIUtils;
 import org.xbmc.kore.utils.Utils;
 
 import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.Locale;
 
 import static org.xbmc.kore.service.NotificationObserver.NOTIFICATION_ID;
+import static org.xbmc.kore.utils.Utils.getLocale;
 
 /**
  * Simple fragment to display preferences screen
@@ -60,7 +65,7 @@ public class SettingsFragment extends PreferenceFragmentCompat
     public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
 
         // Load the preferences from an XML resource
-        addPreferencesFromResource(R.xml.preferences);
+        setPreferencesFromResource(R.xml.preferences, null);
 
         // Get the preference for side menu items and change its Id to include
         // the current host
@@ -105,6 +110,17 @@ public class SettingsFragment extends PreferenceFragmentCompat
         }
 
         setupPreferences();
+
+        ListPreference languagePref = (ListPreference) findPreference(Settings.KEY_PREF_LANGUAGE);
+        Locale currentLocale = getCurrentLocale();
+        languagePref.setSummary(currentLocale.getDisplayLanguage(currentLocale));
+        languagePref.setOnPreferenceClickListener(new Preference.OnPreferenceClickListener() {
+            @Override
+            public boolean onPreferenceClick(Preference preference) {
+                setupLanguagePreference((ListPreference) preference);
+                return true;
+            }
+        });
     }
 
     @Override
@@ -130,15 +146,15 @@ public class SettingsFragment extends PreferenceFragmentCompat
         setupPreferences();
 
         if (key.equals(Settings.KEY_PREF_THEME) || key.equals(Settings.getNavDrawerItemsPrefKey(hostId))
-                || key.equals((Settings.getRemoteBarItemsPrefKey(hostId)))) {
+            || key.equals((Settings.getRemoteBarItemsPrefKey(hostId)))) {
             // Explicitly clear cache of resource ids that is maintained in the activity
             UIUtils.playPauseIconsLoaded = false;
 
             // restart to apply new theme (actually build an entirely new task stack)
             TaskStackBuilder.create(getActivity())
-                    .addNextIntent(new Intent(getActivity(), RemoteActivity.class))
-                    .addNextIntent(new Intent(getActivity(), SettingsActivity.class))
-                    .startActivities();
+                            .addNextIntent(new Intent(getActivity(), RemoteActivity.class))
+                            .addNextIntent(new Intent(getActivity(), SettingsActivity.class))
+                            .startActivities();
         }
 
         // If the pause during call is selected, make sure we have permission to read phone state
@@ -203,7 +219,7 @@ public class SettingsFragment extends PreferenceFragmentCompat
         String nameAndVersion = getActivity().getString(R.string.app_name);
         try {
             nameAndVersion += " " +
-                    getActivity().getPackageManager().getPackageInfo(getActivity().getPackageName(), 0).versionName;
+                              getActivity().getPackageManager().getPackageInfo(getActivity().getPackageName(), 0).versionName;
         } catch (PackageManager.NameNotFoundException exc) {
         }
         Preference aboutPreference = findPreference(Settings.KEY_PREF_ABOUT);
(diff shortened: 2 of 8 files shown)


app/build.gradle
```diff
@@ -21,6 +21,12 @@ android {
         versionCode 25
         versionName = getVersionName()
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+
+        def supportedLocales = ["en", "ast", "ru", "it", "ca", "cs", "zh_CN", "ja", "pt", "pl", "sl", "sk",
+                                "lt", "eu", "iw", "fr", "es", "hr", "hu", "nl", "bg", "de", "ko",
+                                "pt", "pt_BR", "uk"]
```
SyncedSynapse: pt is duplicated

app/src/main/java/org/xbmc/kore/utils/Utils.java
```diff
@@ -250,4 +253,28 @@ public void onError(int errorCode, String description) {
...
+    }
+
+    public static Locale getLocale(String localeName) {
+        Locale locale;
+        String[] languageAndRegion = localeName.split("_", 2);
+        if (languageAndRegion.length > 1) {
+            locale = new Locale(languageAndRegion[0], languageAndRegion[1]);
+        } else {
+            locale = new Locale(localeName);
+        }
```
SyncedSynapse: There's a third constructor with a third parameter - variation. Not sure if it's widely used though.

SyncedSynapse:
In my phone, the languages appear somewhat strangely. They appear in a mix of lower case and upper case, and some are right adjusted. I know that you're using the Locale.GetDisplayName, which should be the right one, but it's very different from what i get in the device's language selector.

It would also be nice if:
- The list would be sorted;
- The current selected locale be displayed in the settings line, below the "Change language". And probably that should just be called "Language", as i believe it's what's called in the default device's settings.

And maybe there should be an option called "Default" that would be the device's default locale, but maybe this is superfluous.

Can't finish this right now, i'll take another look soon.

app/build.gradle
poisdeux: fixed

app/src/main/java/org/xbmc/kore/utils/Utils.java
poisdeux: I have no idea where variant is used for and where to get it. So I suggest we leave it out.

poisdeux:
Updated the PR. 
I have not implemented a "Default" to have the app use the system locale as I'm not sure what the use case would be.

SyncedSynapse:
Much better, but there are still some issues. Dealing with Locales isn't simple, and i believe that some of the code simplifies it too much, leading to subtle bugs that can pass unnoticed.

For instance, as is the current code doesn't work with language_country variants (for instance pt_BR, i'll point out where it fails).

My point is that we should use the default framework functions to deal with locale, namely:
https://developer.android.com/reference/java/util/Locale.html#forLanguageTag(java.lang.String)
https://developer.android.com/reference/java/util/Locale.html#toLanguageTag()

Unfortunately, these are v21+, so some fallback needs to happen for older versions, but i believe we should use those.

Furthermore, you'll notice that sorting the entries won't really make them appear sorted. This is because you're sorting the entries based on the device's locale (not the locale currently selected for the app), and you're presenting the locale according to it's name in that locale (hence we get chinese/japanese/arabic characters). There might be some reasoning to presenting the locale name localized to that locale, but i think that it introduces some inconsistencies that we should avoid. I propose that the sort and display are done in the locale currently selected. I'll point that out in the code.

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
```diff
@@ -217,4 +233,76 @@ public boolean onPreferenceClick(Preference preference) {
             }
         });
     }
+
+    private void setupLanguagePreference(final ListPreference languagePref) {
+        Locale[] locales = getLocales(BuildConfig.SUPPORTED_LOCALES);
+
```
SyncedSynapse: `Locale currentLocale = getCurrentLocale();` here

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
```diff
@@ -217,4 +233,76 @@ public boolean onPreferenceClick(Preference preference) {
...
+        Arrays.sort(locales, new Comparator<Locale>() {
+            @Override
+            public int compare(Locale o1, Locale o2) {
+                return o1.getDisplayName().compareToIgnoreCase(o2.getDisplayName());
```
SyncedSynapse: Should be `getDisplayName(currentLocale);`

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
```diff
@@ -217,4 +233,76 @@ public boolean onPreferenceClick(Preference preference) {
...
+            }
+        });
+
+        String[] displayNames = new String[locales.length];
+        String[] entryValues = new String[locales.length];
+        for(int index = 0; index < locales.length; index++) {
+            Locale locale = locales[index];
+            displayNames[index] = locale.getDisplayName(locale);
```
SyncedSynapse: Should be `getDisplayName(currentLocale);`

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
```diff
@@ -217,4 +233,76 @@ public boolean onPreferenceClick(Preference preference) {
...
+            entryValues[index] = getLanguageCountryCode(locale);
```
SyncedSynapse: Should pass `currentLocale` to the function

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
```diff
@@ -217,4 +233,76 @@ public boolean onPreferenceClick(Preference preference) {
...
+        }
+
+        languagePref.setValue(getLanguageCountryCode(getCurrentLocale()));
```
SyncedSynapse: Use `currentLocale`

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
```diff
@@ -217,4 +233,76 @@ public boolean onPreferenceClick(Preference preference) {
...
+                updatePreferredLanguage(o.toString());
+                return true;
+            }
+        });
+    }
+
+    private String getLanguageCountryCode(Locale locale) {
+        String result = locale.getLanguage();
+        if (!locale.getCountry().isEmpty()) {
+            result += locale.getCountry();
```
SyncedSynapse: Missing the `-`... ex: pt-BR

Now that i look into it, is it a `-` or a `_`? I believe it's a `-`. One more reason to use the default functions

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
```diff
@@ -217,4 +233,76 @@ public boolean onPreferenceClick(Preference preference) {
...
+            @Override
+            public boolean onPreferenceChange(Preference preference, Object o) {
+                languagePref.setValue(o.toString());
+                updatePreferredLanguage(o.toString());
+                return true;
+            }
+        });
+    }
+
+    private String getLanguageCountryCode(Locale locale) {
```
SyncedSynapse: https://developer.android.com/reference/java/util/Locale.html#toLanguageTag()

in v21+

app/src/main/java/org/xbmc/kore/utils/Utils.java
```diff
@@ -250,4 +253,28 @@ public void onError(int errorCode, String description) {
...
+
+        Resources resources = context.getResources();
+
+        Configuration configuration = resources.getConfiguration();
+        configuration.locale = locale;
+
+        resources.updateConfiguration(configuration, resources.getDisplayMetrics());
+    }
+
+    public static Locale getLocale(String localeName) {
```
SyncedSynapse: https://developer.android.com/reference/java/util/Locale.html#forLanguageTag(java.lang.String)

app/src/main/java/org/xbmc/kore/utils/Utils.java
```diff
@@ -250,4 +253,28 @@ public void onError(int errorCode, String description) {
...
+        Locale locale;
+        String[] languageAndRegion = localeName.split("_", 2);
```
SyncedSynapse: Ok, so this is where the `_` comes from...

Don't think it's a good idea to use `_`, when the locale specification uses `-`.

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
poisdeux: Nice catch. Fixed.

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
poisdeux: fixed

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
poisdeux: I don't agree. If you change the language, either by mistake or just to try it out, it can be really hard to change it back if you don't understand the language at all.

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
poisdeux: fixed

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
poisdeux: fixed

app/src/main/java/org/xbmc/kore/utils/Utils.java
poisdeux: Currently we do not have translations requiring variant or extensions and we only have language codes and some supplemented with country codes. As we still support API 15 I think we should stick to this and prevent as much of the API if-statements as possible.
When we finally bump the minimum supported SDK to 21 or higher we can always replace this (and test it then).

app/src/main/java/org/xbmc/kore/utils/Utils.java
poisdeux: It's `_`. see `BuildConfig.SUPPORTED_LOCALES`

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
poisdeux: See my comment below.

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
SyncedSynapse: Ok

app/src/main/java/org/xbmc/kore/ui/sections/settings/SettingsFragment.java
SyncedSynapse: Ok

app/src/main/java/org/xbmc/kore/utils/Utils.java
SyncedSynapse: We won't remember to do this, when/if we bump to SDK 21. It's not much trouble doing an if here and using the new versions.

app/src/main/java/org/xbmc/kore/utils/Utils.java
SyncedSynapse: The `_` is something that you defined, it's not standard. The standard is `-`, so that's what should be used. `BuildConfig.SUPPORTED_LOCALES` should be changed also.

poisdeux: Dug into it a little deeper and you're right, IETF seems to be the standard way to do it and uses the hyphen as separator.

PR #681, opened 2019-10-26
Stream local files with NanoHTTPD server

Upabjojr:
This PR add a local file browser. Files can be streamed to Kodi from the Android device.

- Adds NanoHTTPD as a Gradle dependency.
- _Local Files_ section is added just below _Files_.
- The tabs of _Local Files_ are the external storage, the DCIM folder and the detected SD cards.
- The class HttpApp creates a local HTTP server returning the content of a single local file.
- The selected file is linked to the HttpApp and sent to Kodi with the `Player.Open` JSON-RPC call.
- I have created a local thumbnail loader for JPEG pictures (sorry, I just don't trust Picasso web-app with my local files. By the way, I was unable to make Picasso work with local files).  Asynchronous loading of thumbnails through AsyncTask, the code is quite simple.

Current issues:
- A lot of the code in this PR was just copied from `org.xbmc.kore.ui.sections.file`, maybe it can be refactored in order to use class inheritance to keep identical code shared.
- Queuing does not work. This probably requires extending HttpApp to allow multiple files to be downloaded, and creating a way to keep the queue in HttpApp synchronized with Kodi's queue.
- Thumbnails are generated for JPEG files only.
- No media information beyond file name (this can be added in the future).
- No security for the HTTP server. Maybe support for tokens should be added using cryptographically-secure random number generators. SSL would also nice to have, but a bit fussy due to the requirement of key files.


app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
```diff
@@ -0,0 +1,112 @@
...
+        Map<String, List<String>> parms = session.getParameters();
+        if (filePath == null) {
+            return newFixedLengthResponse(Response.Status.FORBIDDEN, "", "");
+        }
+
+        FileInputStream fis = null;
+        try {
+            fis = new FileInputStream(filePath);
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
```
Upabjojr: Forgot to return HTTP 403 FORBIDDEN here.

SyncedSynapse:
Thanks for this. I'm sorry but i still haven't been able to check this out. I expect to do it in the next couple of days, and i'll report back.

SyncedSynapse:
Ok, this is a big PR so there's some stuff to review. But first of all, thanks again. In my (somewhat limited) tests, this worked fine, and just needs some fine tuning.

Starting from the issues you point out, my opinion is:
1. I have no problem with copying the file section. @poisdeux might not like it as much, but i think it's fine by now.
2. If queuing isn't working the options "Play from here" and "Queue" should be removed from the the overflow menu in the files list. Or just remove it completely, as the only possible action is to play an item, and that is done by clicking on it.
3. Regarding picture display, i'm not sure i understood your comment regarding Picasso. This is the [Picasso library](https://square.github.io/picasso/), nothing to do with Picass**a** web app. And it should work fine with local files (try a Google search with "picasso library local file" and you probably find the issue). I think that using Picasso should be preferred, because it's highly optimized and tested against leaks for instance.
4. Ok
5. No problem by me. But if there's a way to shut down the http server once the user navigates away from the activity, it should be done.

Now, some comments i have
1. The Singleton `HttpApp` could be better implemented. I'll point out some issues on the code, and you can take a look at the `HostManager` class for an example;
2. There's an obvious code error in `LocalPictureListFragment` `getFileMimeType` (Java uses equals() on strings, not ==). I suspect that you're not using Android Studio as it pointed that out immediately - use whatever you want obviously, but Android Studio is actually quite nice.
3. The push to refresh on the file list isn't working. I was going to point this out as an issue, but just noticed that the same happens with the default Files section. I'll check it out later, but do you could also check if it's possible to simply implement the `onRefresh` with some code that's already on the `onCreateView`.
4. Related to the previous one. The code doesn't handle the request permissions result `onRequestPermissionsResult`, which is odd in either case. At least, if the user gave it permission, it should reload.
5. Adding more tabs with music and movies seems very straightforward and easy. We might have to tweak the item layout, as it doesn't make sense to display a picture.
6. You probably missed this: in Kore's settings there's the option "Side Menu Shortcuts" that control which elements appear on the navigation drawer. Local files isn't appearing because it's not on the array `entries_nav_drawer_items` in `arrays.xml`. Needs to be added, in the same order as `entry_values_nav_drawer_items`, so that it works.

This is it for now. @poisdeux it would be useful if you could take a look at this too.

app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.Map;
+
+import fi.iki.elonen.NanoHTTPD;
+
+import static android.content.Context.WIFI_SERVICE;
+
+
+public class HttpApp extends NanoHTTPD {
+
+    public HttpApp(Activity activity, int port) throws IOException {
```
SyncedSynapse: `protected` or `private`

app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
```diff
@@ -0,0 +1,113 @@
...
+        super(port);
+        this.port = port;
+        this.activity = activity;
+        start(NanoHTTPD.SOCKET_READ_TIMEOUT, false);
+    }
+
+    static int port;
```
SyncedSynapse: Neither of this should be static. Both should me regular members. 
Do we need `port` here? Isn't it accessible via the superclass?
It probably should be an `Activity` but a `Context`.

app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
```diff
@@ -0,0 +1,113 @@
...
+    static Activity activity;
```
SyncedSynapse: Neither of this should be static. Both should me regular members.
Do we need port here? Isn't it accessible via the superclass?
It probably should be an Activity but a Context.

app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
```diff
@@ -0,0 +1,113 @@
...
+
+    String fileName = null;
+    String filePath = null;
+    String mimeType = null;
```
SyncedSynapse: `private`

app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
```diff
@@ -0,0 +1,113 @@
...
+        Map<String, List<String>> parms = session.getParameters();
+        if (filePath == null) {
+            return newFixedLengthResponse(Response.Status.FORBIDDEN, "", "");
+        }
+
+        FileInputStream fis = null;
+        try {
+            fis = new FileInputStream(filePath);
+        } catch (FileNotFoundException e) {
+            e.printStackTrace();
```
SyncedSynapse: Please use `LogUtils.LOGW`

app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
```diff
@@ -0,0 +1,113 @@
...
+        try {
+            ip = getIpAddress();
+        } catch (UnknownHostException uhe) {
+            return null;
+        }
+        try {
+            if (!isAlive())
+                start();
+        } catch (IOException ioe) {
+            Log.e(ioe.getClass().toString(), ioe.getMessage());
```
SyncedSynapse: `LogUtils.LOGE`

app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
```diff
@@ -0,0 +1,113 @@
...
+        return "http://" + ip + ":" + String.valueOf(port) + "/" + fileName;
+    }
+
+    private static HttpApp http_app = null;
+
+    public static HttpApp getInstance() throws IOException {
+        if (http_app == null) {
+            http_app = new HttpApp(activity, port);
+        }
+        return http_app;
```
SyncedSynapse: `getInstance(Context context, int port)`

I don't expect to have issues with threading here, but using a `synchronized` can't hurt. If you'd like take a look at `HostManager`.

app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    public static void setActivity(Activity activity) {
+        HttpApp.activity = activity;
+    }
+
+    public static void setPort(int port) {
+        HttpApp.port = port;
+    }
```
SyncedSynapse: Not needed

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalFileActivity.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2015 DanhDroid. All rights reserved.
```
SyncedSynapse: Your name or XBMC Foundation

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalFileListFragment.java
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright 2015 DanhDroid. All rights reserved.
```
SyncedSynapse: Your name or XBMC Foundation

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalFileListFragment.java
```diff
@@ -0,0 +1,98 @@
...
+        Bundle dcimFileListArgs = new Bundle();
+        String dcim = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DCIM).getAbsolutePath();
+        dcimFileListArgs.putString(LocalPictureListFragment.ROOT_PATH_LOCATION, dcim);
+        dcimFileListArgs.putParcelable(LocalPictureListFragment.SORT_METHOD, sortMethod);
+
+        Bundle externalStorageFileListArgs = new Bundle();
+        String externalStorage = Environment.getExternalStorageDirectory().getAbsolutePath();
+        externalStorageFileListArgs.putString(LocalPictureListFragment.ROOT_PATH_LOCATION, externalStorage);
+        externalStorageFileListArgs.putParcelable(LocalPictureListFragment.SORT_METHOD, sortMethod);
+
```
SyncedSynapse: If it's just a matter or adding `Environment.DIRECTORY_MUSIC` and `Environment.DIRECTORY_MOVIES` here, i think it's worth it.

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalPictureListFragment.java
```diff
@@ -0,0 +1,766 @@
...
+        Bundle args = getArguments();
+        FileLocation rootPath = null;
+
+
+        if (ContextCompat.checkSelfPermission(this.getActivity(), Manifest.permission.WRITE_EXTERNAL_STORAGE)
+                != PackageManager.PERMISSION_GRANTED) {
+            // Permission to read files was not granted,
+            // prompt the user for storage access permissions:
+            ActivityCompat.requestPermissions(getActivity(),
+                    new String[]{Manifest.permission.WRITE_EXTERNAL_STORAGE},
```
SyncedSynapse: `WRITE_EXTERNAL_STORAGE` ? Don't you mean `READ_EXTERNAL_STORAGE`. That's what's on the manifest.

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalPictureListFragment.java
```diff
@@ -0,0 +1,766 @@
...
+                    42);
```
SyncedSynapse: Yeah, it's the answer for everything, but in this case it should be `Utils.PERMISSION_REQUEST_READ_EXTERNAL_STORAGE` (not created).

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalPictureListFragment.java
```diff
@@ -0,0 +1,766 @@
...
+        }
+
+        try {
+            if (http_app == null) {
+                HttpApp.setActivity(getActivity());
+                HttpApp.setPort(8080);
+                http_app = HttpApp.getInstance();
+            } else {
+                http_app.start();
```
SyncedSynapse: If there's a way to start it, should there be a way to stop it (and avoid consuming resources)?
And i don't get this, in the previous branch shouldn't it also be started?

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalPictureListFragment.java
```diff
@@ -0,0 +1,766 @@
...
+
+    private String getFileMimeType(final String filename) {
+        String mimeType = null;
+        if (mediaType == "video") {
+            mimeType = "video/mp4";
+        } else if (mediaType == "image") {
+            mimeType = "image/jpeg";
+        } else if (mediaType == "audio") {
+            mimeType = "audio/mp3";
+        }
```
SyncedSynapse: `mediaType.equals()` instead of `==`

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalPictureListFragment.java
```diff
@@ -0,0 +1,766 @@
...
+            }
+        }, callbackHandler);
+    }
+
+    /**
+     * Queues the given media file on the active playlist, and starts it if nothing is playing
+     * @param filename File to queue
+     */
+    private void queueMediaFile(final String filename) {
+        Toast.makeText(getContext(), "Media not really implemented for local device files.", Toast.LENGTH_LONG);
```
SyncedSynapse: This needs to be a string resource

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalPictureListFragment.java
```diff
@@ -0,0 +1,766 @@
...
+        File directory = null;
+        if (rootFileLocation != null) {
+            directory = new File(rootPathLocation);
+        } else {
+            return;
+        }
+        File[] files = directory.listFiles();
+
+        if (files == null) {
+            Toast.makeText(getActivity(), "Did you give Kore permission to access file storage?", Toast.LENGTH_LONG);
```
SyncedSynapse: This needs to be a string resource

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalFileActivity.java
Upabjojr: I think I need to keep the original author as well, as the file has been copied.

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalPictureListFragment.java
Upabjojr: `PERMISSION_REQUEST_READ_EXTERNAL_STORAGE` appears to be non-existent. Shall I add it?

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalPictureListFragment.java
Upabjojr: Stopping is a bit more complicated, because Kodi may still be reading from the HTTP server. For example, if you are playing a local video or MP3.

The HTTP server should only be destroyed when you are sure Kodi isn't reading anything anymore.

SyncedSynapse: Got it, hadn't thought of that.

poisdeux:
@SyncedSynapse I'll try to make some time but can't promise anything.

Upabjojr:
I have so far:

1. Added the support for queuing local files.
2. Switched to Picasso for thumbnails. On my phone only the first JPEG appears to have its thumbnail loaded, the following ones fail.
3. `HttpApp` now has a `LinkedList` of `LocalFilePosition` objects. In this way it is possible to create playlists with more than one local file. The HTTP GET request now requires a `number` parameter to retrieve the correct file from the ones linked.
4. An abstract class superseding both the local files package and the remote file package has been created. Ideally the code in common should be uplifted on this new abstract class, but I'm a bit afraid of breaking working code.
5. `onRefresh( )` is still broken. I'm not going to fix it in this PR.
6. `onRequestPermissionsResult` is still not handled.
7. Rules for shutting down the HTTP server are a bit complicated. You may wish to keep it running even after leaving Kore. The class `HttpApp` should be able to communicate with Kodi and stop if Kodi's playlists are empty and Kodi isn't playing anything. I would prefer not to implement this logic in this PR.

So the remaining issues to address in this PR are Picasso failing and `onRequestPermissionsResult`.

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalFileListFragment.java
```diff
@@ -0,0 +1,110 @@
...
+            throw new ClassCastException(activity.toString() + " unable to register BackPressedListener");
+        }
+    }
+
+    @Override
+    public boolean onBackPressed() {
+        // Tell current fragment to move up one directory, if possible
+        LocalPictureListFragment curPage = (LocalPictureListFragment)((TabsAdapter)getViewPager().getAdapter())
+                .getStoredFragment(getViewPager().getCurrentItem());
+        if ((curPage != null) && !curPage.atRootDirectory()) {
```
poisdeux: It seems the `atRootDirectory` check is faulty. I get into a loop when pressing back on the music tab (currently only one I'm trying to test).

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalPictureListFragment.java
```diff
@@ -0,0 +1,745 @@
...
+                parentDirectory = rootPath;
+                dir.setRootDir(true);
+            } else if (dir.file != null) {
+                parentDirectory = getParentDirectory(dir.file);
+            }
+        }
+
+        // TODO: maybe async task here?
+        File[] files = (new File(dir.file)).listFiles();
+        Arrays.sort(files);
```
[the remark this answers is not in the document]
Upabjojr: I've added a check for null pointers.

app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public void addLocalFilePath(String filename, String filePath, String mimeType) {
+        this.fileName = filename;
+        this.filePath = filePath;
+        this.mimeType = mimeType;
+    }
+
+    private String getIpAddress() throws UnknownHostException {
+        WifiManager wm = (WifiManager) applicationContext.getSystemService(WIFI_SERVICE);
```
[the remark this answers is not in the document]
Upabjojr: Ideally the device should check if Kodi is able to connect to the device. Is there any way to do this with JSON-RPC?

Upabjojr:
> I tried testing the feature using Android API 15 and 29 with Kodi 15.4. Unfortunately, I couldn't get a local file to run on Kodi.

I only have the latest version of Kodi and it works for me (both pictures and audio and video files).

Upabjojr:
So, I understand that this is a huge PR and the new feature it introduces is not-so-well tested.

This is my first time ever with Android app development, and I'm not even used to Java, so I can only give you some limited contributions to the code (and I also need a lot of googling). I just wanted to stream local files to my Kodi and thought it would be a nice idea to share my edits back to the community.

I would suggest at this point to add a warning about the `Local Files` tab being experimental, and warn the users to report any possible bugs or issues here. Maybe in the future someone else will fix the issues with this PR.

I'm sorry about older Kodi versions apparently not working, but I don't have time right now to do some thorough testing.

Would you consider this PR to be ready for merging? Shall I add a warning about this addition being experimental?

app/src/main/java/org/xbmc/kore/ui/sections/localfile/LocalFileListFragment.java
SyncedSynapse: This particular code is equal to the existing code in the `Files` section, so maybe we have a bug there that needs solving.

app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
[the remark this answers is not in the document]
SyncedSynapse: There is, but this is the same issue that's all over the place. In the Movies/TV Shows/Music section it also doesn't check if it's connected to Kodi, and if the user tries to open one of those items without a connection an error is shown, so there's not much of a difference to what's happening here.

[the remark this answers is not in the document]
SyncedSynapse: Marking this as resolved. @poisdeux  if you don't agree feel free to reopen it.

Upabjojr:
OK, fixed the last (easy) issues.

poisdeux:
> > I tried testing the feature using Android API 15 and 29 with Kodi 15.4. Unfortunately, I couldn't get a local file to run on Kodi.
> 
> I only have the latest version of Kodi and it works for me (both pictures and audio and video files).

Sorry, I tested on 18.4 not 15.4. I tested with an emulator and the problem is the emulator. The request is sent to Kodi but Kodi gets a connection timeout while trying to connect to the emulator. Currently my HTPC is broken so I can't do a proper test. As you both seem to be having no issues I guess it's all fine.

PR #693, opened 2019-12-07
Share local files - part 2

Upabjojr:
This is the second part to https://github.com/xbmc/Kore/pull/681

Now files can be shared to Kodi through Kore from other apps. This PR extends the _"Play on Kodi"_ intent to use `HttpApp` to add shared files.

List of edits:

1. Media files can now be shared with Kodi (see edits to manifest file).
2. `HttpApp` now generates a _token_ for security reasons. Without _token_ you won't be able to download any file.
3. Added `finish()` in order not to open Kore's screen after a file has been shared (do you really care about seeing Kore? I suppose you are looking at Kodi).

Some remaining doubts:

4. ~~Media files share opens the files in Kodi. The code for YouTube videos is queuing the video instead.  What should the correct behaviour be? I've noticed that Yatse has two ways of sharing: play and queue, maybe Kore should add the same?~~ ___Update:___ this PR now adds a second share option `Queue on Kodi`. Now `Play on Kodi` plays the shared media file directly, disregarding the playlist, while `Queue on Kodi` adds the shared media file to the end of the playlist.
5. Unrelated: the current sharing options always assume that the shared link is a video, and it's queued to the video playlist (even if it's an image or audio file), and Kodi is unable to play it.
6. Local files are shared as `content:// ... ` paths, not as file paths. Apparently Android's API allows to open an input file stream from `content://` paths, but it's a bit hard to locate the file path. I've added a second sharing link to `HttpApp` in order to handle generic `Uri` objects beside file paths.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/OpenSharedUrl.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/OpenSharedUrl.java
@@ -39,22 +39,26 @@ public class OpenSharedUrl implements Callable<Boolean> {
 
     private static final String TAG = LogUtils.makeLogTag(OpenSharedUrl.class);
     private final HostConnection hostConnection;
-    private final String pluginUrl;
+    private final String url;
     private final String notificationTitle;
     private final String notificationText;
+    private boolean queue;
+    private int playlistType;
 
     /**
-     * @param pluginUrl The url to play
+     * @param url The url to play
      * @param notificationTitle The title of the notification to be shown when the
      *                          host is currently playing a video
      * @param notificationText The notification to be shown when the host is currently
      *                         playing a video
      */
-    public OpenSharedUrl(HostConnection hostConnection, String pluginUrl, String notificationTitle, String notificationText) {
+    public OpenSharedUrl(HostConnection hostConnection, String url, String notificationTitle, String notificationText, boolean queue, int playlistType) {
         this.hostConnection = hostConnection;
-        this.pluginUrl = pluginUrl;
+        this.url = url;
         this.notificationTitle = notificationTitle;
         this.notificationText = notificationText;
+        this.queue = queue;
+        this.playlistType = playlistType;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
```diff
@@ -362,9 +365,13 @@ private static ExecutorService getShareExecutor() {
      * Handles the intent that started this activity, namely to start playing something on Kodi
      * @param intent Start intent for the activity
      */
-    private void handleStartIntent(Intent intent) {
+    protected void handleStartIntent(Intent intent) {
+        handleStartIntent(intent, false);
```
SyncedSynapse: Shouldn't `queue` be set to `true` by default? The previous behaviour queued the files. I'll make a brief comment on the code pointing this out.

app/src/main/java/org/xbmc/kore/ui/sections/remote/OpenSharedUrl.java
```diff
@@ -71,37 +75,46 @@ public Boolean call() throws Error, InterruptedException {
...
+                hostConnection.execute(new Playlist.Clear(playlistType))
                     .get();
             }
 
-            LogUtils.LOGD(TAG, "Queueing file");
-            PlaylistType.Item item = new PlaylistType.Item();
-            item.file = pluginUrl;
-            hostConnection.execute(new Playlist.Add(PlaylistType.VIDEO_PLAYLISTID, item))
-                .get();
+            if (queue) {
```
SyncedSynapse: [related to the other comment]

AFAICT the previous behaviour is similar to having `queue` set to `true` so that should be the default in `handleStartIntent`

app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
SyncedSynapse: Ignore this and the other comment. Noticed now that it has nothing to do with what i wrote, and is only used to display a message.

SyncedSynapse:
I've taken a quick look at it, and it looks very good.

Unfortunately i can't test and finish reviewing it right now, so i ask you to give me a couple more days to check this. 

Thanks.

app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
Upabjojr: > is only used to display a message.

Well... it calls `Player.Open` instead of `Playlist.Add`.

I have changed the behaviour of `Play with Kodi` because the word _play_ sounds like playing immediately, and queuing is a bit weird. When I was using Kore for the first time I thought it wasn't working, not realizing it was queuing the medias.

SyncedSynapse:
This does indeed look good. Can't say that i tested all possible scenarios, but it worked flawlessly in what i tested, so in it goes.

Thanks.

Issue #695, opened 2019-12-13
Search field for PVR TV/Radio channels and Recordings.


Selaron:
In the views Music, TV Shows and Movies there is a magnifier icon after tipping on which you can input a search term filtering the lists.

I'd like to implement a search field like this for PVR secions:
TV Channels
Radio Channels
Recordings.

As Kodi JSON API does not seeme to support filtering parameters for PVR data, the filtering would need to be implemented within the app after the full list of items has been received from the server.

Some PVRs have a large number of Channels (especially when used with SAT) and recordings. Even though recordings can now be sorted (#385) a search field would help to find desired recordings even faster.

I'm ready to prepare a PullRequest.

Selaron:
The search input functionality is implemented in AbstractCursorListFragment but it would not be viable to extend from AbstractCursorListFragment with PVRRecordingsListFragment as no database is queried by PVRRecordingsListFragment.

Should search functionality (menu item + SearchView) instead be pulled up into AbstractListFragment so that PVRRecordingsListFragment can extend AbstractListFragment ?

PR #697, opened 2019-12-17
Implemented search field for PVR views. (#695)

Selaron:
This implements a multi-word search for PVR Recordings, PVR Channels and PVR Radiochannels.
On channel views, when typing multiple words, a channel is displayed if all words each match either the channel name or the title of the current broadcast (match mode: case insensitive contains).
On recordings view, a recording is displayed if all words each match either the recording title or the channel name it has been recorded on (match mode: case insensitive contains)

For `AbstractSearchableFragment` I mostly copied code from `AbstractCursorListFragment` and I'm not sure if all of this is at all reqiured. I'm not sure what savedSearchFilter is used for, what `onPause()` or `onResume()` does and if `AbstractSearchableFragment#onCreateView` is required at all.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/AbstractSearchableFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AbstractSearchableFragment.java
@@ -140,11 +140,7 @@ public abstract class AbstractSearchableFragment extends Fragment implements Sea
 
         /**
          * When this fragment is paused, onQueryTextChange is called with an empty string.
-         * This causes problems restoring the list fragment when returning. On return the fragment
-         * is recreated, which will cause the cursor adapter to be recreated. Although
-         * the search view will have the query restored to its saved state, the CursorAdapter
-         * will use the empty search filter. This is due to the fact that we don't restart the
-         * loader when it is still loading after its been created.
+         * This causes problems restoring the list fragment when returning.
          */
         if (isPaused)
             return true;
--- a/app/src/main/java/org/xbmc/kore/ui/AbstractSearchableFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AbstractSearchableFragment.java
@@ -183,6 +183,15 @@ public abstract class AbstractSearchableFragment extends Fragment implements Sea
         isPaused = true;
     }
 
+    @Override
+    public void onSaveInstanceState(Bundle outState) {
+        if (!TextUtils.isEmpty(searchFilter)) {
+            savedSearchFilter = searchFilter;
+        }
+        outState.putString(BUNDLE_KEY_SEARCH_QUERY, savedSearchFilter);
+        super.onSaveInstanceState(outState);
+    }
+
     /**
      * Event bus post. Called when the syncing process ended
      *
--- a/app/src/main/java/org/xbmc/kore/ui/AbstractSearchableFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/AbstractSearchableFragment.java
@@ -19,8 +19,6 @@ import android.widget.EditText;
 import org.xbmc.kore.R;
 import org.xbmc.kore.jsonrpc.event.MediaSyncEvent;
 
-import de.greenrobot.event.EventBus;
-
 public abstract class AbstractSearchableFragment extends Fragment implements SearchView.OnQueryTextListener {
 
     private String searchFilter = null;
@@ -28,7 +26,6 @@ public abstract class AbstractSearchableFragment extends Fragment implements Sea
     private boolean supportsSearch;
     private boolean isPaused;
 
-    private EventBus bus;
     private SearchView searchView;
 
     private final String BUNDLE_KEY_SEARCH_QUERY = "search_query";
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/xbmc/kore/ui/AbstractSearchableFragment.java
```diff
@@ -0,0 +1,211 @@
...
+import de.greenrobot.event.EventBus;
+
+public abstract class AbstractSearchableFragment extends Fragment implements SearchView.OnQueryTextListener {
+
+    private String searchFilter = null;
+    private String savedSearchFilter;
+    private boolean supportsSearch;
+    private boolean isPaused;
+
+    private EventBus bus;
```
SyncedSynapse: I don't think this is needed.

app/src/main/java/org/xbmc/kore/ui/AbstractSearchableFragment.java
```diff
@@ -0,0 +1,211 @@
...
+    private SearchView searchView;
+
+    private final String BUNDLE_KEY_SEARCH_QUERY = "search_query";
+
+    @Nullable
+    @Override
+    public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {
+        View root = super.onCreateView(inflater, container, savedInstanceState);
+
+        bus = EventBus.getDefault();
```
SyncedSynapse: I don't think this is needed.

app/src/main/java/org/xbmc/kore/ui/AbstractSearchableFragment.java
```diff
@@ -0,0 +1,211 @@
...
+        View view = searchView.findViewById(R.id.search_close_btn);
+        if (view != null) {
+            view.setOnClickListener(new View.OnClickListener() {
+                @Override
+                public void onClick(View v) {
+                    EditText editText = (EditText) searchView.findViewById(R.id.search_src_text);
+                    editText.setText("");
+                    searchView.setQuery("", false);
+                    searchFilter = savedSearchFilter = "";
+                    restartLoader();
```
SyncedSynapse: Call something else to this function, as it has nothing to do with Android's Loader - refreshList() for instance.

app/src/main/java/org/xbmc/kore/ui/AbstractSearchableFragment.java
```diff
@@ -0,0 +1,211 @@
...
+                break;
+        }
+        return super.onOptionsItemSelected(item);
+    }
+
+
+
+    @Override
+    public void onResume() {
+ //       bus.register(this);
```
SyncedSynapse: You've already commented it, remove it completly. Same in `onPause`

SyncedSynapse:
I made some comments that point out things that i think are dispensable. Can you please check them, remove them and see if everything works or if i missed something.

Regarding your comments, `onPause`, `onResume`, `onCreateView` are part Android's view lifecycle and are required. The stuff related to the `EventBus` is unnecessary, i believe.

The `savedSearch` can be tested like this: 
1) type something in the search filter, and check that it filters
2) exit kore (with the home button)
3) get back to kore and check if the filter is still on, and coherent with what's displayed on the search box.

Selaron:
Thanks for your quick feedback. I applied your suggestions and it works.

Trying to test `savedSearchFilter` I set a breakpoint where the field is written but this never gets hit.
My emulator is a Pixel3 API 28.
Never the less pressing home and then resuming the app also keeps the search field input and item list state as is. Maybe recent android versions simply preserve the full app state?

Selaron:
Need to fix one issue: on channel programme guide, the search has no effect. When inputting sth. and going back to channel list, Kore crashes. I'll notify when done.

poisdeux:
> Thanks for your quick feedback. I applied your suggestions and it works.
> 
> Trying to test `savedSearchFilter` I set a breakpoint where the field is written but this never gets hit.
> My emulator is a Pixel3 API 28.
> Never the less pressing home and then resuming the app also keeps the search field input and item list state as is. Maybe recent android versions simply preserve the full app state?

Yes it does. But not when you change the device orientation from portrait to landscape or vice versa. In that case it has to recreate the whole view and we need to restore the state by saving the search filter first and get it back from the `bundle` in `onCreateView`.

app/src/main/java/org/xbmc/kore/ui/AbstractSearchableFragment.java
```diff
@@ -0,0 +1,207 @@
...
+    /** {@inheritDoc} */
+    @Override
+    public boolean onQueryTextChange(String newText) {
+        if ((!searchView.hasFocus()) && TextUtils.isEmpty(newText)) {
+            //onQueryTextChange called as a result of manually expanding the searchView in setupSearchMenuItem(...)
+            return true;
+        }
+
+        /**
+         * When this fragment is paused, onQueryTextChange is called with an empty string.
```
poisdeux: Please update this comment as it is referring to a `cursoradapter` that is not used here.
If I understand correctly I think the first two sentences will do just fine.

Selaron:
I fixed the crash und made the search field work on CannelEPGlist either.

I also confirmed that `savedSearchFilter` is functional and actually used when device orientation is changed thanks to @poisdeux 's hint.

Issue #125, opened 2015-10-02
[Feature Request] local playback, streaming media to android device


115ek:
Are there any plans concerning that? I only found a download button in the current release.
Actually it should be possible to play (at least some music...) media on your android device without doing it in Kodi at the same time.
(I think yatse also has got such a feature.)
Please enlight me if I missed something in the settings of Kore. Maybe I'm only blind.
Best regards :)

cryptobanana:
I'm working on this feature, as well as video. I have a partially working prototype for movies implemented in my own development branch: https://github.com/cryptobanana/Kore/tree/stream_video

I plan to finish this functionality for videos and audio in my free time over the holidays and submit a pull request soon.

timofonic:
@cryptobanana 

Great!

I'm more interested in the audio-only option, to use headphones. Bluetooth is shitty!

cryptobanana:
Hi, I have a proof-of-concept implementation (in progress) in my forked repository that allows a Kore remote to play TV and Videos locally on the remote device (as opposed to the kodi media center). The user interface is implemented by repurposing the "download" button in the TvShowEpisodeDetails and MovieDetails views. Audio would be a next step. I'm not yet ready to issue a pull request because it requires more iteration + cleanup, but feel free to try it yourself:

https://github.com/cryptobanana/Kore/tree/stream_video

cryptobanana:
@SyncedSynapse: I'd like to kindly request some feedback on the video + audio streaming functionality I'm working on, as I'd like to ensure that it can be merged upstream: 

1) Choose a user interface for playing media (stream to device): Currently in my branch, I am re-purposing the "download" button in the TVShowEpisodeDetailsFragment and MovieDetailsFragment.java to stream a video and play it. Consequently download functionality no longer works. I would propose to address this by either: a) adding an additional button next to the download button specifically to initiate streaming local playback or b) modify the behavior of the floating action button so that upon pressing it, the user has the option to play on the kodi media center or to stream for local playback on the kore remote. I personally tend to prefer (b).

2) Reduce Code duplication: MovieDetailsFragment.java and TVShowEpisodeDetailsFragment.java duplicate a significant amount of code. I would propose to address this by moving as much of the duplicated code into higher classes as possible. Thoughts?

SyncedSynapse:
@cryptobanana I won't have much time this week to review your branch, but i'll try to look at it at the weekend/next week.

Regarding your questions:
[...]

PR #703, opened 2020-01-10
Added ability to play media files locally (#125)

rabits:
Just the same way as played from Movies section, but from Files browser.


poisdeux:
Looks and works good! Thx!
Nice, that it also works for audio files.

poisdeux:
Sorry it took so long. Kinda forgot about this.

rabits:
No worries - not urgent)

PR #710, opened 2020-02-06
Uri encoding for local filenames.

mbien:
Kodi couldn't play http-served files with spaces or other special characters in their names. It turned out that it was easy to fix by simply URI encoding the filename so that a valid link is created.

The else branch seems to be using URIs already, so i didn't touch it, I assume those are already properly encoded.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/localfile/HttpApp.java
@@ -163,9 +163,8 @@ public class HttpApp extends NanoHTTPD {
             String filename = localFileLocationList.get(currentIndex).fileName;
             path = Uri.encode(filename) + "?number=" + currentIndex;
         } else {
-            Uri uri = localUriList.get(currentIndex);
-            String filename = getFileNameFromUri(uri);
-            path = filename + "?uri=" + currentIndex;
+            String filename = getFileNameFromUri(localUriList.get(currentIndex));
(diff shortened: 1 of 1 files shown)


SyncedSynapse:
This looks good to me, but i would like to know if @Upabjojr  has something to add, given that this was his code originally.

Upabjojr:
Only `?number={...}` really matters on the file selection. Filename may be anything, it's just what you want Kodi to show.

Upabjojr:
Do you think this problem can occur on https://github.com/xbmc/Kore/pull/710/files#diff-d809a4ce9bf44d49ddbe142ebff36c8fR168 as well?

mbien:
> 
> 
> Do you think this problem can occur on https://github.com/xbmc/Kore/pull/710/files#diff-d809a4ce9bf44d49ddbe142ebff36c8fR168 as well?

i don't think it can, because it already is a URI so it *should* be properly encoded already. 

I didn't really know how to debug it since I didn't know what to do in Kore to get to the else branch. So I didn't touch it like i mentioned in the pull request msg (encoding a URI twice would break things most likely).

Upabjojr:
You should be able to trigger the else branch if you share a file from an external application to Kore.

mbien:
> You should be able to trigger the else branch if you share a file from an external application to Kore.

ah I see. It turned out that the else branch needed the same treatment despite the fact that it was already dealing with URIs. Added commit 0dc1ae6

PR #719, opened 2020-03-22
Added support for Prime Videos shared from the official Amazon app

Upabjojr:
Prime Video app supports sharing the links to videos. They can be shared through Kore.

Kore is currently unable to recognize the link, because the Prime Video app shares the link in a strange way (I cast the intent to a string, then match the link with regular expressions).

As a last step, I added the conversion logic to `plugin://plugin.video.amazon-test/...`, so that the movie can be watched on Kodi with the Amazon VOD addon.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
@@ -389,6 +389,21 @@ public class RemoteActivity extends BaseActivity
             videoUri = intent.getData();
         }
 
+        if (videoUri == null) {
+            // Some apps hide the link in the clip, try to detect any link by casting the intent
+            // to string a looking with a regular expression:
+
+            Matcher matcher = Pattern.compile("https?://[^\\s]+").matcher(intent.toString());
+            String matchedString = null;
+            if (matcher.find()) {
+                matchedString = matcher.group(0);
+                if (matchedString.endsWith("}")) {
+                    matchedString = matchedString.substring(0, matchedString.length() - 1);
+                }
+                videoUri = Uri.parse(matchedString);
+            }
+        }
+
         String url;
         if (videoUri == null) {
             url = getShareLocalUri(intent);
@@ -597,6 +612,12 @@ public class RemoteActivity extends BaseActivity
         } else if (host.endsWith("soundcloud.com")) {
             return "plugin://plugin.audio.soundcloud/play/?url="
                     + URLEncoder.encode(playuri.toString());
+        } else if (host.startsWith("app.primevideo.com")) {
+            Matcher amazonMatcher = Pattern.compile("gti=([^&]+)").matcher(playuri.toString());
+            if (amazonMatcher.find()) {
+                String gti = amazonMatcher.group(1);
+                return "plugin://plugin.video.amazon-test/?asin=" + gti + "&mode=PlayVideo&adult=0&name=&trailer=0&selbitrate=0";
+            }
         }
         return null;
     }
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
@@ -390,26 +390,11 @@ public class RemoteActivity extends BaseActivity
         }
 
         if (videoUri == null) {
-            // Some apps hide the link in the clip, try to detect any link by casting the intent
-            // to string a looking with a regular expression:
-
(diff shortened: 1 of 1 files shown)


poisdeux:
Can't really test this but the code looks fine.
@SyncedSynapse do you have any comments?

SyncedSynapse:
I do have some comments - i pointed them out in the code, but yesterday somehow i managed to delete them, so i'll write them again.

Can't test this either, but i noticed that there's the potential for this PR to interfere with the local sharing of videos, so i would feel more comfortable if the code took that explicitly under consideration, as i'll point out in a separate comment.

app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
```diff
@@ -389,6 +389,21 @@ protected void handleStartIntent(Intent intent, boolean queue) {
...
+            Matcher matcher = Pattern.compile("https?://[^\\s]+").matcher(intent.toString());
+            String matchedString = null;
+            if (matcher.find()) {
+                matchedString = matcher.group(0);
+                if (matchedString.endsWith("}")) {
+                    matchedString = matchedString.substring(0, matchedString.length() - 1);
+                }
+                videoUri = Uri.parse(matchedString);
+            }
+        }
```
SyncedSynapse, line 405: This whole `if (videoUri == null)` precedes the next one, which is similar, and deals with local sharing, so it can "capture" requests that should be done by it. I think that the logic should be unified, and that any division between the 2 logics should be explicitly acknowledge and solved.

Upabjojr: I've changed it to try to detect `videoUri` if `null`, then `toPluginUrl` will always be called. I have made some tests and it keeps working. Hope this makes sense.

Upabjojr:
> Can't test this either, but i noticed that there's the potential for this PR to interfere with the local sharing of videos, so i would feel more comfortable if the code took that explicitly under consideration, as i'll point out in a separate comment.

Is there a way to add unit testing with a simulated Kore-Kodi environment on Travis?

poisdeux:
> > Can't test this either, but i noticed that there's the potential for this PR to interfere with the local sharing of videos, so i would feel more comfortable if the code took that explicitly under consideration, as i'll point out in a separate comment.
> 
> Is there a way to add unit testing with a simulated Kore-Kodi environment on Travis?

Yes there is. Have a look at the tests in https://github.com/xbmc/Kore/blob/master/app/src/test/java/org/xbmc/kore/tests/jsonrpc/. They use a TCP server that mocks some of the JSON RPC calls from Kodi. You can do something similar. Although I think in your case you do not need it and just need to mock the intent that is passed to `handleStartIntent` and for that I would use Mockito.
If you need help setting this up I'm glad to help.

SyncedSynapse:
This looks good to me.

@poisdeux not sure if you want to check anything else, but for me this is good to me merged. If you're confortable with it, please merge it.

One other thing: i've tried to play a video locally (to VLC), but i'm getting an error. Not sure if it has something to do with my network configuration, so can you please check whether the play locally feature still works for you, or have we bungled it somewhere along the way...

Upabjojr:
> One other thing: i've tried to play a video locally (to VLC), but i'm getting an error. Not sure if it has something to do with my network configuration, so can you please check whether the play locally feature still works for you, or have we bungled it somewhere along the way...

Do you mean the nanohttpd server inside Kore? I've noticed that it's not really stable with videos, but can't say exactly what the problem is. Maybe nanohttpd isn't that good for streaming large files.

SyncedSynapse:
> Do you mean the nanohttpd server inside Kore? I've noticed that it's not really stable with videos, but can't say exactly what the problem is. Maybe nanohttpd isn't that good for streaming large files.

No, the nanohttp is used to stream from Kore to Kodi, which i'm able to do, though i didn't test with large files, just some demo files that i had laying around in my phone...
I was refering to trying to play the file from Kodi locally on the phone: Kore seems to get the correct url and tries to launch a video player, but then it can't get to the source (i suspect it's something on my end).

poisdeux:
> This looks good to me.
> 
> @poisdeux not sure if you want to check anything else, but for me this is good to me merged. If you're confortable with it, please merge it.
> 
> One other thing: i've tried to play a video locally (to VLC), but i'm getting an error. Not sure if it has something to do with my network configuration, so can you please check whether the play locally feature still works for you, or have we bungled it somewhere along the way...

Indeed. I just tried with VLC and it fails to open the URL. I'm using https://play.google.com/store/apps/details?id=video.player.videoplayer and that works fine. So I'm guessing this might be a VLC thingy. I don't have time now to look into it.

PR #721, opened 2020-03-26
Remove XBMC from interface strings 

### LINKED ISSUE #720: Remove XBMC from interface strings
XBMC was renamed to Kodi almost [6 years ago](https://kodi.tv/article/xbmc-getting-new-name-introducing-kodi-14). Probably, no one is still using versions before 14 up to now.

Shouldn't "XBMC" be dropped from interface strings like this one? https://github.com/xbmc/Kore/blob/e24dcca1ce2e3349f014f898deb4eed786abac78/app/src/main/res/values/strings.xml#L81

homocomputeris:
Closes #720, fixes minor issues (like inconsistent tags) and formats the file for a bit improved readability

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -84,20 +84,20 @@
     <string name="wizard_welcome_message"><![CDATA[
     Let\'s start by adding a media center. Make sure your Kodi is running, properly configured and on the same network as your device.<br/><br/>
     You can get help configuring it <a href="http://syncedsynapse.com/kore/kore-faq/">here</a>.<br/><br/>
-    When you\'re ready press <b><i>Next</i></b>.
+    When you\'re ready press <b>Next</b>.
     ]]></string>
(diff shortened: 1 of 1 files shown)


homocomputeris:
Maybe it's worth fixing it in the same PR:
- can this string
https://github.com/xbmc/Kore/blob/master/app/src/main/res/values/strings.xml#L256
be reused here?
https://github.com/xbmc/Kore/blob/master/app/src/main/res/values/strings.xml#L257-L266
- is this string
https://github.com/xbmc/Kore/blob/master/app/src/main/res/values/strings.xml#L267
the same as "swipe down"?

SyncedSynapse:
Yes it can be reused, but i don't see much of a gain there.
`pull_to_refresh` isn't being used.

homocomputeris:
> Yes it can be reused, but i don't see much of a gain there.

True. It's just a bit tidier.

PR #727, opened 2020-04-14
Migration to AndroidX and many more changes

GitHubMyHub:
Hallo everyone, first I want to thank all of the contributers for their hard work to this project.
I have updated the whole project to the new level.
261 changed lines in 236 files.

Kore changes
- updated to API level 16 to get all these new Google features
- migrating Project to AndroidX
- compiled with Android-R sdk
- Removed Butterknife (deprecated/no longer mainteined) and moved to View Binding recomended by Google
- Replaced Picasso by Glide
- Updated all libs to the recent versions
- Pictures were converted to webp to save space
- Remote Icons were converted to Vector drawables
- Launcher icon is now a vector drawable available for Api level 16+
- German translation is now in sync

The Kore app lost 11 - 7 = 3mb in size and is now state of the art.
I have not tested everything.

But i noticed in the notification above the remote control the Channel name is not displayed anymore.
I did not found out the error.
And the remote buttons are big now due to vector drawable migration.
I couldnt find out how to change the size.
But it could be an improvement for people with vision incapabilities.

Hope I could contribute to the project.


poisdeux:
Thank you. This is really nice!

But in my opinion, as this contains many changes that can break a lot of things I really urge you to issue separate PRs for this. Making the PRs smaller allows us to discuss the changes and makes it easier to merge them. Some can probably be merged immediately. My suggestion is to issue separate PRs for each of the following changes:

* migrating Project to AndroidX
* compiled with Android-R sdk
* Removed Butterknife (deprecated/no longer mainteined) and moved to View Binding recommended by Google
* Updated drawables (can go in a single PR)
   * Pictures were converted to webp to save space
   * Remote Icons were converted to Vector drawables
   * Launcher icon is now a vector drawable available for Api level 16
* German translation is now in sync

I'm not so sure about the following:
* updated to API level 16 to get all these new Google features. What new google features are you talking about?
* Updated all libs to the recent versions. **Why do these need to be updated? Do you use new APIs, are there bugs that need to be fixed?**
* Replaced Picasso by Glide. **Why do you want Glide?**

GitHubMyHub:
Thanks for your reply.
Unfortunately i havent commited after every step.
The changes after the Androidx (libs+xml files) migration were huge and I thought I should also update everything else.
And I saw that Kore has a Beta Program on Google Play Store because of that I didnt find the urge to push everything separately. Neither would I expect any immediate push to the master branch.
I thought rather of an Beta branch on GitHub.

1) I needed API level 16 for the 'expandabletextview" lib and the vector drawable support. And some other google libraries whicht I dont remember. You are right it is not quite a feature but a support.
2) There were no flaws but I just like everything up to date. :)
3) Glide is better supported than Picasso (Updates and Userbase).
It is also used in Googles Android programming courses which are held by their Developer Advocates.

SyncedSynapse:
First of all, thanks for the contribution. It's useful and it contains stuff that needed to be done. 

Nevertheless, i share @poisdeux concern's, regarding the scope of this PR. It's quite difficult to review, comment and test it as is, and it introduces a level of risk that could be much better managed if it was done in separate, smaller and more focused bits.

The separation that @poisdeux suggests seems good, but we are open to other that you feel is more appropriate. Taking the migration to AndroidX separately, for instance, would make a world of a difference IMO - that's a relatively straightforward change, relatively easy to review and test, and it constitutes a large portion of the code changes, so subsequent changes should be cleaner. Likewise for changes to assets (icons and such) and code. I might also have issues with using Glide instead of Picasso, that i would like to review and discuss separately, and that is impossible and confusing in this PR current form.

I hope you understand our point. We feel it's a valuable contribution that we want to get in, but it's an unmanageable risk merging it all at once (alpha/beta/whatever programs don't substitute code reviews - i prefer to prevent bugs than to fix them).

GitHubMyHub:
Thanks for you your reply.
Yes, I totally understand your point of view. Even though I have to do everything again. :D
I will start again in smaller chuncks like poisdeux said above.
It will now take longer because I now dont have much time to work on side projects.

SyncedSynapse:
Guess this can be closed, as it is being done in separate PRs. If not, comment.

PR #729, opened 2020-04-20
Open context menu when long-pressing select button

ByteHamster:


SyncedSynapse:
Why have 2 buttons do the same thing? It might be useful to you but i think it only confuses regular users.

ByteHamster:
> Why have 2 buttons do the same thing?

Because I feel like it is a more natural interaction when you are already used to Android. Most apps open context menus by long-pressing. I just installed Kore for the first time (coming from Yatse) and was confused how to open the context menu. The first thing I did was to long-press the center button. After that, I tried pressing the OSD button. After that, I finally found the right context menu button.

> It might be useful to you but i think it only confuses regular users.

I have used that same argument a lot in pull requests for [AntennaPod](https://github.com/AntennaPod/AntennaPod). Now it hits me too :D I feel like the long-press action does not hurt because it is not directly visible to the users.

I did a little test with 2 of my flat mates who have never used Kodi. I opened the sources page and requested them to edit the source. One flat mate pressed info, then context menu. The other one long-pressed the center (action disabled in that build), then pressed OSD, then the context menu button (just like me).

---

Off-topic: Just now, I found out what the button with the 9 squares actually does. It did not do anything in most menus I visited. After reading the source code, I even had to google the term OSD to finally get it (not a native English speaker).

app/src/main/java/org/xbmc/kore/ui/widgets/ControlPad.java
```diff
@@ -56,6 +56,7 @@
         void upButtonClicked();
         void downButtonClicked();
         void selectButtonClicked();
+        boolean selectButtonLongClicked();
```
losmika90, line 59:
```suggestion
        boolean selectButtonLongClicked();
```

app/src/main/java/org/xbmc/kore/ui/widgets/ControlPad.java
```diff
@@ -161,8 +162,17 @@ public void onClick(View v) {
 
     @Override
     public boolean onLongClick(View v) {
-        if ((onPadButtonsListener != null) && (v.getId() == R.id.info)) {
-            return onPadButtonsListener.infoButtonLongClicked();
+        if (onPadButtonsListener == null) {
```
losmika90, line 165:
```suggestion
        if (onPadButtonsListener == null) {
```

app/src/main/java/org/xbmc/kore/ui/widgets/ControlPad.java
```diff
@@ -161,8 +162,17 @@ public void onClick(View v) {
...
+            return false;
+        }
+
+        switch (v.getId()) {
+            case R.id.info:
+                return onPadButtonsListener.infoButtonLongClicked();
+            case R.id.select:
+                return onPadButtonsListener.selectButtonLongClicked();
+            default:
+                LogUtils.LOGD(TAG, "Unknown button " + v.getId() + " long clicked");
```
losmika90, line 175:
```suggestion
                LogUtils.LOGD(TAG, "Unknown button " + v.getId() + " long clicked");
```
```suggestion
                LogUtils.LOGD(TAG, "Unknown button " + v.getId() + " long clicked");
```
```suggestion
                LogUtils.LOGD(TAG, "Unknown button " + v.getId() + " long clicked");
```

SyncedSynapse:
Like i said (and you seem to understand my point of view), there's a dedicated button for the context menu, so offering it via a long press on another button is going to confuse more than help (imo). 

I don't doubt that there will be users that will try to long press the center button before the context button, but i think that those will be the exception. The context button is a visible action, vs a long press which is not, so it is definitely a superior option.

Regarding the OSD button, it's there, it might not be the most useful button, but it keeps the remote symmetric and it's one more function.

Sorry, but i'm closing this, as i don't feel it's the right choice.

SyncedSynapse:
By the way, thanks for pointing out your podcast app, going to try it.

PR #728, opened 2020-04-15
AndroidX migration

GitHubMyHub:
Migration to AndroidX
- Update to API level 29
- Update Java sourcecode and xml import statements
- Upgrade Butterknife Version 10 AndroidX support
- Set Java 8 compile options due Butterknife needs it

PR #730, opened 2020-04-20
Remember last-used tab

ByteHamster:
I never use the "artists" tab, so I always have to switch to the "albums" tab when opening the app. With this PR, Kore now remembers the last-used tab.


SyncedSynapse:
Looks good. We are currently in the process of migrating to AndroidX - there's already a PR for that - so, do you mind if we merge that one first, and then this - you'll have to adapt it slightly as far as i could see.

poisdeux:
Indeed looks good and also works nicely. I don't think we should wait on #728. It still needs some work and it can be rebased without any problems on this PR.

SyncedSynapse:
Merging this, as i trust @poisdeux  git-foo

PR #731, opened 2020-04-21
Do not set download button state when cancelling download

ByteHamster:
When pressing "Cancel" in the dialog or closing it by touching the back button, the download button stays highlighted, which is wrong.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -87,6 +87,11 @@ android {
         exclude 'META-INF/LICENSE.txt'
         exclude 'META-INF/NOTICE.txt'
     }
+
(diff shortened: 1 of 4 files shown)


poisdeux:
Works great. Just wondering why you only updated the Movie info, TV episode, and Music video pages?

ByteHamster:
> Just wondering why you only updated the Movie info, TV episode, and Music video pages?

`UIUtils.downloadSongs` (which displays the confirmation dialog) does not have direct access to the UI. One could pass the download button as a parameter but I don't think that this is too clean.

PR #735, opened 2020-04-26
Compile with Android-R sdk

GitHubMyHub:


SyncedSynapse:
Guess this PR is kind of mangled...

But i have another issue: Why do we want to use Android R sdk so soon? Maybe there's something i'm not seeing, but i can't see a good reason for it.

GitHubMyHub:
Sorry. My PR is not how expected.
But lets debate the Android-R compilation.
In my opinion it is very important to compile with the latest sdk to see if the app functionality is given with the latest Android 11 Version and to see what will be deprecated in the future version.

SyncedSynapse:
Well, the lasted SDK isn't R, that is still in preview mode. And it is advised that in order to use the 11 SDK, one should use Android 4 beta, which i don't think should be done right now.
Android 11 is still about 5-6 months away, there's plenty of time to validate it. 
Furthermore, ;ooking at the behaviour changes that 11 introduces it seems we need to validate if the scoped storage changes have an impact on Kore, and if they do make the necessary changes and update the SDK at the same time. Updating the SDK and see what breaks shouldn't be the default option.

SyncedSynapse:
@GitHubMyHub For the next PRs, you need to pull the latest version of Kore, and rebase your changes on top of that, otherwise they will include the migration to AndroidX, which is already merged (as has happened on this PR). 
I'm closing this for now, as i think that the new part - android r sdk - shouldn't be done right now. @poisdeux if you disagree, please comment.

PR #752, opened 2020-07-08
Support .torrent and magnet sharing to Elementum addon

SecT0uch:
Tested :+1: 
Should I edit the intant-filters in Manifest.xml accordingly ?

For reference: [Elementum](https://github.com/elgatito/plugin.video.elementum)

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
@@ -565,6 +565,8 @@ public class RemoteActivity extends BaseActivity
                     return null;
                 }
 
+            } else if (word.startsWith("magnet:")) {
+                return Uri.parse(word);
             }
         }
         return null;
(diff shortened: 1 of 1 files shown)


KarellenX:
@SecT0uch 

Thanks for your contribution. Unfortunately it is a breach of our forum rules which also applies to all Kodi sites including Github.

We do not provide any support or consideration for add-ons or other software which breaches the Copyright of the owner and enables piracy.

Your PR will be closed. Please read our rules here... https://kodi.wiki/view/Official:Forum_rules

SecT0uch:
I didn't read your rules indeed.
It's sad to banish torrent related apps, because torrent isn't only used for piracy.
A lot of Creative Commons content is also out there.

KarellenX:
I think torrent might be ok, depending on what it is coded to do, but you specifically referred to Elementum addon which is definitely used for Piracy.

SecT0uch:
@KarellenX Elementum addon only allows to download and play a given addon. 
There is absolutely no code in Elementum that fetches content. :+1: 

You must be confused with Elementum Burst. (Which is the one who fetches content)

It would be a good occasion to update the list of banned items to specify `Elementum Burst` rather than just `Elementum`

KarellenX:
Could you give me a link to the Elementum you are referring to?
Thanks.

SecT0uch:
[Elementum](https://github.com/elgatito/plugin.video.elementum/)
Thanks for looking into this

KarellenX:
Yep, that is banned.

SecT0uch:
Unfounded

PR #753, opened 2020-07-12
Add Twitch stream/VOD support to RemoteActivity

bclindner:
Hey! Not a Java guy, let me know if something about this code doesn't look right.

Realized Kore is missing the ability to share Twitch URLs to Kodi the other day. Let's fix that.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -76,6 +76,8 @@
                 <data android:host="soundcloud.com"/>
                 <data android:host="m.soundcloud.com"/>
                 <data android:host="www.arte.tv"/>
+                <data android:host="twitch.tv"/>
+                <data android:host="m.twitch.tv"/>
             </intent-filter>
 
             <meta-data android:name="android.service.chooser.chooser_target_service"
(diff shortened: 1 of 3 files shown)


SyncedSynapse:
Thanks, 

On a first glance, it looks good, though i need to test it a bit more.

2 comments:
- You should add the "twitch.tv" url in [AndroidManifest.xml](https://github.com/xbmc/Kore/blob/8ef29896f49f9814594c40afdfa44bc25cf77f6a/app/src/main/AndroidManifest.xml) , in the 2 sections that have other urls (play on kodi and queue on kodi).
- The code looks good, but i would ask you to move the logic to a new function in `PluginUrlUtils` called `toPluginUrlTwitch' similar to what is being done in the "Arte" branch (the condition check doesn't need to be a function there, just the logic). The other branches haven't been migrated but we should encourage new additions to be done there.

bclindner:
Fixed - just keeping the check inside the RemoteActivity code. Let me know how that looks.

SyncedSynapse:
Looks good. So sorry for the delay in merging this.

PR #755, opened 2020-07-29
Update strings.xml

izendegi:


SyncedSynapse:
Thanks.

In the future, please use Transifex to enter translations, as we are moving there (at least temporarily).
https://www.transifex.com/teamxbmc/kore-android-remote/dashboard/

PR #762, opened 2020-09-14
Add toggle to disable direct share

### LINKED ISSUE #761: Toggle to disable share to specific instance
## Feature suggestion
**Detailed Description**

On my device when I want to share a link there are four direct share targets and [two of them are taken by Kore](https://github.com/xbmc/Kore/issues/572) (Play on host and Queue on host). I'm using Kore with a single Kodi instance, so I don't really need these two targets, but rather have more space for other share targets, e.g. chats.
Therefore I suggest to add a setting to disable Kore's share targets.

### Possible Implementation

If I get an OK from you for this, I can propose a PR.

mueller-ma:
Fixes #761

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Kore - Kodi/XBMC remote for Android
 
 GitHub repository for the [Kore][1] Android app.
 
-Kore is the official remote for [Kodi](http://kodi.tv/), and aims to be a simple and easy to use  remote.
+Kore is the official remote for [Kodi](http://kodi.tv/), and aims to be a simple and easy to use remote.
 
 
 Building
--- a/app/src/main/java/org/xbmc/kore/Settings.java
+++ b/app/src/main/java/org/xbmc/kore/Settings.java
@@ -180,6 +180,8 @@ public class Settings {
     public static final String KEY_PREF_DISABLE_LOCAL_PLAY = "pref_disable_local_play";
     public static final boolean DEFAULT_PREF_DISABLE_LOCAL_PLAY = false;
 
+    public static final String KEY_PREF_DIRECT_SHARE_ENABLED = "pref_direct_share";
+
     public static final String KEY_PREF_LANGUAGE = "pref_language";
     public static final String KEY_PREF_SELECTED_LANGUAGE = "pref_selected_language";
 
--- a/app/src/main/java/org/xbmc/kore/service/HostChooserTargetService.java
+++ b/app/src/main/java/org/xbmc/kore/service/HostChooserTargetService.java
@@ -22,12 +22,14 @@ import android.content.IntentFilter;
 import android.graphics.drawable.Icon;
 import android.os.Build;
 import android.os.Bundle;
+import android.preference.PreferenceManager;
 import android.service.chooser.ChooserTarget;
 import android.service.chooser.ChooserTargetService;
 
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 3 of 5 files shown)


poisdeux:
The idea behind this is that you can disable the share targets to show up when you share a link from for instance the YouTube player right? If that is so I can't get that to work. I also do not understand your solution as the Android system determines the possible share targets by the intent filter as defined in the manifest.

mueller-ma:
This PR disables Kodi being shown as direct share target, e.g. https://i.stack.imgur.com/rOYol.png
There's only a limited amount of slots for the targets and I don't want Kodi to take up slots.

poisdeux:
Ah yes. Sorry I was confused and thought this would disable the share option completely.
It works great and although I never use it I can see why it can be annoying. 
I do have one issue which is that now the direct share option is by default disabled. I do think we should keep it enabled as this is the current behaviour.

@SyncedSynapse do you have anything to say about this? (We might need to think about restructuring the settings)

SyncedSynapse:
I understand the need of this, and agree with @poisdeux : the setting should be enabled by default, and i'm worried that if we create a setting for every difference in behavior, it will get overwhelming soon.
Any ideas for solving the settings issue? The obvious one is to create an "Advanced" submenu and just stick there everything that has minor impacts on the app.

mueller-ma:
IMO the direct share feature is only useful if more than one server is configured, so what about hiding the preference (and disable it internally) when there's only one server configured?

SyncedSynapse:
That's better, though it might be slightly confusing.

Another option is to, instead of having a global setting for this, create a specific setting for each host - on `HostInfo` -, which is configured on the host management screen - a checkbox bellow the current ones "Use TCP", etc. This would allow selecting which hosts to appear as direct share targets, tough it is a bit more work. It would be clearer to the user IMO.

mueller-ma:
> Another option is to, instead of having a global setting for this, create a specific setting for each host - on HostInfo -, which is configured on the host management screen - a checkbox bellow the current ones "Use TCP", etc. This would allow selecting which hosts to appear as direct share targets, tough it is a bit more work. It would be clearer to the user IMO.

That sounds good to me. I'm going to create a new PR.

mueller-ma:
So I started with this, but it seems to be much more work than I initially thought: I requires a database migration and it requires a save button in the settings. Currently you can only save the change if you test the connection.
Is there an easy way to migrate the database?

https://github.com/mueller-ma/Kore/commits/direct-share-instance-based

SyncedSynapse:
I took a look at your branch, and it looks good, though i haven't looked in detail. Seems that 99% is done, just missing the migration.

Regarding the migration, i believe it's just a matter of altering some more things on `MediaDatabase`:
1. Add a new constant `DB_VERSION_PRE_HOST_DIRECT_SHARE_TARGET = 11` and change `DB_VERSION=12`
2. `OnCreate` is already done
3. In `OnUpgrade`, add something like:
```
case DB_VERSION_PRE_HOST_DIRECT_SHARE_TARGET:
                // Add DIRECT_SHARE columns to HOSTS table
                db.execSQL("ALTER TABLE " + Tables.HOSTS +
                           " ADD COLUMN " + MediaContract.HostsColumns.DIRECT_SHARE +
                           " INTEGER DEFAULT 1;");
```

And i think that's it.

Regarding it only saving when the connection is tested, yeah, that's a bummer... We should do something about it, allowing saving even when the connection test fails, but i think that's another PR.

PR #768, opened 2020-10-07
Add ability to disable direct share on a per host basis

### LINKED ISSUE #761: Toggle to disable share to specific instance

mueller-ma:
Fixes #761

See #762 for more information.


SyncedSynapse:
Looks good to me. @poisdeux do you want to take a look before merging

Issue #776, opened 2021-01-03
Feature Request: Properly support sendtokodi (youtube-dl resolving of random links in Kodi)


gerion0:
## Feature suggestion
**Detailed Description**
Since quite a time the addon [sendtokodi](https://github.com/firsttris/plugin.video.sendtokodi/) exists, that allows playback of nearly [every](https://github.com/ytdl-org/youtube-dl/tree/master/youtube_dl/extractor) video and audio site with help of youtube-dl. For this addons also an [Android app](https://play.google.com/store/apps/details?id=com.yantcaccia.stk) exists to request the playback from an Android device. However, since Kore already is an Android app for Kodi, I want to suggest to enable this functionality also in Kore if the addon is available in the connected Kodi instance.

### Context
I regulary facing the situation to search some nice presented website (for example Youtube, die öffentlich-rechtlichen TV sender sites in Germany, Reddit, some Podcast sites and Online radio streams) for a medium. If I have found a specific medium, I have scripted my browser (on the desktop PC) so it sends this link via one click to Kodi to present it to other people (better sound and bigger display). This is currently not possible with my smartphone.

I assume that other people want the same, see #737, #671, #318 (kind of similar to this request) and the sheer existence of the other android app.

### Possible Implementation
I'm not an Android developer, only a user. I would expect to have a "Send to Kodi" option via the share menu of Android.

Upabjojr:
That's a great idea actually.

axelsimon:
Just tried the app, it's terrible : you have to enter the IP manually (and there's an issue, so i had to copy-paste my Kodi box's IP address) **and** you have to manually paste the link you want to open in a box.
Kore could / should provide a much better alternative, the "stream to Kodi" function in NewPipe, from Firefox / Fennec is super useful.

Upabjojr:
> you have to enter the IP manually (and there's an issue, so i had to copy-paste my Kodi box's IP address)

Do you mean the DHCP autodetect is not working on your WiFi?

> you have to manually paste the link you want to open in a box.

What do you mean? I just share the page from the browser through Kore, no need for copy-and-paste.

PR #777, opened 2021-01-03
Added SendToKodi addon support

Upabjojr:
Fixes https://github.com/xbmc/Kore/issues/776

### DIFF
--- a/app/src/main/java/org/xbmc/kore/Settings.java
+++ b/app/src/main/java/org/xbmc/kore/Settings.java
@@ -166,6 +166,9 @@ public class Settings {
         return Settings.KEY_PREF_REMOTE_BAR_ITEMS + hostId;
     }
 
+    public static final String KEY_PREF_ALWAYS_SENDTOKODI_ADDON = "pref_always_sendtokodi_addon";
+    public static final boolean DEFAULT_PREF_ALWAYS_SENDTOKODI_ADDON = false;
+
     public static final String KEY_PREF_NAV_DRAWER_ITEMS = "pref_nav_drawer_items";
     public static String getNavDrawerItemsPrefKey(int hostId) {
         return Settings.KEY_PREF_NAV_DRAWER_ITEMS + hostId;
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
@@ -35,6 +35,7 @@ import android.view.Menu;
 import android.view.MenuItem;
 import android.view.ViewTreeObserver;
 import android.view.WindowManager;
+import android.webkit.MimeTypeMap;
 import android.widget.ImageView;
 import android.widget.Toast;
 
@@ -578,6 +579,8 @@ public class RemoteActivity extends BaseActivity
      */
     private String toPluginUrl(Uri playuri) {
         String host = playuri.getHost();
+        String extension = MimeTypeMap.getFileExtensionFromUrl(playuri.toString());
+        String mimeType = MimeTypeMap.getSingleton().getMimeTypeFromExtension(extension);
         if (host.endsWith("youtube.com")) {
             String videoId = playuri.getQueryParameter("v");
             String playlistId = playuri.getQueryParameter("list");
@@ -628,8 +631,27 @@ public class RemoteActivity extends BaseActivity
             return PluginUrlUtils.toPluginUrlTwitch(playuri);
         } else if (PluginUrlUtils.isHostArte(host)) {
             return PluginUrlUtils.toPluginUrlArte(playuri);
+        } else if (!isMediaFile(mimeType)) {
+            // SendToKodi is a Kodi addon that is able to extract URLs from generic
+            // web URIs using the Python library "youtube-dl".
+            // Use it as a last resort, unless the URI extension is a known media file
+            // (in that case Kodi does not require an addon to play the link):
+            return "plugin://plugin.video.sendtokodi/?" + playuri.toString();
         }
         return null;
+    }
+
+    boolean isMediaFile(String mimeType) {
+        if (mimeType == null) {
+            return false;
+        } else if (mimeType.startsWith("audio")) {
+            return true;
+        } else if (mimeType.startsWith("image")) {
+            return true;
+        } else if (mimeType.startsWith("video")) {
+            return true;
+        }
+        return false;
     }
 
     // Default page change listener, that doesn't scroll images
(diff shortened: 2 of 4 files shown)


Upabjojr:
Questions:

1. SendToKodi seems to support anything that is free, even YouTube. Should we add an option to make it the default addon for YouTube videos? (Current YouTube addons requires Google's API key, which is annoying and privacy-invading).
2. Is there a JSON-RPC command to know if a particular addon is installed on Kodi?

gerion0:
> Questions:
> 
> 1. SendToKodi seems to support anything that is free, even YouTube. Should we add an option to make it the default addon for YouTube videos? (Current YouTube addons requires Google's API key, which is annoying and privacy-invading).

The only downside, I know of, is that SendToKodi is currently not in the official Kodi addon repository.

> 2. Is there a JSON-RPC command to know if a particular addon is installed on Kodi?

The Chrome browser extensions of SendToKodi seems to [check for this](https://github.com/firsttris/chrome.sendtokodi/blob/master/src/components/Form.jsx#L12).

Upabjojr:
> The only downside, I know of, is that SendToKodi is currently not in the official Kodi addon repository.

Well, there are some reasons not to officially include addons based on _youtube-dl_.

> The Chrome browser extensions of SendToKodi seems to [check for this](https://github.com/firsttris/chrome.sendtokodi/blob/master/src/components/Form.jsx#L12).

Interesting. The current logic could be modified in order to scan for addons before sharing the content link. A possible downside is a potential slowdown of Kore (having to wait for an answer from Kodi about the existing addons).

I've noticed that Kore [already does this](https://github.com/xbmc/Kore/blob/213374b6a67c24de49e7d91269d4f051804153f2/app/src/main/java/org/xbmc/kore/ui/sections/addon/AddonInfoFragment.java#L151), though that's probably triggered when you open the addon tab in the Kore GUI.

SyncedSynapse:
I need to check with the rest of the team whether there are some reasons i'm not aware to not include/use SendToKodi before merging this, and will report back.

@Upabjojr Regarding your question, as you said `GetAddons` and `GetAddonDetails` ([json-rpc](https://kodi.wiki/view/JSON-RPC_API/v12#Addons.GetAddonDetails) can be used to check whether an addon is present in Kodi (more on this bellow).

On a first look, the implementation looks good. I think that adding a new setting to control sharing to that addon is the correct way to do it (at least i'm not seeing any other way), though on a higher level i must confess that it feels like it's a small detail that shouldn't be in the settings. To reconcile these contradictory feelings i think that we might need to consider adding an "Advanced" section to the settings (though not right now on this PR), or preferably, use a call to `GetAddonDetails` to check if that specific addon is present, and only show the option if it is. This would be done in the `SettingsFragment`, and it might bring some issues as the method call is asynchronous, and i'm not sure if there are any issues updating a preferences fragment after it is constructed, but it would be more user-friendly IMO (showing that option to users that have the addon, and know what it's used for).

Let me know what you think. I'll report back.

Upabjojr:
> I need to check with the rest of the team whether there are some reasons i'm not aware to not include/use SendToKodi before merging this, and will report back.

OK, I know that _youtube-dl_ was temporarily suspended from GitHub last year, but it has since been reinstated. It could be a problem if people use _youtube-dl_ for permanent downloading, but only caching in order to stream the content should be OK.

> @Upabjojr Regarding your question, as you said `GetAddons` and `GetAddonDetails` ([json-rpc](https://kodi.wiki/view/JSON-RPC_API/v12#Addons.GetAddonDetails) can be used to check whether an addon is present in Kodi (more on this bellow).

> On a first look, the implementation looks good. I think that adding a new setting to control sharing to that addon is the correct way to do it (at least i'm not seeing any other way), 

Technically, if we get `GetAddonDetails` to check if _SendToKodi_ is installed, we could skip adding a setting at all. That is, if _SendToKodi_ addon is detected, it is reasonable to prefer it over all other addons. Though some people may still prefer to use the official addons.

> Let me know what you think. I'll report back.

I'm both OK with adding a settings menu option and not having a new one.

SyncedSynapse:
Sorry for the late reply, not much time lately. I didn't get a refusal from the team, so i believe there's no reason not to include this.

As i've said, this looks good to me, and i can merge it as-is. I think it would be nice if the check regarding whether `SendToKodi` is installed was made and only show the option if it was indeed installed, but it's up to you to implement that.

Alternatively (or additionally) the check could also be made in the `RemoteFragment`, before sharing the link to Kodi, therefore making sure that Kodi would process it.

It's up to you - do you want me to merge this as is and possibly improve it later or include something more on this PR?

Upabjojr:
> It's up to you - do you want me to merge this as is and possibly improve it later or include something more on this PR?

Scanning for addons on Kodi in probably going to increase the complexity of this PR a bit. At the moment I don't really have time to delve into it. I would prefer to keep this PR as it is, and maybe come back in the future.

Upabjojr:
I was actually thinking about the current procedure to share content through Kodi addons. Does it really make sense to keep adding regular expressions in Java source code? Wouldn't it be preferrable to either have an option-menu with customizable addons and matching patterns?

Well, even better would be if Kodi itself could be extended to perform the URL dispatching on the server side, and not on the remote app as it's currently being done. But I'm just hypothesizing.

axelsimon:
> Interesting. The current logic could be modified in order to scan for addons before sharing the content link. A possible downside is a potential slowdown of Kore (having to wait for an answer from Kodi about the existing addons).

Quick thought here: would it be useful to check if SendToKodi is installed once and keep track of the result to avoid checking on every case of sharing a video to Kodi? A periodic check could then verify if the add-on is still installed, to avoid issues of sending streams to a no-longer-installed Kodi add-on (say, if our last check `GetAddonDetails` check is over 120 min, check again).

> Technically, if we get GetAddonDetails to check if SendToKodi is installed, we could skip adding a setting at all. That is, if SendToKodi addon is detected, it is reasonable to prefer it over all other addons. Though some people may still prefer to use the official addons.

Fow what it's worth, i would tend to agree with this. If you have SendToKodi installed, you already know what you are doing a lot more than if you just have the Youtube add-on installed from the default repo, and are likely to want things to be sent to that add-on.
Also, as a side note, i doubt there are many "official" add-ons for Kodi. Dedicated, sure, and as such with likely more features than just playing a video stream (the Youtube one has the capacity to log-in, retrieve your favourites, etc.), but not official.

SyncedSynapse:
> I was actually thinking about the current procedure to share content through Kodi addons. Does it really make sense to keep adding regular expressions in Java source code? Wouldn't it be preferrable to either have an option-menu with customizable addons and matching patterns?
> 
> Well, even better would be if Kodi itself could be extended to perform the URL dispatching on the server side, and not on the remote app as it's currently being done. But I'm just hypothesizing.

I agree. That part of the code has very little structure, it grew organically so it isn't very pretty.

gerion0:
Is this worth a release? ;) The last one was quite some time ago.

SyncedSynapse:
I want to include some more fixes, but yeah, a release should come soon

PR #788, opened 2021-02-25
fix for seek bar issue for Kodi 19

daitj:
I have added new array type params in jsonrpc request which is required by Kodi 19. Then used percentage instead of time for the seek bar.

This is more of a quick and dirty fix. I am not so much of an Android developer, so I am leaving this here for some else to pick up from there.


daitj:
This implementation was following the JSON-RPC request created by Kodi' web remote, which apparently is not the official way of doing things. I will close this for now and create a new pull request with proper implementation

PR #792, opened 2021-04-07
Added buttons to go back or forward by 10 or 30 seconds

Upabjojr:
Skipping 10 or 30 seconds during the video playback may be useful, especially if you've missed something in the video.

This PR adds this capability to Kore.

I draw the icons myself, they are not very well shaped but at least they're copyleft :)

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/NowPlayingFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/NowPlayingFragment.java
@@ -318,6 +318,39 @@ public class NowPlayingFragment extends Fragment
         action.execute(hostManager.getConnection(), defaultStringActionCallback, callbackHandler);
     }
 
+    private void moveProgress(int deltaSeconds) {
+        int progress = mediaProgressIndicator.getProgress();
+        progress += deltaSeconds;
+        progress = Math.max(progress, 0);
+        progress = Math.min(progress, mediaProgressIndicator.getMaxProgress());
+        mediaProgressIndicator.setProgress(progress);
+        onProgressChanged(progress);
+    }
+
+    @OnClick(R.id.back_30_seconds)
+    public void backThirtySeconds(View v) {
+        moveProgress(-30);
+    }
+
+    @OnClick(R.id.back_10_seconds)
+    public void backTenSeconds(View v) {
+        moveProgress(-10);
+    }
+
+    @OnClick(R.id.go_to_position)
+    public void seekToPosition(View v) {
+    }
+
+    @OnClick(R.id.forward_10_seconds)
+    public void forwardTenSeconds(View v) {
+        moveProgress(10);
+    }
+
+    @OnClick(R.id.forward_30_seconds)
+    public void forwardThirtySeconds(View v) {
+        moveProgress(30);
+    }
+
     @OnClick(R.id.volume_mute)
     public void onVolumeMuteClicked(View v) {
         Application.SetMute action = new Application.SetMute();
--- a/app/src/main/java/org/xbmc/kore/ui/widgets/MediaProgressIndicator.java
+++ b/app/src/main/java/org/xbmc/kore/ui/widgets/MediaProgressIndicator.java
@@ -168,6 +168,10 @@ public class MediaProgressIndicator extends LinearLayout {
         durationTextView.setText(UIUtils.formatTime(max));
     }
 
+    public int getMaxProgress() {
+        return maxProgress;
+    }
+
     /**
      * Sets the play speed for the progress indicator
      * @param speed set to 0 to stop updating the progress indicator.
--- a/app/src/main/res/drawable/ic_back_10sec_24dp.xml
+++ b/app/src/main/res/drawable/ic_back_10sec_24dp.xml
@@ -1,0 +1,16 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:alpha="1.00"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24.0"
+    android:viewportHeight="24.0">
+  <path
+      android:fillColor="#FFFFFFFF"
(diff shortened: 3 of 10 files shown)


axelsimon:
Hi!
Can you explain what this brings, considering that skipping 10 seconds and 30 sec can already be easily done:
10 sec by pressing the left arrow on the remote control screen, and 30 seconds simply by pressing it 3 times (cf. the Kodi [wiki section on additive skipping](https://kodi.wiki/view/Skip_steps#Additive_skipping)).

Upabjojr:
> Can you explain what this brings, considering that skipping 10 seconds and 30 sec can already be **easily done**

Well, I wouldn't call the current procedure as being **easily done**, as you need to bring up the control bar on kodi, move the cursor to the progress bar and then start moving the progress cursor.

In fact, this PR adds a way to **easily do** this task. Consider that most modern players have this feature.

Upabjojr:
@axelsimon if you don't like these new buttons, I can add a new settings option to hide them.

axelsimon:
I think i'm starting to understand what you are talking about: your PR adds buttons to the "now playing" notification.

Btw, i wouldn't use the word "cursor", usually used to describe a mouse pointer, in the context of a touch-based system like Android, it's just too confusing. So i'm not sure what you mean by "start the moving progress cursor", if you mean moving the dot on a line representing the position in the currently playing media, that's not what i was referring to at all.

I was saying that skipping ahead (and back) 10, 15, 30 or 60 sec at a time can be done by pressing the ➡ (or ⬅) arrows of the 4-direction pad on the main Kore screen, which is a screen that Kore can show in an instant as it doesn't require unlocking the screen to show.

Honestly, i'm not sure this is faster: if i'm watching something and i pick up my phone, it'll automatically turn the screen on and Kore will show the remote screen, as long as that's the last screen i left it on. Then i just press ⬅.
In contrast, with what you suggest, i have to unlock my screen, pull down the notification bar, find the Kore notification and press the button.

I'm in no position to make decisions hear, so we'll just have to wait for @SyncedSynapse's opinion.

Upabjojr:
> I think i'm starting to understand what you are talking about: your PR adds buttons to the "now playing" notification.

Yes, *now playing* is actually a tag inside Kore, not a notification. There is currently no way to skip steps from inside the Kore app panels, except for the clumsy way of finding the Kodi cursor on the TV screen and moving it (or maybe there is, but I'm not aware of).

> I was saying that skipping ahead (and back) 10, 15, 30 or 60 sec at a time can be done by pressing the ➡ (or ⬅) arrows of the 4-direction pad on the main Kore screen, which is a screen that Kore can show in an instant as it doesn't require unlocking the screen to show.

It doesn't work on my app. I have the latest master branch of Kore and Kodi 18.7. Which version are you using? How are your app settings configured?

I can get the *skip steps* feature to work in the notification drawer (that is, outside of the app).

axelsimon:
> How are your app settings configured?

I think you need to enable "Use steps for seeking" in the settings and your Kodi device needs to be set up to connect over EventServer, rather than / as wall as TCP (which is the default these days, i think).

Also, to get the behaviour i was describing, you also want "Show over lockscreen" enabled.

Upabjojr:
> I think you need to enable "Use steps for seeking" in the settings and your Kodi device needs to be set up to connect over EventServer, rather than / as wall as TCP (which is the default these days, i think).
> 
> Also, to get the behaviour i was describing, you also want "Show over lockscreen" enabled.

So maybe the discussion could be turned into finding an easier way to have all of this working by default.

SyncedSynapse:
Just a quick comment, as i haven't been able to properly look at this.

@Upabjojr first i want to thank you for the contribution, but reading through this discussion i have the same questions as @axelsimon, specifically, if this is really needed.

I believe that somehow you aren't using the `EventServer` option, which does this step seeking behaviour by default when you use the remote buttons. Unless you changed with Kodi's advancedsettings.xml it should be enable in Kodi. In Kore, go to the media center configuration screen and make sure that the "Use EventServer" option is selected (is is on by default). Furthermore, make sure that there isn't something in your network blocking the shown port (9777 by default).

I understand that these buttons are added to the "now playing" panel accessible everywhere, but is that a big improvement, and doesn't it risk confusing the UI.

Upabjojr:
> I think you need to enable "Use steps for seeking" in the settings and your Kodi device needs to be set up to connect over EventServer, rather than / as wall as TCP (which is the default these days, i think).

At a first check, it seems not to work for me. Didn't investigate on why, maybe there are some issues with my WiFi.

Anyway, I suggest a different proposal. The option *use steps for seeking* replaces the rewind/fast forward buttons on the notification only.

- Shouldn't this be the case also in the Kore main panel? I keep having rewind/fast forward buttons there even if *use steps for seeking* is set to true.
- Should't the *now playing* fragment display both the rewind/fast forward buttons and the jump backward/forward buttons at the same time?

SyncedSynapse:
Check if, in Kore, the option "Use Event Server" is enabled for your Kodi instance. If not enable it and recheck. If it doesn't stay on that means it couldn't communicate through port 9777, in which case either event server isn't enabled in Kodi or the port is being blocked.

With that option you are able to seek using the remote arrows.

Regarding your suggestions:
1. Maybe, but notice that the `Use steps for seeking` option uses the EventServer channel (in the notification), so if it isn't enabled it won't work. But the behaviour of the rw/ff buttons in the kore main panel could be changed to seek if that option is enabled, using the EventServer if possible or the solution you propose in this PR if not possible. I think that's a viable solution.
2. I'm not fond of adding more buttons to the `now playing` fragment, as i believe it clutters the UI with behaviour that is replicated elsewhere.

PS. I'm going to merge a PR that fixes some issues with thumbnails in Kodi Matrix and that PR touches a lot of files, so i would appreciate if you could base any modifications you consider making on top of those.

SyncedSynapse:
Closing this, as it isn't coherent with the current codebase.

PR #797, opened 2021-05-02
Updated Media notification for Android 11.

spauldhaliwal:
Makes the required changes so that media session notifications conform to the new standard set in Android 11.
(https://developer.android.com/guide/topics/media/media-controls)

Some of the notification icons have been updated to more modern material design standards

Added permanent step skipping buttons in expanded notifications only

Rounded corners of notification artwork to fit in better with other modern android notifications / apps.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/service/NotificationObserver.java
+++ b/app/src/main/java/org/xbmc/kore/service/NotificationObserver.java
@@ -27,14 +27,13 @@ import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.drawable.Drawable;
 import android.os.Build;
-import android.preference.PreferenceManager;
+import android.support.v4.media.session.MediaSessionCompat;
 import androidx.core.app.NotificationCompat;
 import androidx.core.app.TaskStackBuilder;
-
+import androidx.media.app.NotificationCompat.MediaStyle;
 import com.squareup.picasso.Picasso;
 import com.squareup.picasso.Target;
 import org.xbmc.kore.R;
-import org.xbmc.kore.Settings;
 import org.xbmc.kore.host.HostConnectionObserver;
 import org.xbmc.kore.host.HostManager;
 import org.xbmc.kore.jsonrpc.notification.Player;
@@ -43,6 +42,7 @@ import org.xbmc.kore.jsonrpc.type.PlayerType;
 import org.xbmc.kore.ui.sections.remote.RemoteActivity;
 import org.xbmc.kore.utils.CharacterDrawable;
 import org.xbmc.kore.utils.LogUtils;
+import org.xbmc.kore.utils.RoundedTransformation;
 import org.xbmc.kore.utils.UIUtils;
 import org.xbmc.kore.utils.Utils;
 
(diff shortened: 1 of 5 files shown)


SyncedSynapse:
Thanks for contributing, and sorry for the delay.

I have a few comments:
1. Please configure you IDE to not automatically format or change existing whitespace. I don't claim that the codebase is "clean" or uniformly formatted, but if every PR formats whole files according to the submitter settings, it aggravates the uniformity issue and makes reviewing the PR harder; 
2. Does this PR works for you? On my device it never shows the notification. I'll point out in the code the lines i had to remove in order for this to show;
3. I don't fully agree with altering the play/pause icons. Using those icons but keeping the others (fwd, rwd, etc) is inconsistent, and it is inconsistent with the icons on the app, so please keep the standard play/pause icons. A change of icons would have to apply on more than the notification.
4. On my device i see no change regarding the rounded corners. This might be an issue with Samsung's skin, but in my device the artwork is already on the right side and the round corners of the notification are correctly applied, so i don't understand what is the difference here. A picture of before/after would help.

Some more comments in the code.

app/src/main/java/org/xbmc/kore/service/NotificationObserver.java
```diff
@@ -224,51 +230,83 @@ private void notifyPlaying(PlayerType.GetActivePlayersReturnType getActivePlayer
...
+                || getItemResult.type.equals(ListType.ItemsAll.TYPE_MOVIE)
+        ) {
+            builder.addAction(rewindIcon, service.getString(R.string.rewind), rewindPendingIntent) // #0
+                    .addAction(jumpBackIcon, service.getString(R.string.rewind), jumpBackPendingIntent) // #1
+                    .addAction(playPauseIcon, service.getString(R.string.play), playPausePendingIntent)  // #2
+                    .addAction(jumpForwardIcon, service.getString(R.string.fast_forward),
+                            jumpForwardPendingIntent) // #3
+                    .addAction(ffIcon, service.getString(R.string.fast_forward), ffPendingIntent)   // #4
+                    .setStyle(new androidx.media.app.NotificationCompat.MediaStyle()
+                            .setShowActionsInCompactView(1, 2, 3));
```
SyncedSynapse, line 290: The order of the buttons should be 
`jumpBack, rewind, play, ffwd, jumpFwd`
and the buttons shown on the compact view should be the 
`rewind, play, ffwd`
as that is the current default.

app/src/main/java/org/xbmc/kore/service/NotificationObserver.java
```diff
@@ -224,51 +230,83 @@ private void notifyPlaying(PlayerType.GetActivePlayersReturnType getActivePlayer
...
+        // Creates Media Style with MediaSession Token required to display
+        // media controls below quick settings in Android 11 and above.
+        if (Build.VERSION.SDK_INT >= 21) {
+            final MediaSessionCompat mediaSession = new MediaSessionCompat(this.service,
+                    "MediaService");
+            final MediaStyle mediastyle = new androidx.media.app.NotificationCompat.MediaStyle()
+                    .setMediaSession(mediaSession.getSessionToken());
+            builder.setStyle(mediastyle
+                    .setShowActionsInCompactView(1, 2, 3));
+        }
```
SyncedSynapse, line 308: I had to comment this block to make the notification appear. This is probably interfering with the previous code, as it is again setting the style on the builder, and at least my Android version doesn't like that.
This should be better integrated with the previous code somehow.

app/src/main/java/org/xbmc/kore/service/NotificationObserver.java
```diff
@@ -325,6 +364,7 @@ private void showNotification(Bitmap bitmap) {
             HostManager hostManager = HostManager.getInstance(service);
             hostManager.getPicasso()
                     .load(hostManager.getHostInfo().getImageUrl(poster))
+                    .transform((new RoundedTransformation(40, 0)))
```
SyncedSynapse, line 367: As i've said, i didn't see a difference. Is it possible to point it out.

tiagogaspar8:
This is a great feature and it would be amazing for it to be on Kore!
@spauldhaliwal could you help us get this merged into kore?

PR #800, opened 2021-06-09
Delete Kore_App_Icons.zip

mueller-ma:
In my opinion zip files shouldn't be committed to a git repository.
Furthermore this causes an error at F-Droid and has to be removed there: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/7248/diffs?commit_id=815ddad70d72d86ac64fdc8721fa88dadafa11b7


SyncedSynapse:
Ok, but what about adding the files included in the zip, so that they don't get lost.

mueller-ma:
Aren't they included in https://github.com/xbmc/Kore/commit/fa6796475c5f97b2e8367557c7163219cdce3e90 and therefore already in the repository?

> so that they don't get lost

Even when deleted they live in the git history.

But I can unzip the file, if you want.

SyncedSynapse:
Didn't notice they were already included, sorry. Will merge this.

Issue #157, opened 2015-12-23
Allow volume changes with hardware keys even while screen is off


ThiefMaster:
I didn't even know that this is possible, but my audiobook player app has this annoying feature that even when the phone is locked and the screen is off the volume buttons change the playback volume. But while it's annoying in this app, I think it would be a great feature (maybe with an option to toggle it) in Kore.

There are many movies with some very loud and some very quiet scenes and having to unlock the phone to change the volume is not very nice. Being able to use the volume keys immediately would be pretty useful for this.

JoseFSilva:
+1 
I also feel this would be a nice improvement to have.
I currently work around this by having the screen always on when using Kore, which is not good at all for the phone's battery life/health :/

Peque:
Duplicate of #235? :wink:

Peque:
I created a new issue (#298), requesting a more complete media control while the screen is locked (i.e.: volume and playback controls).

PR #855, opened 2022-08-07
Control volume with hardware buttons when Kore is not focused (#157)

caffeineFox:
Since MediaSession was implemented, I thought it'd be a good point to add volume control via the VolumeProviderCompat. This way, playback volume can be adjusted even when Kore is not in focus and/or the screen is locked (but turned on). To my mind, it resolves Issue #157 as far as possible for now.


SyncedSynapse:
Thanks, and sorry for the delay in reviewing this.

Looks good, but a couple of comments:
- There's a setting that defines whether the hardware volume keys should control Kodi's volume, and that should be respected. Take a look at `Settings.java`, look for `KEY_PREF_USE_HARDWARE_VOLUME_KEYS` and check how it's used in `VolumeControllerDialogFragmentListener`. I'll point out in the code where to check if the setting is active.
- For consistency with the current code, please use spaces, with 4 spaces per level.
-

caffeineFox:
Hi, thanks for the feedback! I'll add the changes in the next days :)

caffeineFox:
Hi @SyncedSynapse I guess it's ready now :D I reindented the file and now support the user preference concerning hardware volume key usage

SyncedSynapse:
Ok, thanks.
Just one thing:
Note that in the constructor for `RemoteVolumeProviderCompat` you're calling `setCurrentVolume`, and you're constructing it, before checking if the hardware volume keys are enabled. I don't think it's serious, but can you please consider this. I'll point out in the code.

SyncedSynapse:
By the way, i'll probably be unavailable for the rest of the week, so i might not be able to merge this before next week.

caffeineFox:
Thanks for pointing that out. To my mind, it was not relevant to check the preference there since `setCurrentVolume` solely affects the `RemoteVolumeProviderCompat`. I thought it was necessary to get the currently set volume on Kodi and set it to the Compat so the volume used as reference for display (in the Android volume slider) is in sync. After doing some explorative testing, it seems that everything works just fine without initial synchronization, so I removed the call and method.

SyncedSynapse:
I didn't mind the syncing on the constructor, i just thought that you might only create `remoteVolumePC` after checking if the hardware volume keys were enabled, but this is ok also.

There's a slight bug: when you call any of the `HostConnectionObserver.register*` you need to call `unregister` somewhere (in this case in `onDestroy`), or it will keep a reference to the object, preventing GC, as `hostConnectionObserver` is a singleton.

I'll merge this as is and fix that later.

Thanks

Issue #912, opened 2022-09-21
Non-FOSS in F-Droid build


licaon-kter:
c728508ad65a03c24b65c6b6794a438f223c6a1c

```
$ gradle :app:dependencies --configuration releaseRuntimeClasspath
...
+--- androidx.core:core-google-shortcuts:1.0.1
|    +--- com.google.firebase:firebase-appindexing:20.0.0
|    |    +--- com.google.android.gms:play-services-base:17.5.0
|    |    |    +--- com.google.android.gms:play-services-basement:17.5.0
|    |    |    \--- com.google.android.gms:play-services-tasks:17.2.0
|    |    |         \--- com.google.android.gms:play-services-basement:17.4.0 -> 17.5.0 (*)
|    |    +--- com.google.android.gms:play-services-basement:17.5.0 (*)
|    |    \--- com.google.android.gms:play-services-tasks:17.0.0 -> 17.2.0 (*)
```
ref: https://github.com/xbmc/Kore/blob/v3.0.0-beta.1/app/build.gradle#L106

licaon-kter:
Think we can, for now, exclude gms at build time:
```
    implementation('androidx.core:core-google-shortcuts:1.0.1') {
        exclude group:'com.google.android.gms'
    }
```

Can somebody test this APK? https://gitlab.com/fdroid/fdroiddata/-/commit/4a42435346fda40ab37978bc65e71b8195b88ce8#note_1108571409

Is GMS used elsewhere too? Could I PR such a change here directly?

SyncedSynapse:
Play services aren't used in Kore, didn't think they would be dragged by the shortcuts library...

I'm obviously not sure if excluding that breaks something internally on the shortcuts library, but PR this and i'll make some tests.

SyncedSynapse:
Closing this as it will be managed through the pr

woheller69:
why is `exclude group:'com.google.android.gms` not a fix?

licaon-kter:
Because the lib was build from non-FOSS deps, so it's non-FOSS.

woheller69:
If I remove the whole thing 

```
    implementation('androidx.core:core-google-shortcuts:1.1.0') {
        exclude group:'com.google.android.gms'
    }
```

it still builds and works. What is the dependency used for?

licaon-kter:
[...]

PR #913, opened 2022-09-21
Exclude non-FOSS GMS deps from google-shortcuts

licaon-kter:
Fix https://github.com/xbmc/Kore/issues/912


SyncedSynapse:
I'm going to hold on merging this for a week or so, because we just started running beta testing on the Play store, and i don't want to stop that to include this. I'm also not comfortable sending this to prod without testing, so what i propose is do the release on Play store without this, then merge this (if my tests don't show up anything) and then F-Droid can pick it up.

licaon-kter:
That being said... this might not be enough, see comments: https://gitlab.com/fdroid/fdroiddata/-/commit/a2baa158a7dfd8d31c2bb3d88025aea81a52e042

SyncedSynapse:
Any other suggestions? I'm not familiar with f-droids build process

PR #915, opened 2022-09-27
Fix sharing play intents only working the first time

SyncedSynapse:
When sharing play intents, the `OpenShareActivity` launches the `RemoteActivity` at the end. In Android versions prior to 12, when doing a second share, it doesn't run `OpenShareActivity` and instead directly opens the `RemoteActivity`, so the play intent isn't delivered to Kodi. This doesn't happen on Android 12+, but to fix it we need to set flags when calling `RemoteActivity` `startActivity` so that the `OpenShareActivity` is called in the next sharing intent.


Harvie:
I can't play youtube videos since upgrade to v3.0.0 on Redmi8. So it has probably broken something.
Good think that now it actualy opens the kore window after sharing youtube video to kodi. Before it only started playing the video but didn't opened the kore. Now it does not play the video in kodi, but opens kore reliably.

PR #953, opened 2023-01-22
Update ArtistAlbumsListFragment.java

JungleJim42:
In artist view, show albums sorted by year. I tested this change with Android Studio.


SyncedSynapse:
Ideally one should be able to choose the sort, but this is better than sorting by name, so i'm merging it.

JungleJim42:
I agree that ideally one should be able to choose the sort, but this is beyond my capabilities. Thanks for this solution!

PR #958, opened 2023-02-14
Allow titles in settings to use more than one line

mueller-ma:
Some of them are truncated in the German translation, even with text size set to 'small' (in the OS settings).

### DIFF
--- a/app/src/main/res/xml/preferences.xml
+++ b/app/src/main/res/xml/preferences.xml
@@ -15,42 +15,49 @@
    limitations under the License.
 -->
 <!--suppress AndroidElementNotAllowed -->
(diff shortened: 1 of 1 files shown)


SyncedSynapse:
I'm not against this change, but doesn't this point to the need to review the "offending" titles? I mean, i believe that according to Android's guidelines, settings titles should occupy one line, so we should aim for that. 

Nevertheless i'm merging this, as it's an improvement

mueller-ma:
I agree that some translations should be shortened and someone already did it: https://github.com/xbmc/Kore/commit/023b0b4f738d8c213e7222193fb54285ff19683a#diff-7049c01dbaca09e41462205bc840bb2bdf6766f0b832e1ff68059c7fdfafadc8L284-R287

PR #961, opened 2023-03-16
Match SVT Play alphabetical video ids

### LINKED ISSUE #960: Play on Kodi stopped working for SVT Play
A while ago, SVT changed their link IDs from numbers to text, eg.
`https://www.svtplay.se/video/eDmaWYx/en-vildare-varld/1-de-stora-grasatarna`.

This regex is outdated, so it's falling back to plugin.video.sendtokodi. It could be enough to change `d+` to `w+`, but I don't know yet as I haven't been able to see if the svtplay add-on accepts the rest of the URL as-is or if it needs only the text code.
https://github.com/xbmc/Kore/blob/44cbc6fe8b49d7b703880e217b1282cce63b482f/app/src/main/java/org/xbmc/kore/ShareOpenActivity.java#L278

jswetzen:
Should fix #960, the only issue is a change from numeric to alphabetical video ids.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ShareOpenActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ShareOpenActivity.java
@@ -275,7 +275,7 @@ public class ShareOpenActivity extends Activity {
             } else if (host.endsWith("svtplay.se")) {
                 try {
                     Pattern pattern = Pattern.compile(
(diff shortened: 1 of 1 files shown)


SyncedSynapse:
Thanks, will merge.

PR #963, opened 2023-04-17
EPG: prepend plot outline, show full plot on click

RoccoDev:
This PR displays the plot outline in the EPG list view. Some broadcasters (like Italy's RAI) use the plot outline as part of the actual plot (presumably to have more text).

Additionally, this PR adds an alert containing the full outline and plot when the list item is clicked. Normally, the small description shown in the list view is ellipsized, and there was previously no way to read the full text.

Before:
immagine

After:
immagine

Plot outline + Plot (how it should be):
immagine

Unfortunately, in this case the word "tempo" is split off by the newline I added to delimit outline and plot, though for broadcasters that make proper use of the outline text (the majority, as far as I could see), I feel this is the best way to display it.

### DIFF
--- a/app/src/main/java/org/xbmc/kore/jsonrpc/type/PVRType.java
+++ b/app/src/main/java/org/xbmc/kore/jsonrpc/type/PVRType.java
@@ -161,6 +161,24 @@ public class PVRType {
             }
         }
 
+        /**
+         * Returns both the plot outline and the plot, leaving each empty if absent.
+         *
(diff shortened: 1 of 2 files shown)


SyncedSynapse:
This looks good.

I'm not merging it immediately because i don't use PVR much so i don't have a strong opinion on how it should look/behave. I'll leave this up for a couple of days to see if someone else gives feedback, and merge it afterwards.

PR #973, opened 2023-07-12
Improve haptic feedback of the control pad

Drumber:
The current vibrations on the remote control buttons are verry strong.  I've modified the `handleVibration()` util method to use the HapticFeedbackConstants. This is the [recommended way](https://developer.android.com/develop/ui/views/haptics/haptic-feedback#view) and ensures consistent behavior across apps.

The vibration was only present on the arrow-buttons, so I've also added it to the other buttons on the control pad.


SyncedSynapse:
This is failing the build check and i don't see a cause for it. It build ok locally but apparently not on Github.

Can you trigger a rerun of the check to see if it was a temporary thing (on Actions, click this PR and rerun the check. I don't have that option available probably because i'm not the PR creator).

Thanks.

Drumber:
I can't see a re-run button either. (I think only those with write access to the repository can trigger workflows).

Drumber:
I have merged the changes from master. You should be able to approve the workflow and thus rerun the checks.

SyncedSynapse:
Thanks, it passes now.

Regarding the PR, i find that the double vibration (when a key is pressed and when is released) is somewhat jarring. I would prefer a single vibration, when the key is pressed.

Honestly i don't know what are Android's guidelines on this, but i don't think i've ever encountered a double vibration. I understand its usefulness when navigating through long list, giving feedback on the release, but it feels somewhat unnatural when doing a single click, and i believe those are the most used types of clicks that users perform.

This just needs the removal of the condition ```(action == MotionEvent.ACTION_UP || ``` from ```UIUtils.handleVibration``` so that it only vibrates on key down and on cancel.

Drumber:
I think the "quality" of the haptic feedback might depend on the vibration motor of the device and how well it can simulate key presses. However, I see the point that it could be annoying for some users, so I removed the vibration on key release events.

Issue #978, opened 2023-10-01
Add support for the Invidious add-on when sharing YouTube URLs


wtimme:
## Feature suggestion
YouTube URLs shared with Kore should be playable with the Invidious add-on.

**Detailed Description**

Shared YouTube URLs currently require for the YouTube add-on (`plugin.video.youtube`) to be installed in Kodi. As outlined in #947, however, there are other plugins that support the playback of YouTube videos as well, such as [the Invidious add-on](https://github.com/petterreinholdtsen/kodi-invidious-plugin).

### Context

The YouTube add-on `plugin.video.youtube` requires an API key. As a privacy-conscious user, I do not want to sign in to Google and link my viewing habits to an account. Rather, I want to watch YouTube anonymously - and Invidious is a great way to do so.

### Possible Implementation

Kore's settings could have a new `ListPreference` (e.g. titled "YouTube add-on") that allows users to select their preferred YouTube add-on. Internally, Kore could store the add-on's ID in the preferences.

In `ShareOpenActivity.toPluginUrl()`, Kore would check which add-on is selected and, based on that, create the `Uri` specific to the add-on.

I have already tried this with hard-coded values locally, and it seems to work as expected. Adding a `ListPreference` would allow us to extend this to other add-ons as well.

One thing to note is that not every add-on supports queuing. The Invidious add-on, for example, does not. But for the MVP, I think it would be perfectly fine to only have the video played back.

### Additional context, screenshots or links
Here are some relevant links or screenshots:

- [`ShareOpenActivity.toPluginUrl()`](https://github.com/xbmc/Kore/blob/389abee55a77dd630722cd7944f3ef550ca2bd48/app/src/main/java/org/xbmc/kore/ShareOpenActivity.java#L242)

*Note: Team Kodi will consider this item however we will not make any promises if it will be included.*

basilgello:
> not every addon supports queuing

Look at https://github.com/Tolriq/repository.yatse.kodi/pull/41 how queuing is done - the plugin:// url is put into playlist which is queued. Also I encountered an issue with -1 as addon_handle and made a workaround PR referenced in the same link (but not merged yet :( )

PR #979, opened 2023-10-01
Add support for the Invidious YouTube add-on

wtimme:
# Introduction

This branch adds basic support for the Invidious YouTube add-on, as proposed in #978.

The only thing that is working so far is the direct playback of shared video URLs, which - as mentioned in the original issue - is an MVP that could be extended upon.

This new feature is lacking translation. If I can help with that as well let me know.

# How to test

1. Checkout the branch and run the app in the emulator
2. Connect the app to a Kodi instance
3. Install NewPipe
4. Open NewPipe, select a video and share the video with Kore

Kodi should display a brief loading indicator and then playback the video.

---

Thanks in advance for your time and consideration!

### DIFF
--- a/app/src/main/java/org/xbmc/kore/Settings.java
+++ b/app/src/main/java/org/xbmc/kore/Settings.java
@@ -173,6 +173,7 @@ public class Settings {
     public static final boolean DEFAULT_PREF_ALWAYS_SENDTOKODI_ADDON = false;
 
     public static final String KEY_PREF_YOUTUBE_ADDON_ID = "pref_youtube_addon_id";
+    public static final String DEFAULT_PREF_YOUTUBE_ADDON_ID = "plugin.video.youtube";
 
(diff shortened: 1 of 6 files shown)


SyncedSynapse:
Ok, i'm not 100% sure about this. 

My issue is that there's no modular way for specifying plugins in Kore, so each plugin that is supported is being tightly coupled in Kore's code, and that's not something that i find scalable. Up until now, plugins dealt with different providers (arte.tv, svtplay, etc), so that was transparent to the user, but this requires a new option in the settings, getting very visible to the user, and it can get unwieldy soon.
This had already happened with the SendToKodi plugin, which i merged, but i had even greater reservations...

Anyway, i'm merging this for now, but with the caveat that i might revisit this decision in the future (and certainly also the inclusion of the SendToKodi one). The UI on the settings might change, or this might get removed - apologies in advance if that happens, but hope you understand the rationale.

PR #980, opened 2023-10-03
Add back-button navigation in plugin directories

### LINKED ISSUE #691: [Feature Request] Android "Back" button should go to previous screen
## Feature suggestion
**Detailed Description**
When browsing Add-ons in Kore, the back-button always goes out of the Add-on rather than go back one screen within the add-on. This is highly inconvenient. People expect to go back one time, not jump out of the Add-on when pressing Back.

### Context
I think it is contra-intuitive that Back jumps out of the add-on, rather than go back to the previous screen.

autumngray:
Fixes #691


SyncedSynapse:
Looks good.

For consistency sake, it should also be applied to `AddonListContainerFragment`, which i'll do.

PR #990, opened 2023-11-19
[gradle] signing and jenkins cleanup

fuzzard:
Gradle signing wasnt actually working. Have added v1-v3. I was under the impression you needed v2+ to submit these days, but either way, it doesnt change anything for your playstore submission stuff.

Added the zipalign to the gradle build file.

Removed the apksign script, as its no longer needed. Ive removed all the variables pertaining to it from jenkins.

Ran this test job https://jenkins.kodi.tv/view/Android/job/Kore/4871/ to confirm signing. Below snippet to verify

```
static:Kore administrator$ jarsigner -verify -verbose -certs kore-20231119-1c03adc-gradle_signing.apk 
<snipped for brevity>
- Signed by "CN=XBMC Foundation, OU=Android platform, O=XBMC Foundation, L=US, ST=US, C=us"
    Digest algorithm: SHA-256
    Signature algorithm: SHA256withRSA, 2048-bit key

jar verified.

Warning: 
This jar contains entries whose certificate chain is invalid. Reason: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
This jar contains entries whose signer certificate is self-signed.
This jar contains signatures that do not include a timestamp. Without a timestamp, users may not be able to validate this jar after any of the signer certificates expire (as early as 2042-03-13).
POSIX file permission and/or symlink attributes detected. These attributes are ignored when signing and are not protected by the signature.

The signer certificate will expire on 2042-03-13.
```

Feel free to just commit your own variant of this, just wanted to provide the info to save you hunting around.


SyncedSynapse:
Thanks.

PR #1005, opened 2024-01-19
Remove unused non-FOSS dependency

woheller69:
See #912


SyncedSynapse:
That library was introduced in
https://github.com/xbmc/Kore/pull/839/commits/ddb05f2b85a4eac04d14452efd57d5a3abf32782

Because Google's documentation regarding dynamic shortcuts recommends it:
https://developer.android.com/develop/ui/views/launch/shortcuts/creating-shortcuts

On a more detailed read, it seems that it is only necessary for pushing shortcuts to things like Google Assistant, and the goal was to have dynamic shortcuts available on the launcher, which is possible without the library, so i agree with the removal.

Issue #966, opened 2023-06-08
Sending live Youtube videos to Kodi doesn't work


AlexNk:
## Bug report

* The option "Prefer SendToKodi" is disabled in Kore
* "SendToKodi" addon is not installed
* Kodi has the latest YouTube addon: https://github.com/anxdpanic/plugin.video.youtube/
* Sending live videos to Kodi ends with an error or silently does nothing
* The same videos are sent fine from desktop browser addon: https://github.com/dirkjanm/firefox-send-to-xbmc
* Logs contain the following:
```
2023-06-08 22:16:25.160 T:744     error <general>: Unable to find plugin plugin.video.sendtokodi
2023-06-08 22:16:25.160 T:744     error <general>: Playlist Player: skipping unplayable item: 1, path [plugin://plugin.video.sendtokodi/?https://www.youtube.com/live/scyY2bmL7gE?feature=share]
```

Sample video: https://www.youtube.com/live/scyY2bmL7gE

AlexNk:
I was able to send the same video to Kodi by doing the following:
1. Open it in a browser. On this step sending still doesn't work
2. Change it to https://www.youtube.com/watch?v=scyY2bmL7gE - same ID, but no /live/
3. Share - send to Kodi

SyncedSynapse:
You are probably right, the url parser only supports a limited number of variations, and the /live/ isn't one of them.

PR #1010, opened 2024-02-17
Support parsing youtube.com/{live,shorts} URLs shared from other apps

peat-psuwit:
YouTube now use the new format for live streams and shorts. The pattern
is `https://www.youtube.com/{live,shorts}/<video id>`. But other than
pattern change, the video ID is otherwise completely compatible with the
old format.

Since this pattern is similar to https://youtu.be/ format, re-use that
code by re-shuffle the conditions a bit. The change is applied to both
default and Invidious plugins.

Fixes: https://github.com/xbmc/Kore/issues/966

### DIFF
--- a/app/src/main/java/org/xbmc/kore/utils/PluginUrlUtils.java
+++ b/app/src/main/java/org/xbmc/kore/utils/PluginUrlUtils.java
@@ -40,8 +40,9 @@ public class PluginUrlUtils {
     @Nullable
     public static String toDefaultYouTubePluginUrl(Uri playUri) {
         String host = playUri.getHost();
+        String path = playUri.getPath();
 
-        if (host.endsWith("youtube.com")) {
+        if (host.endsWith("youtube.com") && path.equals("/watch")) {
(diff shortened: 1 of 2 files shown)


MoojMidge:
@peat-psuwit - plugin.video.youtube supports loading a variety of different YouTube urls using the following plugin url format
```plugin://plugin.video.youtube/uri2addon/?uri=PERCENT_ENCODED_ORIGINAL_YOUTUBE_URL```

All you would need to do is check whether the host ends with `youtube.com` or `youtu.be` (there are other hosts that are used and supported, but those two would cover the majority of shared urls), and then send the original url as a percent encoded value using the `uri2addon` plugin endpoint.

This allows a number of additional urls (channels/users, clips, embeds, etc.) to be supported, including shorts and live streams, without needing to parse the url in Kore.

peat-psuwit:
@MoojMidge's idea is now implemented in #1019. Closing this PR in favor of that one. Thank you.

PR #1024, opened 2024-10-13
revert "Flip RTL text before sending it #184"

amiremohamadi:
seems like it's been fixed in kodi while ago and we should not reverse the input string anymore.
@SyncedSynapse

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/RemoteActivity.java
@@ -28,7 +28,6 @@ import android.view.WindowManager;
 
 import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.widget.Toolbar;
-import androidx.core.text.TextDirectionHeuristicsCompat;
 import androidx.core.view.OnApplyWindowInsetsListener;
 import androidx.core.view.ViewCompat;
 import androidx.core.view.WindowCompat;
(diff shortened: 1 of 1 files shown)


basilgello:
If minimal version of Kodi that Kore supports has the original issue fixed, then simply removing the code is enough. Otherwise, you need to guard that with a flag that is set once on connection estabilishment.

amiremohamadi:
@basilgello reversing the text has never been a correct solution. imagine a situation in which our text is a mixture of rtl and ltr words.

basilgello:
OK, just checking as I was not aware of the problem (and its "fix")

SyncedSynapse:
No idea about what the correct behaviour should be. 
@amiremohamadi  Do you have a link to the Kodi PR that fixed the issue?
@royiarchy Do you have any input into this.

If the issue was fixed in a specific version of Kodi, what basilgello was referring to was keeping the code, but guarding it with something like 
'''
if (! hostManager.getHostInfo().isLeiaOrLater) {
    // Removed code to reverse text
}
'''
Not sure if necessary

amiremohamadi:
@SyncedSynapse I guess it's solved in [this commit](https://github.com/xbmc/xbmc/pull/21746/commits/f07bb93ed8bca546d9bfdf9a57cfb6f370e65280).
IMO keeping the previous code (reversing the whole text) for the older versions is unnecessary as it will still mess up mixed RTL/LTR texts (even in the older versions of kodi).

PR #1043, opened 2025-08-16
feat: enable sharing multiple files to Kore

rien333:
This PR  enables sharing multiple audio/video/image files to Kore.  Previously, sharing multiple files to Kore required a share action per file.

# Use cases

My main use case is sharing ("streaming") whole albums from a music player app — previously, I had to queue every song one by one, which quickly got tiresome. An example of an app that can share whole albums is [Auxio](https://github.com/OxygenCobalt/Auxio).

Additionally, I can see this PR being useful for creating slideshows from your camera roll.

# Notable changes

Sorry for the long diff, but this change called for some refactoring. The real meat is in the first two commits, so its wise to start there.  The rest are revertible cosmetic tweaks (e.g. renaming vars) and minor follow up fixes. 

* Moved URI-extraction logic into a dedicated function. This is a separate commit ([the first one](94a9d229c5fe60216530a6e4f0888941897ca533)),  so that it is easier to verify that this new function produces identical results.
   * This new function IMO makes the code easier to follow, and fixes some small stuff in the previous routine (e.g. a duplicate call to `getData()`), which frankly was a bit all over.
* Adjusted `HttpApp` to make read permission for `content://` URIs more persistent, using `context.grantUriPermission`. In my tests, this read permission is now only revoked when Kore is killed. Having this permission be more persistent is essential for the use case of this PR:  without it, Kore  would hit permission errors after just one song.
*  Renamed `applicationContext` to just `context` in `HttpApp` (see ca2aa4a28d97cfc055ca24a2c82f4e50bc35e9c9). This aligns with actual usage. In fact, it's precisely because we pass the `Activity` context instead of the application context that the current code works ([this blogpost discusses URI permissions in greater detail](https://commonsware.com/blog/2025/05/31/once-again-uri-access-lifetime-shorter-than-you-might-think.html))  
* Some functions now expect _lists_ of URIs/URLs, instead of singletons. In case of `OpenSharedUrls`, that proved really helpful: its `onSuccess` callback now signals "hey, I'm done with all the URLs", so that we avoid running `finish()` prematurely. 

# Discussion

*  Kore can now send mixed content to Kodi (e.g. slideshow with pictures and videos). Kore currently interprets these as videoplaylist, which matches the old behavior for unknown/odd intent types. 
    * We could instead extract mimeypes from individual URIs, match them to playlist types, and then rewrite `OpenSharedUrls` such that it explicitly handles collections of URIs of different types. This seems overkill, though. 
    * Or we could add a `UNKNOWN` playlist type (equal to `-1`), and handle that. There is precedent for an `UNKNOWN` type both [in the API](https://kodi.wiki/view/JSON-RPC_API/v13#Playlist.Type) and Kodi's `playlistTypes.h`.
* I tried a lot of patterns until I hit upon my current solution. For example, we share a lot functionality with `LocalMediaFileListFragments`, which I tried to extract into a shared class.  I'm not sure why I eventually abandoned this approach  (maybe `content://` URIs were more difficult to convert to `LocalFileLocation` than I initially expected), but I'm happy to reconsider this or other approaches.
* I can make a `OpenSharedUrl` (singular) action, if desired. But frankly, that code would see no use currently. 

# Future work

* Kodi logs still show frequent HTTP errors. Likely harmless, but worth investigating.
* Music shared in this way won't get any metadata, such as albumartist, release date, etc. That is a bummer if you use last.fm, and it also sucks cosmetically.  IMO, this is Kodi's problem: from my test, ffmpeg is perfectly capable of extracting metadata from the first 0.1mb of a music file served over http. However, Kodi has been built with a 'local library'-first philosophy, so it's currently not not fully leveraging ffmpeg's capabilities here.


app/src/main/java/org/xbmc/kore/ShareOpenActivity.java
```diff
@@ -178,32 +195,6 @@ private Uri getUrlInsideIntent(Intent intent) {
         return null;
     }
 
-    private Uri getShareLocalUriOrHiddenUri(Intent intent) {
```
rien333, line 181: This function became redunant after `extractUriFromIntent` — it checks for URIs at the exact same "locations". (`intent.getParcelableExtra(Intent.EXTRA_STREAM)` is a bit more modern tho). 

I also don't really like how it casts to URLs, then back to URIs, and then later in the code, back to a URL again.

app/src/main/java/org/xbmc/kore/ShareOpenActivity.java
```diff
@@ -92,29 +92,18 @@ protected void handleStartIntent(Intent intent, boolean queue) {
...
-        if (videoUri == null) {
-            // Couldn't understand the URI
+        Uri uri = extractUriFromIntent(intent);
+        if (uri == null) {
+            // Failed to extract a URI
             finish();
             return;
         }
 
-        String url = toPluginUrl(videoUri);
```
rien333, line 114: the video centric language was unneeded imo

rien333:
I'll fix the conflicts in a bit!

PR #1051, opened 2025-10-25
feat: show/hide local hidden files

### LINKED ISSUE #1050: [Feature Request] Add option to hide hidden local files in Kore
## Feature suggestion
**Detailed Description**
In Kore, all local files are currently shown, including hidden files. This can clutter the file browser and make it harder to navigate. A setting to hide hidden files would improve usability.

### Context
Most users do not need access to hidden files (like system files or directories). Hiding them by default would make list cleaner, while a toggle could allow advanced users to view them if needed.

### Possible Implementation
Add a “Show hidden local files” toggle in settings. When enabled, hidden files appear; when disabled, they are filtered out.

### Additional context, screenshots or links
- [Android File.isHidden() reference](https://developer.android.com/reference/java/io/File#isHidden())  

*Note: Team Kodi will consider this item however we will not make any promises if it will be included.*

gopal-panigrahi:
Added a toggle in settings (“Show hidden local files”) to hide or display hidden local files.

fixes #1050

### DIFF
--- a/app/src/main/java/org/xbmc/kore/Settings.java
+++ b/app/src/main/java/org/xbmc/kore/Settings.java
@@ -90,6 +90,10 @@ public class Settings {
     public static final String KEY_PREF_PAUSE_DURING_CALLS = "pref_pause_during_calls";
     public static final boolean DEFAULT_PREF_PAUSE_DURING_CALLS = false;
 
+    // Show hidden local files
+    public static final String KEY_PREF_SHOW_HIDDEN_LOCAL_FILES = "pref_show_hidden_local_files";
+    public static final boolean DEFAULT_PREF_SHOW_HIDDEN_LOCAL_FILES = false;
+    
     // Other keys used in preferences.xml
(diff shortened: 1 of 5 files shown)


SyncedSynapse:
Thanks for this, looks good, though i'm not convinced that this should be an option on the global application settings. It is a specific option, that should only be available where it makes sense (in the `LocalMediaFileListFragment`), and preventing the uncontrolled growth of options in the global settings that will confuse the user, sooner or later. 

I would suggest the removal of the option from the global settings and the inclusion on the local menu of `LocalMediaFileListFragment`. Basically create `public void onCreateOptionsMenu(@NonNull Menu menu, MenuInflater inflater)` and `public boolean onOptionsItemSelected(MenuItem item)` there and include the new option.

Similar to what is being done on other fragments, for instance on `MovieListFragment`.

Hope you understand the reasoning.

gopal-panigrahi:
Yeah, the argument makes sense. I have implemented the options menu in `LocalMediaFileListFragment`.

PR #1054, opened 2025-11-03
refactor: playlist position tracking using player properties

### LINKED ISSUE #1053: [Enchancement] Optimize playlist position tracking using player properties
## Enchancement suggestion
**Detailed Description**

Optimize playlist position tracking by using the player's position property directly instead of searching through all playlist items by ID comparison.

Currently, `updateCurrentlyPlayingItem()` loops through the entire playlist to find the active item. Since `PlayerType.PropertyValue.position` already contains this information, we can use it directly for better performance.

### Context

This optimization improves performance, especially for large playlists, by eliminating unnecessary searching. It also simplifies the code while maintaining all existing functionality.

### Possible Implementation

- Store `PlayerType.PropertyValue` alongside other player state
- Use the `position` property directly instead of looping through playlist items

*Note: Team Kodi will consider this item however we will not make any promises if it will be included.*

gopal-panigrahi:
This PR optimizes playlist position tracking by using `PlayerType.PropertyValue.position` directly instead of searching through all playlist items by ID comparison. The current implementation uses an O(n) loop to find the active item, but since the player already provides the position, we can access it directly with O(1) complexity. This change improves performance (especially for large playlists), simplifies the code, and maintains all existing functionality including drag detection and UI selection behavior.

closes #1053

### DIFF
--- a/app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
+++ b/app/src/main/java/org/xbmc/kore/ui/sections/remote/PlaylistFragment.java
@@ -97,6 +97,7 @@ public class PlaylistFragment extends Fragment
      */
     private ListType.ItemsAll lastGetItemResult = null;
     private PlayerType.GetActivePlayersReturnType lastGetActivePlayerResult;
+    private PlayerType.PropertyValue lastGetPropertiesResult = null;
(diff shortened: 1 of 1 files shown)


SyncedSynapse:
Looks good, merging

